Issue #6, opened 2026-03-04
Migrate panel do() from NumPy propagation to PyMC-native engine


drbenvincent:
## Context

The `pymc` branch (#5) migrated cross-sectional `do()` to a PyMC-native engine using `pm.do()` + `compute_deterministics` / `pm.sample_posterior_predictive()`. This eliminated the duplicated NumPy reimplementation of structural equations, link functions, and residual noise.

However, **panel `do()` (`run_panel_do`) still uses the old NumPy propagation engine**. It manually iterates over units and time steps, resolving lagged variables, tracking adstock state, applying transforms via `apply_numpy()`, and adding residual noise — all outside the PyMC computation graph.

## What remains in `simulate.py`

The following functions exist solely to support `run_panel_do()`:

- `run_panel_do()` (~200 lines) — time-forward NumPy propagation
- `_apply_link()` — manual inverse link functions
- `_add_residual_noise()` — manual residual noise sampling
- `_build_transform_map()` — duplicates transform resolution from the compiler
- `_apply_panel_transform()` — NumPy transform application with adstock state
- `_get_panel_col_value()` — lag resolution helper
- `_parse_lag()` — lag column name parser

## Why this matters

1. **Duplicated logic**: transforms, link functions, and residual distributions are implemented twice (PyMC tensors in `compile.py` and NumPy arrays in `simulate.py`). Bugs fixed in one place can be missed in the other.
2. **Consistency**: cross-sectional and panel `do()` use different engines, which could produce subtly different results for the same model.
3. **Maintenance**: adding a new family or transform requires updating both implementations.

## Challenges

Panel `do()` has requirements that make a naive `pm.do()` approach insufficient:

- **Time-forward propagation**: lagged variables (e.g., `sales_lag1`) at time *t* depend on simulated values at time *t-1*, creating a sequential dependency.
- **Adstock state**: geometric carry-over accumulates across time steps within each unit.
- **Unit-specific random effects**: each unit has its own intercept/slopes drawn from a shared distribution.

Possible approaches:

1. **`pytensor.scan`**: encode the time-forward loop as a scan operation in the PyMC model at compile time. This would make the full panel simulation part of the computation graph, but may be complex for arbitrary DAG structures with multiple lagged variables.
2. **Hybrid**: use `pm.do()` per time-step in a Python loop, accumulating results. Simpler but potentially slower.
3. **Batched `pm.do()`**: restructure the panel data so that each time step can be processed as a batch intervention on the generative model.

[...]

Issue #2, opened 2026-03-01
Docs: explain conditional equations vs full joint SCM and estimand semantics


drbenvincent:
## Summary

Add a focused documentation update that explains:

1. conditional-equation fitting vs full joint generative SCM modeling,
2. how pathmc's causal DAG layer differs from the PyMC factor/computational graph,
3. how intervention integration choices affect estimands (ATE/ATT/ATU semantics).

## Motivation

Advanced PyMC users often expect interventions to be represented as modified PyMC model graphs. In pathmc, inference and intervention simulation are intentionally separated. This is powerful but currently under-explained, which can cause confusion when interpreting `pm.model_to_graphviz(...)` outputs and `do()` behavior.

### Concepts page

- Add/expand a "mental model" section with:
  - two-layer architecture (causal DAG vs PyMC estimation graph),
  - what gets compiled once vs what is simulated post-fit,
  - clear statement that current `do()` is posterior-draw propagation, not PyMC graph surgery.

### Mathematical explainer

- Side-by-side explanation for mediation:
  - structural equations,
  - joint factorization `p(X,M,Y)=p(X)p(M|X)p(Y|X,M)`,
  - conditional-estimation view used in current compile path.
- Explain when results align and when they can differ.

### Estimand semantics

- Define ATE, ATT, ATU explicitly.
- Document how exogenous integration policy impacts reported intervention summaries.
- Add practical guidance for interpreting `kind="mean"` vs `kind="predictive"`.

### Example updates

- Mediation example should include:
  - single-node intervention on treatment,
  - single-node intervention on mediator,
  - direct-only contrast with fixed mediator,
  - explicit linkage from manual contrast to `.ate()` helper.

## Acceptance criteria

- [ ] Concepts docs include architecture diagram and explicit DAG-vs-factor-graph explanation.
- [ ] A dedicated section explains conditional vs full-joint modeling with equations.
- [ ] ATE/ATT/ATU definitions are documented in one place with clear semantics.
- [ ] Mediation example demonstrates treatment-only, mediator-only, and direct-only contrasts.
- [ ] Cross-links added between Concepts, do-operator docs, and mediation example.

## Nice-to-have

- Add one compact "Which graph should I inspect?" callout:
  - `model.graph()` for causal structure,
  - `pm.model_to_graphviz(model.pymc_model)` for estimation factorization.

PR #7, opened 2026-03-04
Add panel do() engines (batched, scan) and restructure concepts docs

drbenvincent:
## Summary

- **Two new PyMC-native panel `do()` engines**: `"batched"` (Python loop + `pm.do()` per time step) and `"scan"` (`pytensor.scan`-based), selectable via `panel_engine=` keyword alongside the existing `"numpy"` engine.
- **Time-varying interventions**: `set=` now accepts NumPy arrays for per-time-step intervention values (e.g., boost spend for weeks 11–20 only).
- **Per-time-step output**: `DoResult.by_time(var)` exposes the time dimension for visualising temporal dynamics like adstock ramp-up and ramp-down.
- **Comprehensive engine equivalence tests** (`test_panel_engines.py`) confirming identical results across all three engines within numerical precision.
- **Restructured documentation**: "Concepts" is now a navbar dropdown with 5 focused sub-pages (Model Specification, Transforms & Families, Causal Inference, Panel Data, Time-Forward Panel Simulation) replacing the old monolithic overview.
- **New `panel_engines.qmd` example** demonstrating a digital-only spend intervention with adstock ramp-up/ramp-down across all three engines.
- **Explanatory diagrams** across concept pages: graph surgery, transform pipeline, latent mediators, residual covariance, unit boundaries, adstock cascade.

Towards #6 — adds two PyMC-native panel engines; the existing NumPy engine is retained for comparison rather than deleted.

Towards #2 — new concept pages explain the do-operator architecture, causal inference tools, and panel simulation mechanics.

## Test plan

- [x] All 208 non-slow tests pass (`pytest -x -v -m "not slow"`)
- [x] `ruff check` clean on all changed files
- [x] Engine equivalence tests verify numpy/batched/scan produce matching results (within tolerance) for models with adstock, saturation, random intercepts, and lagged endogenous variables
- [ ] Render docs site (`quarto render docs/`) and verify navigation, diagrams, and cross-references
- [ ] Run slow (MCMC sampling) tests to confirm panel engine tests pass end-to-end


drbenvincent:
## Code review fixes

Three issues identified and fixed:

### 1. Bug fix: `latent = set()` shadowing the `set` parameter

In both `run_panel_do_batched` and `run_panel_do_scan`, the default initialization `latent = set()` called the *parameter* `set` (a dict) instead of the Python builtin `set()`. This would crash with `TypeError` if anyone called these functions without passing `latent`. Fixed by moving the `latent` default before the `set` reassignment and using `frozenset()`.

### 2. Warning + regression tests for scan predictive mode on temporal models

The scan engine adds residual noise *post-hoc* (after computing the full mean trajectory), so noise at time t does not propagate through lags/adstock to time t+1. This is correct for `kind="mean"` but produces less faithful predictive draws than numpy for models with autoregressive structure.

Added:
- A `UserWarning` when scan + `kind="predictive"` + temporal state (lags or adstock) is detected, recommending `panel_engine="numpy"`.
- Code comment at the post-hoc noise site explaining the limitation.
- Expanded docstring in `run_panel_do_scan`.
- 5 warning tests (warns on lag model, warns on adstock, no warn for mean mode, no warn for numpy).
- 1 variance regression test: verifies numpy predictive produces higher late-time variance than scan on a lag model (noise propagation through temporal dependencies amplifies variance over time).

### 3. Input validation for time-varying intervention arrays

Added `_validate_set_lengths()` called at the top of all three engine functions. Gives a clear error like `set['spend'] has shape (5,), expected (15,)` instead of a confusing NumPy reshape/index error. 4 tests cover all engines + unknown engine name.

All 212 non-slow tests pass. `ruff check` clean.

PR #173, opened 2026-03-22
Fix scan carry recursion for endogenous lagged outcomes

### LINKED ISSUE #144: [CRITICAL] Scan panel models propagate noise-free carry state for endogenous lags
## Summary

Inside the `pytensor.scan` step function in `_compile_scan_panel`, Gaussian endogenous variables store their linear predictor `mu` (noise-free) in the carry state. For autoregressive models like `Y ~ lag(Y)`, the scan computes `mu_t = beta * mu_{t-1}` — a deterministic trajectory from the initial condition — instead of the correct `mu_t = beta * Y_{t-1}` (conditioned on the actual observed/sampled value).

## Impact

- **Estimation:** The likelihood surface sees a deterministic recursion rather than conditioning on observed intermediate values, producing biased parameter estimates for autoregressive panel models.
- **do():** Temporal propagation is deterministic even for `kind="predictive"`.

## Scope

- **Affects:** Panel models with `lag(Y)` where Y is endogenous (autoregressive models)
- **Does NOT affect:** MMM models with adstock/saturation of exogenous channels (the primary use case). Adstock carry state is a deterministic function of exogenous data — no noise propagation issue arises.

## Root cause

The `pytensor.scan` body computes all `mu` values deterministically, but free RVs are emitted outside the scan. There is no mechanism for the carry state to reflect the actual (sampled/observed) variable values.

## Possible fix direction

Stochastic latent variables already work around this with an innovation term (`mu + sigma * innovation_t`), suggesting a similar approach could address observable endogenous variables. This would require emitting noise inside the scan and structurally changing the RV emission pattern.

## Reference

Technical audit report: `docs/dev/technical_audit_2026_03_16.md` finding C1.

drbenvincent:
Closes #144

## What changed
- add scan-level carry innovations for lagged Gaussian/StudentT endogenous variables
- add `_use_observed_carry` switch so observed models carry forward realized panel values
- set `_use_observed_carry=1` in the observed model while keeping generative/do recursion stochastic

## Validation
- conda run -n pathmc ruff check --fix pathmc/compile.py pathmc/model.py
- conda run -n pathmc ruff format pathmc/compile.py pathmc/model.py
- conda run -n pathmc pytest tests/test_lag_syntax.py -x -v
- conda run -n pathmc pytest tests/test_panel_engines.py::TestPredictiveModeTemporal::test_scan_predictive_on_lag_model -x -v

### DIFF
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -1357,6 +1357,7 @@ def _compile_scan_panel(
             [_init_carry(init_endo[k]) for k in endo_keys]
             + [_init_carry(init_adstock[k]) for k in adstock_keys]
             + [_init_carry(init_exog_lag[k]) for k in exog_lag_bases]
+            + [None for _ in stochastic_carry_vars]
         )
 
         # Non-sequences: all parameters
@@ -1428,6 +1429,7 @@ def _compile_scan_panel(
 
             new_endo: dict[str, Any] = {}
             new_adstock = dict(prev_adstock_state)
+            carry_mu: dict[str, Any] = {}
 
             for var in endo_keys:
                 beta = ns_map.get(f"beta_{var}")
@@ -1460,6 +1462,7 @@ def _compile_scan_panel(
                     else:
                         new_endo[var] = mu
                 elif var in stochastic_carry_vars:
+                    carry_mu[var] = mu
                     sigma_val = ns_map[f"sigma_{var}"]
                     sampled_state = mu + sigma_val * carry_innov_t[var]
                     obs_state = obs_carry_t[var]
@@ -1479,6 +1482,7 @@ def _compile_scan_panel(
             out = [new_endo[k] for k in endo_keys]
             out += [new_adstock[k] for k in adstock_keys]
             out += [exog_t.get(k, pt.zeros(n_units)) for k in exog_lag_bases]
+            out += [carry_mu[k] for k in stochastic_carry_vars]
             return out
 
         results = pytensor.scan(
@@ -1493,9 +1497,16 @@ def _compile_scan_panel(
         if not isinstance(results, list):
             results = [results]
 
+        carry_mu_start = n_carry
+        carry_mu_results: dict[str, Any] = {
+            var: results[carry_mu_start + i]
+            for i, var in enumerate(stochastic_carry_vars)
+        }
+
         # --- emit deterministics and free RVs ---
         for i, var in enumerate(endo_keys):
-            mu_all = results[i]  # (n_times, n_units)
+            carry_all = results[i]  # (n_times, n_units)
+            mu_all = carry_mu_results.get(var, carry_all)
 
             if var in latent:
                 if var in stochastic_latent_set:
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -1299,15 +1299,47 @@ def _compile_scan_panel(
             col: np.zeros(n_units, dtype="float64") for col in adstock_keys
         }
 
-        innovation_nodes: dict[str, Any] = {}
+        # Flag toggled by caller: 0 for generative recursion, 1 for observed carry.
+        use_observed_carry = pm.Data("_use_observed_carry", np.array(0, dtype="int8"))
+
+        endo_lag_bases = sorted(
+            {base for _col, (base, _k) in lag_cols.items() if base in endo_set}
+        )
+        stochastic_carry_vars = sorted(
+            v
+            for v in endo_lag_bases
+            if v not in latent
+            and families.get(v, "gaussian") in ("gaussian", "studentt")
+        )
+
+        observed_carry_nodes: dict[str, Any] = {}
+        for var in stochastic_carry_vars:
+            if var in data_sorted.columns:
+                observed_panel = _reshape_to_panel(data_sorted, var, n_units, n_times)
+            else:
+                observed_panel = np.full((n_times, n_units), np.nan, dtype="float64")
+            observed_carry_nodes[var] = pm.Data(
+                f"_obs_carry_{var}", observed_panel.astype("float64")
+            )
+
+        latent_innovation_nodes: dict[str, Any] = {}
         for var in stochastic_latent:
-            innovation_nodes[var] = pm.Normal(
+            latent_innovation_nodes[var] = pm.Normal(
                 f"innovations_{var}", mu=0, sigma=1, shape=(n_times, n_units)
             )
 
-        sequences = [exog_data_nodes[k] for k in exog_keys] + [
-            innovation_nodes[k] for k in stochastic_latent
-        ]
+        carry_innovation_nodes: dict[str, Any] = {}
+        for var in stochastic_carry_vars:
+            carry_innovation_nodes[var] = pm.Normal(
+                f"carry_innovations_{var}", mu=0, sigma=1, shape=(n_times, n_units)
+            )
+
+        sequences = (
+            [exog_data_nodes[k] for k in exog_keys]
+            + [observed_carry_nodes[k] for k in stochastic_carry_vars]
+            + [latent_innovation_nodes[k] for k in stochastic_latent]
+            + [carry_innovation_nodes[k] for k in stochastic_carry_vars]
+        )
 
         def _init_carry(arr: np.ndarray) -> Any:
             """Convert init array to tensor for scan carry state.
@@ -1330,6 +1362,8 @@ def _compile_scan_panel(
         # Non-sequences: all parameters
         non_seq_list: list[Any] = []
         non_seq_names: list[str] = []
+        non_seq_list.append(use_observed_carry)
+        non_seq_names.append("_use_observed_carry")
         for var in endo_keys:
             if beta_rvs[var] is not None:
                 non_seq_list.append(beta_rvs[var])
@@ -1348,9 +1382,14 @@ def _compile_scan_panel(
         for var in stochastic_latent:
             non_seq_list.append(sigma_rvs[var])
             non_seq_names.append(f"sigma_{var}")
+        for var in stochastic_carry_vars:
+            non_seq_list.append(sigma_rvs[var])
+            non_seq_names.append(f"sigma_{var}")
 
         n_seq = len(sequences)
         n_exog_seq = len(exog_keys)
+        n_obs_carry_seq = len(stochastic_carry_vars)
+        n_latent_innov_seq = len(stochastic_latent)
         n_endo = len(endo_keys)
         n_adstock = len(adstock_keys)
         n_exog_lag = len(exog_lag_bases)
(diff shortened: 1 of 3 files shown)


drbenvincent (commented):
## Review: Fix scan carry recursion for endogenous lagged outcomes

### Summary

**Issue #144 is valid and important.** The PR correctly diagnoses the root cause and introduces the right *mechanism* (teacher-forcing + stochastic innovations), but the implementation conflates the carry state with the likelihood mean, introducing two new bugs that are worse than the original. The PR must be revised before merge.

---

### Issue #144 assessment

The issue correctly identifies that `pytensor.scan` in `_compile_scan_panel` propagates the noise-free linear predictor `mu` as carry state. For AR models (`Y ~ lag(Y)`), this means `mu_t = beta * mu_{t-1}` — a deterministic recursion — instead of the correct `mu_t = beta * Y_{t-1}`. The scope note that MMM models with exogenous adstock are unaffected is also correct.

**Verdict: issue is warranted and correctly scoped.**

---

### Bug 1: Degenerate likelihood — regression coefficients are unidentifiable

The scan step function sets `new_endo[var]` for stochastic carry vars to the teacher-forced observed value (when `use_observed_carry=1`). This same `new_endo[var]` is emitted as the scan output and stored as `mu_{var}`:

```python
# compile.py, inside step_fn (~L1462-1471)
elif var in stochastic_carry_vars:
    ...
    new_endo[var] = pt.switch(
        use_observed_carry_t, observed_or_sampled, sampled_state
    )
```

Then outside the scan (~L1507-1536):

```python
pm.Deterministic(f"mu_{var}", mu_all)   # mu_all = observed data!
...
pm.Normal(var, mu=mu_all, sigma=sigma, ...)  # Normal(obs | obs, sigma)
```

During estimation, `mu_all` equals the observed data itself. The likelihood becomes `Normal(observed | observed, sigma)`, which has **zero gradient for all regression coefficients**. Parameters are identified only by their priors.

**Empirical proof:** Per-variable logp decomposition shows that perturbing `beta_sales` by +5.0 changes logp by exactly −0.25 nats — entirely accounted for by the `Normal(0, 10)` prior on beta. The `sales` likelihood contribution is invariant to beta.

### Bug 2: Double noise in the generative model

In generative mode (`use_observed_carry=0`), `new_endo[var] = mu + sigma * carry_innov` (the sampled state including noise). This is emitted as `mu_all`. Then `pm.Normal("sales", mu=mu_all, sigma=sigma)` adds a *second* independent noise term:

```
sales = (mu + sigma * innov_carry) + sigma * innov_normal
      = mu + sigma * (innov_carry + innov_normal)
```

Effective variance is `2 * sigma²`, not `sigma²`. In recursive AR models, this compounds across time steps, leading to explosive trajectories.

---

### Regression evidence

| Test | Main | PR branch |
|------|------|-----------|
| `TestLagModel::test_mean_is_finite` | PASS | PASS |
| `TestLagModel::test_ate_positive` | PASS | **FAIL** (ATE = −1.16e12) |
| `TestLagModel::test_hdi_is_valid` | PASS | PASS |
| All non-slow tests (378) | PASS | PASS |

---

### Root cause

The scan uses `new_endo[var]` for two purposes:

1. **Carry state** — fed to `prev_endo` at the next time step for lag resolution
2. **Emitted output** — becomes `mu_all`, used for `mu_{var}` deterministic and the Normal likelihood

Before this PR, `new_endo[var] = mu` (linear predictor) — correct for purpose 2 but wrong for purpose 1 (the original bug). After this PR, `new_endo[var]` = observed data (estimation) or `mu + sigma * innov` (generative) — correct for purpose 1 but wrong for purpose 2.

**The fix must decouple these two roles.**

---

### Recommended fix approach

Add a separate non-carry output from the scan for the linear predictor `mu`. Pytensor scan supports non-carry outputs by setting `outputs_info` to `None`:

1. **Inside `step_fn`**: compute `mu` as the linear predictor (unchanged), set `new_endo[var]` to the carry state (teacher-forced or sampled), and return the `mu` values as additional non-carry outputs.

2. **In `outputs_info`**: append `None` for each stochastic carry var's mu output (pytensor collects these but does not feed them back).

3. **After the scan**: use the non-carry mu outputs for `pm.Deterministic` and the Normal likelihood. The carry outputs are consumed only by the scan's own recursion.

4. **Consider removing `carry_innovations` from the estimation model.** During teacher-forcing, innovations are bypassed but MCMC still samples them from `Normal(0, 1)` — harmless but wasteful.

Sketch of the step_fn change:

```python
mu_outputs = {}
for var in endo_keys:
    mu = build_mu(...)
    # ... intercept, slopes ...

    if var in stochastic_carry_vars:
        sigma_val = ns_map[f"sigma_{var}"]
        sampled_state = mu + sigma_val * carry_innov_t[var]
        obs_state = obs_carry_t[var]
        observed_or_sampled = pt.switch(
            pt.isnan(obs_state), sampled_state, obs_state
        )
        new_endo[var] = pt.switch(
            use_observed_carry_t, observed_or_sampled, sampled_state
        )
        mu_outputs[var] = mu  # linear predictor for likelihood
    elif ...:
        ...

out = [new_endo[k] for k in endo_keys]
out += [new_adstock[k] for k in adstock_keys]
out += [exog_t.get(k, pt.zeros(n_units)) for k in exog_lag_bases]
out += [mu_outputs[k] for k in stochastic_carry_vars]  # non-carry
return out
```

After the scan, use the non-carry mu outputs for Deterministic and likelihood, not the carry outputs.

---

### Test corrections needed

**`test_observed_model_carries_realized_lagged_values`** asserts `mu_sales == observed_data`, which validates the buggy behavior. With the correct fix, `mu_sales` should contain the linear predictor. For `sales ~ 1*lag(sales) + 0` with data `[1, 2, 4, 8]` and init carry = 1.0, the expected `mu_sales = [1.0, 1.0, 2.0, 4.0]`, not `[1, 2, 4, 8]`.

**`test_generative_model_scan_mu_is_stochastic_for_endogenous_lag`** checks `std > 0`, which would pass even with double noise. A stronger test would verify variance is consistent with `sigma²`, not `2 * sigma²`.

**Suggested new test:** Verify parameter identifiability — e.g. check that `compile_logp(vars=[sales_rv])` has non-zero gradient w.r.t. beta, or run a short MCMC on a known DGP and verify the AR coefficient is recovered.

---

### Checklist for revision

- [ ] Decouple scan carry state from emitted mu (non-carry scan output)
- [ ] Use linear predictor `mu` for likelihood, not carried state
- [ ] Fix `test_observed_model_carries_realized_lagged_values` to assert `mu_sales` equals the linear predictor
- [ ] Add a test that verifies parameter identifiability (logp gradient w.r.t. beta is non-zero)
- [ ] Verify `TestLagModel::test_ate_positive` passes
- [ ] Run full test suite including slow tests

drbenvincent (commented):
Thanks for the careful review — great catch. I’ve now decoupled scan carry state from emitted  by adding non-carry scan outputs for stochastic-carry vars and using those linear predictors for  + likelihood, while carry outputs are used only for recursion. I also updated regression tests to assert the corrected  behavior and added a beta-sensitivity check for observed . Re-ran targeted lag tests including the previously failing , which now passes.

drbenvincent (commented):
Thanks for the careful review — great catch. I’ve now decoupled scan carry state from emitted mu by adding non-carry scan outputs for stochastic-carry vars and using those linear predictors for mu_* + likelihood, while carry outputs are used only for recursion. I also updated regression tests to assert the corrected mu behavior and added a beta-sensitivity check for observed mu. Re-ran targeted lag tests including the previously failing TestLagModel::test_ate_positive, which now passes.

drbenvincent (commented):
## Re-review after commit bcdace3

All concerns from the initial review are resolved. Ready to merge.

### Verification

**Bug 1 (degenerate likelihood) — fixed.** `mu_all` for stochastic carry vars now comes from the non-carry scan output (the linear predictor), not the teacher-forced carry state. Confirmed empirically: perturbing `beta_sales` by +5 produces a `sales` logp delta of −2183.75 (was 0.0 before the fix).

**Bug 2 (double noise) — fixed.** The linear predictor emitted via `carry_mu` no longer has `sigma * innovation` baked in, so `pm.Normal("sales", mu=mu_all, sigma=sigma)` applies noise exactly once. `TestLagModel::test_ate_positive` now passes (was −1.16e12).

### Test results

- 391 non-slow tests: all pass (0 regressions)
- 16 slow panel engine tests: all pass (including the previously failing `TestLagModel::test_ate_positive`)
- New regression tests verify: `mu_sales` equals the linear predictor (not observed data), mu varies with beta, and generative mu remains the linear predictor without carry noise

### Implementation

The fix is minimal and clean: `carry_mu[var] = mu` captures the linear predictor inside the step function, `outputs_info += [None ...]` emits it as a non-carry scan output, and `mu_all = carry_mu_results.get(var, carry_all)` selects it for the deterministic and likelihood. Carry state continues to use teacher-forcing/stochastic innovations for correct lag recursion.

drbenvincent (commented):
## Test quality follow-up

The implementation is correct, but two of the three regression tests have gaps that should be tightened before merge.

### Test 2 (`test_observed_model_mu_depends_on_beta`) — does not discriminate

This test asserts `std(mu_samples[:, 2, 0]) > 0` to verify that mu depends on beta. However, **this assertion also passes with the original bug** (pre-PR code where carry = mu). With the original deterministic recursion, `mu_t = beta^t * init` — different beta draws still produce different mu values, so `std > 0` holds trivially.

Confirmed empirically: simulating the original bug gives `std(mu[:, 2]) = 3069` vs correct behavior at `19.7`. Both are `> 0`.

**Suggested replacement:** A logp sensitivity check directly tests what matters — that the data likelihood depends on beta:

```python
def test_observed_model_logp_depends_on_beta(self):
    """Observed-data likelihood must be sensitive to regression coefficients."""
    df = self._simple_panel()
    model = pathmc.model(
        "sales ~ lag(sales)",
        data=df,
        panel={"unit": "region", "time": "week"},
    )
    obs_model = model.pymc_model
    sales_rv = next(rv for rv in obs_model.observed_RVs if rv.name == "sales")
    logp_fn = obs_model.compile_logp(vars=[sales_rv])
    ip = obs_model.initial_point()

    point_pert = dict(ip)
    point_pert["beta_sales"] = ip["beta_sales"] + np.array([5.0, 5.0])

    delta = abs(logp_fn(point_pert) - logp_fn(ip))
    assert delta > 1.0, f"sales logp should change with beta, delta={delta}"
```

This would fail with the degenerate likelihood (delta = 0) and pass with the correct fix (delta = 2183).

### Test 3 (`test_generative_model_scan_mu_is_stochastic_for_endogenous_lag`) — misleading name

The test verifies that `mu_sales` is deterministically zero when `coeff=0`, confirming mu is the linear predictor and not the carry state. This is correct behavior to test, but the method name says "stochastic" when it's actually testing determinism. Suggest renaming to `test_generative_model_mu_is_linear_predictor_not_carry` or similar, to match the docstring.

### Test 1 — good as-is

`test_observed_model_carries_realized_lagged_values` uniquely discriminates all three failure modes (original bug → `[1,1,1,1]`, Bug 1 → `[1,2,4,8]`, correct → `[1,1,2,4]`). No changes needed.

drbenvincent:
Addressed the latest test-quality follow-up.

I tightened `tests/test_lag_syntax.py` in two ways:
- replaced the weak `std(mu) > 0` check with a direct observed-logp sensitivity test using `compile_logp(vars=[sales_rv])`, so the test now specifically fails if the observed `sales` likelihood becomes beta-insensitive again
- renamed the generative `mu` test to reflect what it actually verifies: that emitted `mu` is the linear predictor, not the carry state

Re-ran `conda run -n pathmc pytest tests/test_lag_syntax.py -x -v` and all 13 tests passed.

Issue #217, opened 2026-06-10
Pre-existing identification problems exposed by PyMC 6 RNG change: partial-pooling intercepts and residual-cov block wiring


drbenvincent:
## Summary

Four slow tests fail under the PyMC 6 / ArviZ 1 stack (#212), and the failures are **not regressions from the upgrade**: the same models produce equally degenerate posteriors on PyMC 5.28.5 with unmodified `main`. The old fixed seeds simply landed on the lucky side of sign assertions; PyMC 6's different RNG stream flips them. Two distinct root causes, both pre-existing pathmc model-specification problems.

Failing tests:

- `tests/test_panel_do.py::TestTemporalPropagation::test_higher_spend_higher_sales`
- `tests/test_panel_do.py::TestContrastArithmetic::test_contrast_subtraction`
- `tests/test_residual_cov.py::TestBlockVarDoOperator::test_do_on_block_var_mean`
- `tests/test_residual_cov.py::TestBlockVarDoOperator::test_do_on_block_var_predictive`

## Root cause 1: partial pooling includes both a fixed Intercept and random intercepts with a free mean

With `pooling="partial"`, the design matrix keeps the fixed `Intercept` column while `_compile_random_intercepts` adds `alpha_var ~ Normal(mu_alpha, sigma_alpha)` per unit with `mu_alpha` free. The global intercept is therefore unidentified (any constant can shift between `beta[Intercept]` and `mu_alpha`), producing divergences and chain-dependent garbage. Evidence (panel lag fixture from `test_panel_do.py`, seed 42, 200/200 draws): PyMC 5 gives 91+ divergences and `beta_sales[Intercept]` chain means of 8.7 / 3.9; PyMC 6 gives ~200 divergences and means 2.4 / -0.1. Same geometry, different garbage.

Standard fixes: drop the fixed `Intercept` column when random intercepts are enabled (making `mu_alpha` the global intercept), and/or use a non-centered parameterization. Note several gate tests assert `alpha_*` appears in `free_RVs`, which constrains the non-centered option.

## Root cause 2: downstream equations wire through block-variable *predicted means*, not observed components

[...]

Issue #214, opened 2026-06-10
Re-adopt pymc-marketing transform backends once it supports PyMC 6


drbenvincent:
## Summary

As part of #212 (PyMC 6 / ArviZ 1 upgrade + conda removal), the `pymc-marketing` dependency was dropped because no released version supports PyMC 6 / ArviZ 1 (`pymc-marketing` currently pins `pymc<6` and `arviz<1`). The two transform kernels pathmc used — geometric adstock and logistic saturation — are now vendored as plain pytensor implementations in `pathmc/transforms.py` (`_geometric_adstock`, `_logistic_saturation`).

Once pymc-marketing ships PyMC 6 / ArviZ 1 support, decide whether to delegate the built-in transforms back to `pymc_marketing.mmm.transformers` or keep the vendored kernels.

Upstream tracking: pymc-labs/pymc-marketing#2486 (compatibility issue) and pymc-labs/pymc-marketing#2616 (migration PR). Expected upstream ETA: end of July 2026.

## Proposed work

- Watch for a pymc-marketing release with `pymc>=6` support.
- Evaluate whether delegating buys anything beyond the vendored ~30 lines (e.g. additional adstock/saturation variants worth exposing through the transform registry).
- If re-adopting: restore the dependency in `pyproject.toml`, reinstate the xtensor bridging helpers, and remove the vendored kernels. If not: close this issue and keep the vendored implementations as the permanent home.
- Either way, verify numerical equivalence on the transform gate tests (`tests/test_transforms_*.py`) and refresh any affected docs freeze caches.

## Notes

This is not a release blocker for 0.1.0 — the vendored kernels are tested and behaviorally equivalent (truncated-at-`l_max` geometric adstock, same logistic saturation formula).

drbenvincent:
Block check date: 2026-08-07

Determination: still blocked. The blocker has not been resolved under the issue stated condition: pymc-marketing has not yet shipped a released version with PyMC 6 / ArviZ 1 support.

What I checked:
- pathmc issue #214 is OPEN and still carries the exact `blocked?` label.
- Upstream compatibility issue https://github.com/pymc-labs/pymc-marketing/issues/2486 is still OPEN.
- Upstream migration PR https://github.com/pymc-labs/pymc-marketing/pull/2616 merged on 2026-06-26 into `v1.0.0`, and upstream `main` now declares `pymc>=6.0.0` and `arviz>=1.2.0`.
[...]

PR #218, opened 2026-06-10
Drop conda dev env, adopt uv, upgrade to PyMC 6 / ArviZ 1

### LINKED ISSUE #212: Drop conda dev env; adopt uv (bundle with PyMC 6 / ArviZ 1 upgrade)
## Summary

PyMC 6.0 and ArviZ 1.0 are pip-installable, so the historical reason for a conda-forge contributor environment no longer applies. The repo currently presents a split story: end users and CI install via pip (`README`, docs workflow, release workflow), while contributors are directed to conda (`CONTRIBUTING.md`, `AGENTS.md`, `environment.yml` generated by `pyproject2conda`).

Before public release, simplify the development environment to a single pip-first workflow. Prefer adopting [uv](https://docs.astral.sh/uv/) for reproducible lockfiles, but bundle this with the PyMC 6 / ArviZ 1 dependency upgrade so dependency resolution happens once.

## Current findings

- `pyproject.toml` still pins `pymc>=5.27.0,<6` and `arviz<1`; the PyMC 6 / ArviZ 1 migration is prerequisite work for updated pins.
- `environment.yml` is autogenerated from `pyproject.toml` via `pyproject2conda` (pre-commit hook, `make sync-env`, and the `check-environment-yml` CI job).
- `Makefile` `setup` and docs CI already use pip (`pip install -e ".[dev,docs,samplers]"`); conda is not used in any runtime or release path.
- System dependencies (`graphviz` binary, Quarto for docs) remain required regardless of env manager.

## Proposed work

Bundle with the PyMC 6 / ArviZ 1 upgrade (same PR or tightly coupled sequence):

- Bump `pymc` and `arviz` pins and fix any breaking API changes in `pathmc/`.
- Remove conda tooling: delete `environment.yml`, `[tool.pyproject2conda]` from `pyproject.toml`, the `pyproject2conda-yaml` pre-commit hook, `make sync-env`, and the `check-environment-yml` CI job.
- Add uv project files (`uv.lock`, optional `.python-version`) and document `uv sync` as the contributor quick start.
- Update `CONTRIBUTING.md`, `AGENTS.md`, `great-docs.yml` kernel notes, and any skill/docs references that mention conda or `miniforge3`.
- Update CI to use `uv` for dev-env install (or keep plain pip if simpler — either is acceptable if contributor docs match).
- Verify optional extras still resolve on macOS and Linux: `samplers` (jax, nutpie, numpyro), git-pinned `great-docs`, and editable install with all dev extras.
- Re-register the `pathmc` Jupyter kernel after env setup (Quarto / `great-docs freeze` workflow unchanged in intent).

**Fallback if uv adds friction:** drop conda/`pyproject2conda` and document pip-only dev setup (matching current docs CI) without `uv.lock`; add uv in a follow-up. Do not block release on uv specifically — only on removing the misleading conda contributor path.

## Acceptance criteria

- No `environment.yml`, `pyproject2conda` config, or conda references in contributor-facing docs.
- `pyproject.toml` depends on PyMC 6 and ArviZ 1 (or whatever versions we ship for 0.1.0).
- Contributor quick start is a single documented path (`uv sync` or `pip install -e ".[dev,docs,samplers]"`) that matches CI.
- `make setup`, `make test-fast`, `make lint`, and `great-docs build` work from a fresh clone using the new instructions.
- Pre-commit hooks install and run without the removed `pyproject2conda` hook.

## Related

- #174 (closed) — prior major-version pins; this issue supersedes that constraint for the 0.1.0 release.
- #196 — public-release CI; the `check-environment-yml` job should be removed when this lands.
- #199 — test runtime; orthogonal, but update any conda-specific wording in that issue's docs when touched.

drbenvincent:
Closes #212

## Summary

- **Dev environment**: removed all conda tooling (`environment.yml`, `[tool.pyproject2conda]`, the `pyproject2conda-yaml` pre-commit hook, `make sync-env`, the `check-environment-yml` CI job) and adopted [uv](https://docs.astral.sh/uv/) — `uv.lock` + `.python-version` are committed, `make setup` runs `uv sync --all-extras`, and all Makefile targets go through `uv run`. `CONTRIBUTING.md`, `AGENTS.md`, and `great-docs.yml` notes updated to match (the stale Quarto-era docs section in CONTRIBUTING was also brought in line with the Great Docs workflow).
- **Dependency upgrade**: `pymc>=6,<7`, `arviz>=1.1,<2`, `pytensor>=3,<4`, `numpy>=2`. ArviZ 1.1 requires Python ≥3.12, so `requires-python` moves from 3.11 to 3.12 (classifiers now 3.12/3.13).
- **Vendored transforms**: `pymc-marketing` pins `pymc<6` / `arviz<1` in every released version (their migration is pymc-labs/pymc-marketing#2616, unmerged), so the dependency is dropped and the two kernels pathmc used — geometric adstock and logistic saturation — are now plain-pytensor implementations in `pathmc/transforms.py`. Re-adopting pymc-marketing once it supports PyMC 6 is tracked in #214 (upstream ETA ~end of July).
- **ArviZ 1 / DataTree API migration** in source: `az.hdi(hdi_prob=)` → `prob=`, no more `.stack()` on `DataTree` group nodes, `predict()` uses `extend_inferencedata=True` (DataTree has no `.extend`), `fit()` computes the log-likelihood via `pm.compute_log_likelihood()` instead of the deprecated `idata_kwargs` path, and `PathModel.summary()` passes `round_to="none"` so it keeps returning numeric columns.
- **Mechanical test migrations** (approved during the session): new `az.from_dict` signature, `.to_dataset()` before `.stack()`, `ppc.children` instead of `ppc.groups()`, `round_to="none"` in a direct `az.summary` call.

## Known issues — 4 slow tests fail, pre-existing causes (#217)

`TestTemporalPropagation::test_higher_spend_higher_sales`, `TestContrastArithmetic::test_contrast_subtraction` (panel partial-pooling intercept non-identification) and `TestBlockVarDoOperator::test_do_on_block_var_{mean,predictive}` (residual-cov block members wire downstream equations through predicted means, making the parallel-mediators fixture unidentified) fail under the new stack. Verified **not regressions**: the same models produce equally degenerate posteriors on PyMC 5.28.5 with unmodified `main` — the old fixed seeds just landed on the lucky side of the sign assertions, and PyMC 6's different RNG stream flips them. Full diagnosis with evidence in #217; left failing rather than papering over with seed/xfail changes.

## Notes

- The committed `_freeze/` docs cache still holds outputs rendered under PyMC 5. Builds render fine from cache (verified); pages can be re-frozen under the new stack as a follow-up if desired.
- `docs.yml` / `release.yml` CI remain plain pip, which the issue allows; only contributor-facing tooling moved to uv.

## Test plan

- [x] `make test-fast` — 396 passed
- [x] `make test` (slow suite) — all pass except the 4 documented in #217
- [x] `make lint` — all hooks pass
- [x] `uv run great-docs build` — builds successfully from the new env
- [x] Fresh `uv sync --all-extras` resolves on macOS, including `samplers` extra and git-pinned great-docs


juanitorduz:
@bugbot review

pathmc/transforms.py
```diff
@@ -31,34 +31,31 @@
...
-    xx = as_xtensor(x, dims=dims)
-    adstocked = _pmm_geometric_adstock(xx, alpha=alpha, l_max=l_max, dim=dims[0])
-    return adstocked.values
+    w = pt.power(alpha, pt.arange(l_max))
+    result = w[0] * x
+    for i in range(1, l_max):
+        shifted = pt.zeros_like(x)
+        shifted = pt.set_subtensor(shifted[i:], x[:-i])
+        result = result + w[i] * shifted
+    return result
```
daimon-pymclabs, line 59: The loop builds `l_max - 1` separate `set_subtensor` ops into the symbolic graph at construction time. For the typical range (l_max ≤ 16) this is fine. Two things worth noting for the future: (1) if l_max ever grows large, a Toeplitz matrix-multiply or `pt.signal.convolve` would keep the graph O(1) in size; (2) `l_max=0` will `IndexError` on `w[0]` — if the `Adstock` dataclass doesn't already enforce `l_max >= 1`, a guard or assert here would make the failure obvious.

pathmc/transforms.py
```diff
@@ -31,34 +31,31 @@
...
-    """Pointwise logistic saturation: ``(1 - exp(-lam*x)) / (1 + exp(-lam*x))``.
-
-    The transform is elementwise, so the xtensor dim labels are immaterial; the
-    result is unwrapped back to a plain ``TensorVariable``.
-    """
-    dims = tuple(f"d{i}" for i in range(x.ndim))
-    return _pmm_logistic_saturation(as_xtensor(x, dims=dims), lam=lam).values
+def _logistic_saturation(x: Any, *, lam: Any) -> Any:
+    """Pointwise logistic saturation: ``(1 - exp(-lam*x)) / (1 + exp(-lam*x))``."""
+    return (1 - pt.exp(-lam * x)) / (1 + pt.exp(-lam * x))
```
daimon-pymclabs, line 64: `(1 - exp(-lam*x)) / (1 + exp(-lam*x))` is `tanh(lam*x/2)`, range (-1, 1) for all x and [0, 1) for x ≥ 0. This is what pymc-marketing's kernel already computed — the old class docstring (`y = 1 - exp(-lam * x)`) was the error. Good that it's corrected here; behavior is unchanged.

pathmc/model.py
```diff
@@ -565,10 +565,9 @@ def fit(self, **kwargs: Any) -> az.InferenceData:
         assert self._pymc_model is not None
         if sys.platform == "darwin" and "mp_ctx" not in kwargs:
             kwargs.setdefault("mp_ctx", "forkserver")
-        kwargs.setdefault("idata_kwargs", {})
-        kwargs["idata_kwargs"].setdefault("log_likelihood", True)
         with self._pymc_model:
             self._idata = pm.sample(**kwargs)
+            pm.compute_log_likelihood(self._idata, progressbar=False)
```
daimon-pymclabs: `progressbar=False` is hardcoded here. A user who wants the progress bar for the log-likelihood pass can't opt in via `fit(**kwargs)`. Low priority for an alpha, but something to revisit when the public API stabilises.

pathmc/model.py
```diff
@@ -599,9 +598,9 @@ def predict(self, **kwargs: Any) -> az.InferenceData:
             raise RuntimeError(
                 "No posterior samples available. Call .fit() before .predict()."
             )
+        kwargs.setdefault("extend_inferencedata", True)
         with self._pymc_model:
-            pp = pm.sample_posterior_predictive(self._idata, **kwargs)
-        self._idata.extend(pp)
+            pm.sample_posterior_predictive(self._idata, **kwargs)
```
daimon-pymclabs: Silent footgun: if a caller passes `extend_inferencedata=False`, `pm.sample_posterior_predictive` does not mutate `self._idata`, but the method returns `self._idata` anyway — the predictive samples are silently dropped. Either document this in the docstring ("do not pass `extend_inferencedata=False`") or raise early if the caller tries to override the default.

daimon-pymclabs (commented):
Clean, well-scoped migration. The conda→uv switch is straightforward and the PyMC 6 / ArviZ 1 migrations (DataTree API, log-likelihood path, `az.from_dict` signature) are all correct. A few notes below on the vendored kernels and one silent footgun in `predict`.

pathmc/transforms.py
drbenvincent: Good catch on the edge case — fixed in 1498f9799d0a64d2ece281f472a8da66aba57b15: `_geometric_adstock` now raises a ValueError with an actionable message when `l_max < 1` (the `Adstock` class itself doesn't enforce it, and a subclass or future DSL knob could pass a bad value through). On graph size: agreed it's O(l_max) ops, which is fine at the default `l_max=12`; if we ever expose `l_max` to users or need large values, a `pt.signal.convolve`-style O(1) graph is the right move — I'd fold that into the #214 decision about delegating back to pymc-marketing's convolution kernel rather than optimize the vendored one now.

pathmc/transforms.py
drbenvincent: Right — it's `tanh(lam*x/2)`, and the old class docstring was the only thing that was wrong; the kernel behavior is identical to what pymc-marketing computed. I deliberately kept the `(1 - exp(-lam*x)) / (1 + exp(-lam*x))` form rather than rewriting as `pt.tanh` so the vendored kernel stays line-for-line comparable with `pymc_marketing.mmm.transformers.logistic_saturation`, which makes the equivalence check planned in #214 trivial. No code change needed here.

pathmc/model.py
drbenvincent: Fixed in 1498f9799d0a64d2ece281f472a8da66aba57b15: the log-likelihood pass now honors the caller's choice via `kwargs.get("progressbar", False)`, so `fit(progressbar=True)` shows a bar for it and `fit(progressbar=False)` keeps everything quiet. The default stays silent on purpose — before PyMC 6 the log-likelihood was computed inside `pm.sample()` via `idata_kwargs` with no separate bar, so a silent default matches the previous behavior exactly.

pathmc/model.py
drbenvincent: Agreed, that was a silent footgun. Fixed in 1498f9799d0a64d2ece281f472a8da66aba57b15 by going one step further than documenting: `predict(extend_inferencedata=False)` now returns the standalone posterior predictive samples instead of `self._idata`, so nothing is ever dropped — the default path still extends and returns the stored InferenceData. Docstring updated to spell out both behaviors.

pathmc/model.py
```diff
@@ -432,7 +432,7 @@ def summary(self) -> pd.DataFrame:
             raise RuntimeError(
                 "No posterior samples available. Call .fit() before .summary()."
             )
-        return az.summary(self._idata)
+        return az.summary(self._idata, round_to="none")
```
daimon-pymclabs, line 437: test comment on model.py:435

daimon-pymclabs (commented):
test

pathmc/simulate.py
```diff
@@ -344,8 +344,8 @@ def run_do_pymc(
             progressbar=False,
         )
 
-        stacked = idata.posterior.stack(sample=("chain", "draw"))  # type: ignore[attr-defined]
-        n_samples = stacked.sizes["sample"]
+        post = idata.posterior  # type: ignore[attr-defined]
+        n_samples = post.sizes["chain"] * post.sizes["draw"]
```
daimon-pymclabs, line 348: x

daimon-pymclabs (commented):
x

pathmc/simulate.py
daimon-pymclabs, line 348: This n_samples computation is correct. Note that a few lines above (line 377), `pm.sample_posterior_predictive(idata, progressbar=False)` is called without `extend_inferencedata=False`. In PyMC 6, the default is `extend_inferencedata=True`, so `ppc is idata` — the caller's InferenceData gets a `posterior_predictive` group attached as a side effect. Adding `extend_inferencedata=False` there (and at line 570) would prevent this silent mutation and make the intent clear.

daimon-pymclabs (commented):
Second pass — read simulate.py in full, verified the stacked→n_samples migration (clean), and checked DataTree access patterns. Three items:

**1. simulate.py — silent mutation of caller's idata (lines 377 and 570)**
In PyMC 6, `pm.sample_posterior_predictive` defaults to `extend_inferencedata=True`, so `ppc = pm.sample_posterior_predictive(idata, progressbar=False)` mutates `idata` in-place and returns it (`ppc is idata`). A caller that passes `model._idata` to `do()` will find a `posterior_predictive` group silently attached afterward. Tests pass because reading `ppc.posterior_predictive` works either way, but the mutation is unintended. Recommend: `pm.sample_posterior_predictive(idata, progressbar=False, extend_inferencedata=False)` at both call sites.

**2. simulate.py — DataTree variable access pattern (lines 429–430, 434–435, 601–602)**
`ppc.posterior_predictive[var].values` accesses a DataTree node's variable directly. This works because xarray DataTree's `__getitem__` routes non-child-node keys to the node's Dataset. The test changes in this PR use the explicit `idata.posterior.to_dataset().stack(...)` form for the same reason. For consistency: `ppc.posterior_predictive.to_dataset()[var].values` is more explicit and forward-safe. Not a blocker since both forms pass today.

**3. Failing slow tests — need xfail before 0.1.0 release**
The four failures in `test_panel_do.py` and `test_residual_cov.py` leave CI permanently red on the slow suite. Per #217 the root causes are pre-existing model-spec bugs; they should be marked `@pytest.mark.xfail(reason='pre-existing identification issue, tracked in #217', strict=False)` before the 0.1.0 release. That keeps them as canaries without blocking CI. Fixing the underlying specs (#217 root causes) is the right long-term answer.

pathmc/model.py
daimon-pymclabs, line 437: `round_to="none"` is a quiet behavioral change: ArviZ 1.x defaults to `round_to=2` (two decimal places), so any downstream code that expected rounded floats from `.summary()` will see full-precision values instead. This is the right default for programmatic use. Worth a one-liner in the docstring's Returns block: "Columns are full-precision floats; apply `.round()` downstream as needed."

pathmc/simulate.py
drbenvincent: This one's incorrect for PyMC 6 — `pm.sample_posterior_predictive` still defaults to `extend_inferencedata=False` (verified against the installed pymc 6.0.1: `inspect.signature(pm.sample_posterior_predictive).parameters['extend_inferencedata'].default` → `False`; it's `pm.sample_prior_predictive` and our own `PathModel.predict()` that opt in explicitly). So `ppc` here is a fresh DataTree and the caller's idata is never mutated by `do()` — passing `extend_inferencedata=False` explicitly at lines 377/570 would be a no-op. Leaving as-is.

pathmc/model.py
drbenvincent: Agreed it's worth calling out — done in 026b4912c9d217d48d410091b08c23c6db03f6b3: the Returns block now says "Columns are full-precision floats; apply `.round()` downstream as needed." One nuance: ArviZ 1.x's default is `round_to='auto'`, which returns *string-formatted* columns (StringDtype), not rounded floats — that's what broke the `0.3 < theta_mean` comparison in `test_v03_smoke.py` and why `round_to='none'` is needed for any programmatic use, not just a precision preference.

drbenvincent: This looks like a stray test comment — let me know if something substantive was intended here.

pathmc/simulate.py
drbenvincent: Looks like a stray test comment (superseded by the fuller comment below on the same line, answered there).

juanitorduz (approved):
Let's go! IMO we should merge this one and then iterate

PR #229, opened 2026-06-12
Document uv setup in CONTRIBUTING and refresh dev/docs tooling

drbenvincent:
## Summary

This branch started as a CONTRIBUTING update but now also carries a couple of dev/docs tooling fixes that landed on the same branch.

**CONTRIBUTING.md**
- Spell out the `uv` commands for environment setup, testing, and linting rather than hiding them behind `make` aliases, since we switched from conda to uv in #218.
- Explain what `uv sync --all-extras` does (`.venv/`, Python from `.python-version`, editable install, extras) and the `uv run` pattern instead of activating the venv.
- Point contributors to where dependency version constraints live in `pyproject.toml`.

**Dependencies / CI**
- Pin `matplotlib<3.11` to unblock CI: matplotlib 3.11 removed `matplotlib.style.core`, which `arviz_plots<=1.1` imports at module load, breaking the docs build and sdist/wheel import checks. Remove once `arviz_plots` ships a fix.
- Drop the unused `polars` dev dependency (and its mypy override).

**Docs tooling**
- Upgrade `great-docs` 0.13.0 → 0.13.1 in `uv.lock`. This release auto-generates `llms.txt` / `llms-full.txt` during the build, so the docs site now ships LLM-readable API context. The previously-installed dev build (0.10.1.dev from a git pin) predated that feature; the lock and venv are now aligned on 0.13.1.

## Test plan

- [ ] CONTRIBUTING.md read-through; docs-only change, no code affected.
- [ ] CI green with the `matplotlib<3.11` pin.
- [ ] `uv run great-docs build` produces `llms.txt` and `llms-full.txt` under `great-docs/_site/`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,16 +4,19 @@ pathmc welcomes contributions from users, researchers, and developers interested
 
 ## Quick Start
 
-The development environment is managed with [uv](https://docs.astral.sh/uv/); [install it](https://docs.astral.sh/uv/getting-started/installation/) first if you don't have it. After forking this repository on GitHub, get up and running in a few commands:
+The development environment is managed with [uv](https://docs.astral.sh/uv/), a fast Python package and project manager that replaces the old conda-based setup. [Install uv](https://docs.astral.sh/uv/getting-started/installation/) first if you don't have it. After forking this repository on GitHub, get up and running in a few commands:
 
 ```bash
 git clone git@github.com:<your-github-handle>/pathmc.git
 cd pathmc
-make setup
-make test-fast
+uv sync --all-extras
+uv run prek install -f
+uv run pytest -x -v -m "not slow"
 ```
 
-`make setup` runs `uv sync --all-extras` (creating `.venv/` with the package installed in editable mode plus all dev, docs, and sampler dependencies) and installs the pre-commit hooks. Common contributor commands are collected in the root `Makefile`; run `make help` to list the available setup, lint, test, docs, and build targets. The Makefile targets invoke tools through `uv run`, so you never need to activate the virtual environment manually.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -59,36 +63,38 @@ Update an existing environment after pulling changes that touch dependencies:
 uv sync --all-extras
 ```
 
-Edit dependency metadata in `pyproject.toml`. The `uv.lock` lockfile pins exact versions for reproducible contributor environments; `uv sync` updates it automatically after `pyproject.toml` changes, and the updated lockfile should be committed.
+Dependencies and their version constraints are declared in `pyproject.toml` under `[project].dependencies` (the runtime stack, including the `pymc>=6.0,<7` and matching `pytensor>=3.0,<4` pins) and `[project.optional-dependencies]` (the `dev`, `docs`, and `samplers` extras). Edit those lists to add or bump a dependency. The `uv.lock` lockfile then pins exact resolved versions for reproducible contributor environments; `uv sync` updates it automatically after `pyproject.toml` changes, and the updated lockfile should be committed alongside the metadata edit.
 
 Run fast tests only, excluding slow MCMC sampling tests:
 
 ```bash
-make test-fast
```
juanitorduz, line 67: why did you remove the make commands? Shall we keep the make commans and simply update them for uv?

juanitorduz (commented):
QQ

CONTRIBUTING.md
drbenvincent: Sorry, only just saw. Will bring back in a new PR

drbenvincent: Good call — there was no good reason to drop them. The Makefile targets already run `uv run` under the hood, so I've restored the `make` aliases as the primary documented commands (keeping the new explanations of what uv does behind them) in #237.

Issue #230, opened 2026-06-12
Center (or QR-decompose) the design matrix internally to avoid stiff posteriors with non-centered, high-variance predictors


drbenvincent:
## Summary

pathmc currently passes the raw design matrix to PyMC. When a predictor has a large mean and/or large variance, a single regression's posterior is poorly conditioned (a strong intercept↔slope correlation), which forces NUTS into deep trees and slow sampling. Standardising the numerics by **mean-centering** predictors internally (and optionally **QR decomposition** for collinear predictors), then back-transforming so the user-facing coefficients, intercepts, and `do()`/`predict()` results are unchanged, is a well-established remedy (`rstanarm`/`brms` and the Stan manual do this). This issue proposes adding that, **and** flags an open puzzle: in the motivating case, per-equation centering helps isolated regressions but did **not** speed up the full coupled multi-equation model at large `n` — so there is more to understand here.

> [!IMPORTANT]
> **Correction (after testing).** An earlier version of this issue claimed centering was "the clean fix" for the slow `n = 100000` structural model below. That is wrong. Centering dramatically speeds up *isolated single-equation* fits, but on the full coupled 6-equation model it made essentially no difference (~113s centered vs ~110s raw). The intercept↔slope correlation is real and worth fixing, but it is **not** the dominant cost for the coupled, large-`n` model. See "Things that do not fix it" and "Open questions".

## Motivation: a concrete pathology

While building a docs example ("the house" — a 7-node DAG, `docs/examples/the_house.qmd`), fitting the full structural model at `n = 100000` took ~110s, almost all of it inside `pm.sample`, not `compute_log_likelihood` or `do()`. Profiling the model equation-by-equation showed that, for an *isolated* regression, sampling cost is driven by the predictor's location and scale:

| Equation (fit alone) | predictor (mean, std) | mean tree depth |
|---|---|---|
| `X1 ~ X0`, `X2 ~ X1`, `S1 ~ X1` | centered (mean ≈ 0) | ~2.5 (fast) |
| `X3 ~ X2 + X0` | `X2` (2.98, 6.8) | 4.33 |
| `S2 ~ X2` | `X2` (2.98, 6.8) | 3.90 |
| `S3 ~ X3` | `X3` (5.97, 12.4) | 4.46 |

[...]

PR #232, opened 2026-06-12
Add "the house" causal-identification example

drbenvincent:
## Summary

Adds a new worked example, `docs/examples/the_house.qmd`, that solves a causal-inference puzzle [posed by Guénolé Le Pennec on LinkedIn](https://www.linkedin.com/posts/gu%C3%A9nol%C3%A9-le-pennec_causalinference-datascience-statistics-activity-7471165943373717504-Myxa) entirely in pathmc. The puzzle ("the house" — a seven-node DAG) asks which nodes *must* be adjusted for, which *can* be, and which *must* be avoided when estimating the effect of `X1` on `X3`.

The example:

- Encodes the house DAG and reads the answer straight off the graph with `adjustment_sets()`, which returns the unique minimal valid set `{X0}`.
- Explains why the sinks `S2`/`S3` bias the estimate — they are descendants of the mediator/outcome, **not** colliders, so the damage is invisible to `collider_warnings()` but correctly excluded by the backdoor criterion's "no descendants of the treatment" rule.
- Reproduces the puzzle's five `lm()` regressions as one-equation pathmc fits (estimates match the R output line for line: 5.44 / 6.02 / 6.02 / 1.22 / 0.64) and visualises them against the true effect of 6.
- Recovers the effect with `do()`/`ate()` on the full structural model (6.003) with no manual adjustment-set selection.
- Runs `test_implications()` as a DAG sanity check.

It keeps the original `n = 100000` for faithfulness to the puzzle's R code, and includes an honest callout that sampling is slow at this size due to the posterior geometry of the high-variance, non-zero-mean predictors — tracked for follow-up in #230.

## Repo hygiene

- `.gitignore`: ignore in-place Quarto render artifacts (`docs/**/*.html`, `docs/**/*_files/`) and the single-page-freeze `libs/` bundle (`_freeze/**/libs/`). Only `.qmd` sources and the executed `_freeze/` cache (`execute-results/` + `figure-html/`) are tracked, matching the existing convention.

## Test plan

- [ ] `make lint` passes (ruff/ruff-format clean on the new `.qmd`).
- [ ] `uv run great-docs build` renders the Examples section including the new page from the committed freeze cache.
- [ ] Spot-check the rendered page: adjustment set `{X0}`, the five fitted estimates, and `ate ≈ 6`.


juanitorduz:
@bugbot review

PR #244, opened 2026-06-12
clean logo

### LINKED ISSUE #238: pathmc logo
I like the logo. Still, (1) we need a version with better resolution, and (2) we need to fix the ath formulas as they look weird (or remove them).

juanitorduz:
Closes https://github.com/pymc-labs/pathmc/issues/238


juanitorduz:
@drbenvincent, here is an initial suggestion (simple cleanup). Feel free to iterate on this branch if you feel like :)

drbenvincent:
I like it! But if we get a version with a transparent background, then it would look nicer when it's rendered on different backgrounds. Maybe let's just merge when we get a better one and we can always iterate with future PRs

juanitorduz:
> I like it! But if we get a version with a transparent background, then it would look nicer when it's rendered on different backgrounds. Maybe let's just merge when we get a better one and we can always iterate with future PRs

Actually, because of the transparent background and the white font of the "path" part, the current logo in the main does not look good in light backgrounds. I also thik we can merge this one (which will work well in both backgrounds) and iterate

drbenvincent (approved):

PR #245, opened 2026-06-12
Internal quality audit: de-duplicate model.py, add CI tests, type/test hardening

juanitorduz:
## Summary

Internal quality audit of `pathmc` focused on **modularity, type hints, design, and testability**. **No public API change** — `pathmc.__all__` and every `PathModel` public signature are byte-for-byte unchanged; the existing test suite passes unmodified and serves as the contract.

## What changed

**Tooling / CI**
- CI never ran `pytest` (only lint/type checks). Added a `tests` job that runs the **fast suite on PRs** and the **full suite on push to `main`**, across **Python 3.12 and 3.13**.
- Added `pytest-cov` + coverage config (`term-missing`).
- Added `pathmc/py.typed` (PEP 561) so downstream consumers get type info.
- Tightened mypy: `warn_unused_ignores`, `warn_redundant_casts`, `check_untyped_defs`.

**`model.py` de-duplication** (the most repetitive file)
- `_require_fitted()` collapses the **9× repeated** "no posterior samples" preamble into one helper that returns the narrowed `InferenceData`.
- `att()`/`atu()` (~100 near-identical lines each) merged into a shared `_subgroup_effect()`; the public methods are now thin wrappers with unchanged signatures.
- `_run_do()` collapses the repeated `run_do_pymc(...)` boilerplate and removes a **redundant branch in `do()`** (the non-scan time path was byte-identical to the cross-sectional fall-through).

**New `pathmc/idata.py` helpers**
- `posterior()` / `beta_draws()` centralize dynamically-typed ArviZ access, removing ~12 inline `# type: ignore[attr-defined]`.
- `hdi()` centralizes the package-default 0.94 credible mass.

**Cross-module validation helpers**
- `_require_nodes()` (identify.py, graph nodes) and `_require_column()` (panel.py, dataframe columns) consolidate repeated existence checks.

**Type hints**
- `_term_base_vars` / `_scan_term_base_vars` now take a concrete `Term` instead of `Any`.

**Tests**
- `tests/test_idata.py` and `tests/test_internal_helpers.py`: fast, deterministic unit tests for the new helpers and the not-fitted guards.

## API stability
`pathmc.__all__` and all `PathModel` public method signatures are unchanged. All refactors are private internals; existing tests pass without modification.

## Line-count delta
`model.py` drops ~125 net lines (248 changed: ~360 removed / ~235 added) of duplication while preserving behavior.

## Verification
- `uv run prek run --all-files` — ruff + ruff-format + mypy all pass.
- Fast suite green locally (428 passed); full/slow suite delegated to the new CI job.


juanitorduz:
@bugbot review

daimon-pymclabs:
## Review — Daimon

Solid, well-scoped refactor. I read the full diff plus `effects.py`/`simulate.py`/`model.py` on `refactor-audit`. The "no public API change, existing suite is the contract" framing holds up: `_require_fitted`, `_subgroup_effect`, `_run_do`, and the `idata.py`/`_require_nodes`/`_require_column` helpers are all behavior-preserving consolidations, and the new unit tests pin the guards and ArviZ accessors directly. A few things I verified and a few small notes.

### Verified
- **`do()` fall-through is correct.** The removed non-scan-panel branch and the cross-sectional branch were byte-identical `run_do_pymc(...)` calls; both now converge on `self._run_do(set, kind)`. Confirmed the `simulate_over == "time"` + `scan_info is None` path de-dents cleanly to the shared return. No behavior change.
- **No unused `arviz` import.** `az` is still referenced via `az.InferenceData` annotations in both `effects.py` and `simulate.py`, so dropping the inline `az.hdi` calls doesn't orphan the import. `sensitivity.py` correctly removes its now-unused `import arviz as az`.
- **`_require_fitted` preserves error ordering** — `_require_data` runs first, so a data-free model still reports "requires data" before the not-fitted error. `test_data_free_model_reports_missing_data` locks this in.
- **`warn_unused_ignores = true` pairs well** with stripping the ~12 `# type: ignore[attr-defined]` — it'll fail CI if any stale ignore is reintroduced.

### Minor notes (non-blocking)
1. **Error-message specificity regression in `att`/`atu`.** The merged `_subgroup_effect` emits `"Check the subgroup value parameter or data values."`, whereas the originals named the actual public param (`treated_value` / `untreated_value`). A caller of `att()` no longer sees the parameter they'd edit. Cheap to restore by threading the param name through.
2. **Slow/MCMC suite is post-merge only.** The `tests` job runs `-m "not slow"` on PRs and the full suite only on push to `main`, so a regression in a sampling path won't gate the PR that introduces it — it surfaces after merge. Reasonable given MCMC cost, but worth making explicit; a nightly or `run-slow` label-triggered full run would close the gap.
3. **`simulate.py` name overlap.** Module-level `hdi` (imported) coexists with `SimulationResult.hdi`. Not a bug — the method body's bare `hdi(...)` resolves to the global — but mildly confusing; an alias (`from pathmc.idata import hdi as compute_hdi`) would remove the ambiguity.
4. **`_term_base_vars`** swaps defensive `getattr(term, "interaction_of", None)` for direct `term.interaction_of` alongside the `Any -> Term` tightening. Fine since all call sites pass a `Term`; just flagging that the safety net is gone if any dynamic caller ever sneaks in.
5. **Trivia:** `-v` is now set in both `addopts` and the CI `pytest` commands (harmless duplication). `posterior()` returning `Any` intentionally opts those access points out of type-checking — acceptable and documented, just noting the tradeoff is real.

Nothing here blocks merge. (1) and (2) are the only two I'd consider addressing before merge; the rest are cosmetic.

juanitorduz:
Thanks for the thorough review, @daimon-pymclabs! Addressed in 9a85349:

1. **Error-message specificity (att/atu)** — `_subgroup_effect` now takes a `value_param` argument, so the no-matching-observations error again names the public parameter the caller would edit (`treated_value` / `untreated_value`). Tightened the existing fast error-path tests to assert on the param name.
2. **Slow suite gating** — added a nightly `schedule:` cron (06:00 UTC) that runs the full suite (incl. `slow` MCMC tests). The full-suite step condition is now explicit (`push || schedule`); PRs still run fast-only, so PR latency is unchanged. A sampling-path regression now surfaces within ~24h rather than only on the next push to `main`.
3. **`simulate.py` name overlap** — module helper imported as `hdi as compute_hdi`, removing the ambiguity with `SimulationResult.hdi`.
5. **`-v` duplication** — dropped from the CI `pytest` commands (already set in `addopts`).

**Left as-is:** (4) `_term_base_vars` direct `term.interaction_of` — the `Any → Term` tightening was intentional and every call site passes a `Term`, so the `getattr` safety net is deliberately gone; (5)-trivia `posterior()` returning `Any` is an accepted, documented tradeoff.

`prek run --all-files` clean; fast suite green.

daimon-pymclabs:
## Re-review — Daimon

Checked commit `9a85349`. All four points from the prior pass are resolved, cleanly:

1. **Error specificity restored** — `_subgroup_effect` now takes `value_param`, with `att` passing `"treated_value"` and `atu` `"untreated_value"`, so the "No observations" error again names the public parameter. The two `test_*_no_matching_rows_raises` tests assert the specific param text, locking it in (the single-line message means the `.*` in the regex matches fine).
2. **Nightly slow suite** — added a `schedule: cron "0 6 * * *"` trigger and gated the full-suite step on `push || schedule`. Sampling-path regressions now surface daily instead of only on push to `main`. Trigger/step coverage is exact: PR → fast, push/schedule → full, no overlap or gap.
3. **`simulate.py` name overlap** — `hdi` is now imported as `compute_hdi`, removing the shadow with `SimulationResult.hdi`.
4. **Cosmetic `-v`** — dropped from the CI commands; `addopts` retains it.

Nothing outstanding. LGTM.

daimon-pymclabs (approved):
Looks great, @juanitorduz — all four are landed cleanly:

1. **att/atu error specificity** — threading `value_param` through `_subgroup_effect` is exactly right; the error again names `treated_value` / `untreated_value`, and the tightened `match=` assertions lock that in so it can't silently regress. 👍
2. **Nightly slow suite** — `schedule:` cron with the explicit `push || schedule` gate is the clean way to do this. PR latency stays fast-only and sampling-path regressions surface within ~24h. No notes.
3. **`hdi as compute_hdi`** — disambiguation from `SimulationResult.hdi` resolved.
5. **`-v` drop** — good, it was redundant with `addopts`.

Agreed on leaving (4) as-is: every `_term_base_vars` call site passes a concrete `Term`, so the `getattr` net is dead weight given the tightened type — removing it is the right call, not a regression. And `posterior() -> Any` as a documented tradeoff is fine.

Nothing blocking from my side. LGTM.

drbenvincent:
Remote ci tests already done on main

juanitorduz:
> Remote ci tests already done on main

ok! Daimon removed it :)

drbenvincent:
## CI code coverage status (local vs. remote)

Now that #248 has landed on `main`, the full pytest suite + coverage runs in remote CI. I ran the same suite locally for comparison:

| | Local (dev machine) | Remote CI (`main`, post-#248) |
|---|---|---|
| Coverage | **84.87%** | **84.85%** |
| Tests | 543 passed / 0 failed | 543 passed / 0 failed |
| `fail_under` gate (80%) | reached | reached |
| Wall time | ~3.5 min | ~19 min |
| Status | green | green |

The two environments agree to within ~0.02% (a single line measured differently — expected). Both clear the 80% `fail_under` gate comfortably, with ~330 lines (~15%) still uncovered.

**Consequence for this PR:** remote testing and coverage are now owned by `main` (#248). I'm stripping the overlapping bits out of this branch (the `ci.yml` test job, the `[tool.coverage.*]` config, and the `--cov` Makefile changes) so #245 carries only its novel changes: the `model.py`/helpers de-duplication, `idata.py`, `py.typed`, the tightened mypy flags, concrete `Term` typing, and the new unit tests.

juanitorduz:
@drbenvincent, can we merge this one, or is there any blocker?

drbenvincent:
Let's merge

drbenvincent (approved):

Issue #268, opened 2026-06-13
Parallelize tests within a runner with pytest-xdist


drbenvincent:
## Summary

Speed up the remote CI `test` job (and local `make test`) by running the suite in parallel **within a single runner** using [`pytest-xdist`](https://github.com/pytest-dev/pytest-xdist) (`-n auto`). This is the lowest-friction, lowest-cost win for our current scale and complements (does not replace) the cross-runner sharding proposed in #249.

## Why: the suite is compile-bound, not sample-bound

Profiling during #248 (which added the remote `test` job + a coverage gate) established that wall-clock time is dominated by **per-test PyTensor model compilation** (each test builds and C-compiles a PyMC model), not by MCMC iterations. Evidence (full suite, 543 tests):

| Configuration | Local (M4, 10 cores) | CI (`ubuntu-latest`) |
| --- | --- | --- |
| Baseline (default cores, no coverage) | ~135s | — |
| `sample_kwargs` draws=50/tune=50/chains=1/cores=1 | ~128s (~6% faster) | ~18m (no change) |
| `+` coverage (serial) | ~219s | ~18m41s job / ~17m56s pytest |
| **`pytest -n auto`** (no coverage) | **~60s (~2.2x)** | — |
| **`pytest -n auto` + coverage** | **~75s (~2.9x)** | — |

Key takeaways from the session:

- Reducing draws/chains/cores barely moved wall-time (locally ~6%, remotely 0%) because compilation cost is independent of sampler settings. In every serial run `user` time ≈ `real` time, i.e. there was no parallelism to exploit.
- The earlier "260s baseline" was an artifact of the laptop sleeping mid-run; the clean baseline is ~135s.
- `pytest-xdist` is the lever that actually helps, because it compiles/samples many models concurrently across worker processes.

## Proposed change

1. Add `pytest-xdist` to the `dev` optional-dependency group in `pyproject.toml`; run `uv lock` and commit `uv.lock`.
2. Run with `-n auto --dist loadscope` in:
   - `make test` (and optionally `make test-fast`)
   - the CI `test` job in `.github/workflows/ci.yml`
3. Record before/after CI wall-time in the implementing PR.

`--dist loadscope` keeps tests from the same module/class on one worker, which preserves our scoped fixtures in `tests/conftest.py` (`mock_pymc_sample_module`, `mock_pymc_sample_class`, and the `scope="class"` fitted-model fixtures) instead of recomputing them on every worker.

## Compatibility / things verified

[...]

Issue #249, opened 2026-06-12
Parallelize CI pytest with pytest-split


drbenvincent:
## Summary

The `test` job in `.github/workflows/ci.yml` currently runs the full pytest suite in a single GitHub Actions runner (`uv run pytest -x -v`). As the suite grows, wall-clock CI time will be dominated by the slowest sequential run. We should shard the suite across multiple parallel CI jobs using [`pytest-split`](https://github.com/jerry-git/pytest-split), which divides tests into groups with roughly equal total execution time.

## Current state

- **Workflow:** `.github/workflows/ci.yml` — one `test` job, one runner.
- **Command:** `uv run pytest -x -v` (same as `make test`).
- **Suite size:** 543 tests across 38 files under `tests/` (as of 2026-06).
- **Markers:** `slow` is registered in `pyproject.toml` and used for MCMC/integration tests; `make test-fast` runs `-m "not slow"`.
- **Runtime context:** After mock-sampling optimizations (#204), the full suite is ~3 minutes locally on a well-specced Mac; CI on `ubuntu-latest` may be slower. Sharding is worthwhile once per-job setup overhead (~30–60s for `uv sync`) is small relative to test time, and becomes essential if the suite grows again.

## What pytest-split does

`pytest-split` is a pytest plugin that:

1. Reads per-test duration data from a committed JSON file (default: `.test_durations` at the repo root).
2. Assigns tests to `--splits N` groups so each group's **total expected duration** is roughly balanced (`duration_based_chunks` algorithm, the default).
3. Runs only the tests for `--group K` in a given CI job.

Tests missing from `.test_durations` are assigned the suite average duration, so minor test additions work without immediately regenerating the file. Major suite changes (new modules, large fixture refactors) should trigger a refresh.

### 1. Add the dependency

Add `pytest-split` to the `dev` optional dependency group in `pyproject.toml`:

```toml
dev = ["build", "ipykernel", "mypy", "polars", "prek", "pytest", "pytest-split", "ruff"]
```

Run `uv lock` and commit the updated `uv.lock`.

### 2. Generate and commit `.test_durations`

From a clean environment with all extras installed:

```bash
make setup   # or: uv sync --all-extras
uv run pytest --store-durations -v
```

This writes `.test_durations` (JSON mapping `nodeid` → seconds) to the repo root. **Commit this file** — CI shards need it at checkout time.

**Regenerate when:**

- Adding a new test file or many new tests.
- Moving or renaming tests (nodeids change).
[...]

PR #271, opened 2026-06-13
Enable uv cache in CI to speed up dependency install

### LINKED ISSUE #270: Enable uv cache in CI to speed up dependency install
## Summary

Enable the uv cache in CI so the dependency-install step (`uv sync --all-extras`) reuses downloaded/built packages across runs instead of rebuilding the environment from scratch each time.

## Why

Every CI job (`test`, `prek`, `Build Docs`, `Build dists`) currently runs `uv sync --all-extras` cold, costing roughly 40–60s of setup per job. Caching the uv cache directory cuts that to a near-instant restore on cache hits.

Scope note: this speeds up the **install** step only, not test *execution* (the ~17 min `test` runtime is compile-bound — see #268). It's a small, broad, low-risk win that applies to all jobs, which is why it's tracked separately from the test-execution speedups in #268 (`pytest-xdist`) and #249 (`pytest-split`).

## Proposed change

`astral-sh/setup-uv` (already used in every job) has built-in caching. Enable it:

```yaml
      - name: Install uv
        uses: astral-sh/setup-uv@v8.2.0
        with:
          enable-cache: true
          cache-dependency-glob: "uv.lock"
```

`enable-cache: true` caches `~/.cache/uv`; `cache-dependency-glob: "uv.lock"` keys the cache on the lockfile so it invalidates when dependencies change. Apply to all jobs that install with uv.

## Acceptance criteria

- [ ] `enable-cache: true` (keyed on `uv.lock`) set on `setup-uv` across all CI jobs.
- [ ] Cache hit observed on a second run (install step measurably faster).
- [ ] No change to resolved dependency versions (lockfile still authoritative).

## Related

- #268 — `pytest-xdist` (test execution parallelism).
- #249 — `pytest-split` (cross-runner sharding).
- Persisting the PyTensor compile cache is tracked in a separate issue (the higher-leverage, compile-bound lever).

drbenvincent:
## Summary

Enables `astral-sh/setup-uv`'s built-in caching (keyed on `uv.lock`) across the `test`, `prek`, `Build Docs`, and `Build dists` jobs — the exact change proposed in #270:

```yaml
      - name: Install uv
        uses: astral-sh/setup-uv@v8.2.0
        with:
          enable-cache: true
          cache-dependency-glob: "uv.lock"
```

`zizmor` is intentionally excluded — it uses `uvx` (a one-off tool run), not a `uv.lock`-resolved project install, so keying a cache on `uv.lock` there is not meaningful.

## CI duration comparison (with vs without caching)

I measured the dependency-install step on real CI runs of this branch — a **cold** run (cache populated) followed by a **warm** run (cache hit). I tested **both** the cache configuration the issue proposes (`setup-uv` default `prune-cache: true`) and the alternative (`prune-cache: false`) that retains downloaded wheels. The cache-restore/save work happens in the `Install uv` step, so it is reported alongside the install step for a fair, net comparison.

### Config A — issue's proposal (`prune-cache: true`, the default)

`setup-uv` runs `uv cache prune --ci` before saving, which strips re-downloadable PyPI wheels. For pathmc's wheel-heavy dependency set the saved cache is only **~0.2 MB**, so the warm install still re-downloads everything:

| Job | Cold install | Warm install (cache hit) | Δ |
|---|---|---|---|
| Build Docs | 11s | 12s | ~0 |

Cache **hit** confirmed on the second run, but **no measurable install speedup** — the pruned cache contains nothing worth restoring. `Install uv` overhead is also ~0 (0.2 MB).

### Config B — `prune-cache: false` (retains wheels, ~416 MB cache)

| Job | Cold `Install uv` | Cold install | Warm `Install uv` (restore) | Warm install | Net Δ (uv + install) |
|---|---|---|---|---|---|
| prek | 3s | 15s | 14s | 3s | **−1s** |
| test | 1s | 13s | 10s | 3s | **−1s** |
| Build Docs | 1s | 9s | 13s | 3s | **+6s** |
| Build dists* | 1s | 4s | 3s | 5s | +3s |

The dependency-install step itself is **~3–5× faster** (e.g. 15s → 3s), satisfying the acceptance criterion in isolation. **But** restoring the 416 MB cache adds ~10–14s to the `Install uv` step, which cancels the savings — net wall-clock is a wash, and slightly *worse* for the docs/dists jobs.

\* `Build dists` runs `uv build`, not `uv sync --all-extras`, so it never installs the full dependency tree; its cache is tiny and caching does not help it. (Its in-job venv smoke tests use plain `pip`, which `setup-uv` caching does not touch.)

### Takeaway

On GitHub-hosted runners, downloading pathmc's dependencies from PyPI is about as fast as restoring them from the Actions cache, so uv caching yields **no net CI speedup** here regardless of `prune-cache`:

- `prune-cache: true` (this PR): cache hits, but ~0.2 MB cache → install unchanged, ~0 overhead → **net neutral, harmless**.
- `prune-cache: false`: install step 3–5× faster, but ~416 MB cache restore offsets it → **net neutral-to-worse**.

This PR ships **Config A** (the issue's literal proposal) — it is low-risk and harmless, and would start to pay off if dependencies ever shift toward source-built (sdist) packages, where the unpruned-vs-pruned distinction matters less. The `prune-cache: false` experiment was tested and reverted (see commit history) because it does not improve net runtime. The real CI-time levers remain test-execution parallelism (#268, #249) and persisting the PyTensor compile cache.

## Acceptance criteria

- [x] `enable-cache: true` (keyed on `uv.lock`) set on `setup-uv` across all relevant CI jobs.
- [x] Cache hit observed on a second run (verified in logs).
- [ ] Install step *measurably* faster — **not achieved on these runners** (see comparison above).
- [x] No change to resolved dependency versions (lockfile still authoritative; `uv sync` unchanged).

Closes #270.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,11 +26,6 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
-          # Keep downloaded wheels in the cache. The default `uv cache prune
-          # --ci` strips re-downloadable PyPI wheels, which for this
-          # wheel-heavy dependency set leaves a ~0.2 MB cache and no install
-          # speedup; retaining them is what makes warm installs fast.
-          prune-cache: false
       - name: Install dependencies
         run: uv sync --all-extras
       - name: Run tests
@@ -50,11 +45,6 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
-          # Keep downloaded wheels in the cache. The default `uv cache prune
-          # --ci` strips re-downloadable PyPI wheels, which for this
-          # wheel-heavy dependency set leaves a ~0.2 MB cache and no install
-          # speedup; retaining them is what makes warm installs fast.
-          prune-cache: false
       - name: Install dependencies
         # Locked install (same as `make setup`) so the mypy hook sees the
         # project dependencies, matching local `make lint` behavior.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,9 @@ jobs:
           persist-credentials: false
       - name: Install uv
         uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
       - name: Install dependencies
         run: uv sync --all-extras
       - name: Run tests
@@ -39,6 +42,9 @@ jobs:
           persist-credentials: false
       - name: Install uv
         uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
       - name: Install dependencies
         # Locked install (same as `make setup`) so the mypy hook sees the
         # project dependencies, matching local `make lint` behavior.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,11 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
+          # Keep downloaded wheels in the cache. The default `uv cache prune
+          # --ci` strips re-downloadable PyPI wheels, which for this
+          # wheel-heavy dependency set leaves a ~0.2 MB cache and no install
+          # speedup; retaining them is what makes warm installs fast.
+          prune-cache: false
       - name: Install dependencies
         run: uv sync --all-extras
       - name: Run tests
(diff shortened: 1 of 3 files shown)


drbenvincent:
Closing without merging. The benchmark in the PR description shows that enabling `setup-uv` caching delivers **no net CI speedup** for this repo:

- **`prune-cache: true` (proposed config):** cache hit confirmed, but the pruned cache is ~0.2 MB → install time unchanged.
- **`prune-cache: false` (retain wheels):** install step ~3–5× faster, but the ~416 MB cache restore (~10–14s) offsets it → net wall-clock is a wash, slightly worse for docs/dists.

Root cause: on GitHub-hosted runners, PyPI download is about as fast as Actions-cache restore for this wheel-heavy dependency set. Full results and per-job numbers are in the PR description above. Tracking the real levers in #268 / #249 (test-execution parallelism) and the PyTensor compile-cache issue instead.

PR #275, opened 2026-06-15
Add whole-graph DAG falsification (permutation-based test)

### LINKED ISSUE #252: Feature: DAG falsification (whole-graph validation via permuted-baseline test)
## Summary

Port dowhy's DAG falsification (`gcm.falsify_graph`) into pathmc as a whole-graph validation tool that sits on top of the existing conditional-independence (CI) tests.

## What it does

Rather than checking implied conditional independencies one edge at a time, it evaluates the **whole DAG as a unit** and grades it on a curve against randomly-permuted ("rewired") competitor graphs. It answers two distinct questions:

1. **Does the graph break its promises?** Count how many implied CIs the data actually violates (LMC — Local Markov Condition test).
2. **Is the graph informative at all?** Permute the node ordering / edge directions to generate random competitor DAGs and check whether the proposed DAG fits the data meaningfully better than the random pile (tPA — test of permutation/adjacency). If it doesn't beat the shuffles, the data cannot distinguish your arrow directions from noise.

A DAG is "not rejected" only if it both (a) survives the LMC violation check and (b) clearly beats the permuted baseline.

## Why it merits a port

- It's the natural capstone on top of the CI tests pathmc already has (PR #71) — aggregating many edge-level checks into one interpretable verdict.
- It's *falsification*, complementary to the existing `sensitivity.py` (which asks "how strong a confounder would overturn my answer?"). This asks "is my causal story compatible with the data at all?"
- With a full posterior, pathmc can report this distributionally rather than as a single p-value.

## Reference

- dowhy example: https://www.pywhy.org/dowhy/main/example_notebooks/gcm_falsify_dag.html
- Method packaging: https://www.pywhy.org/dowhy/main/user_guide/modeling_gcm/model_evaluation.html

## Notes

Discussed in the technology channel — Carlos (@cetagostini) is scoping this. Distinct from the existing implied-conditional-dependence tests.

cetagostini:
## Summary

Ports dowhy's `gcm.falsify_graph` (Eulig et al., 2023, *Toward Falsifying Causal Graphs Using a Permutation-Based Test*) into pathmc as a whole-graph validation tool that sits on top of the existing conditional-independence machinery in `identify.py`. Closes #252.

Instead of checking one missing edge at a time (`test_implications()`), it grades the **entire DAG** against a baseline of randomly relabeled competitor graphs and answers two questions:

- **Does the graph break its promises?** Counts Local Markov Condition (LMC) violations via the same partial-correlation CI test used by `test_implications()`.
- **Is the graph informative at all?** A node-permutation baseline (tPA) measures how many relabelings share the DAG's Markov equivalence class, and whether the DAG beats the permuted pile on LMC violations.

The verdict follows the reference exactly: `falsifiable` ⟺ `p_tpa ≤ α`; `falsified` ⟺ `p_lmc > α` **and** `p_tpa < α`.

## What's included

- **`pathmc/falsify.py`** — `falsify_graph()` and `FalsificationResult` (with `.falsified` / `.falsifiable`, per-test `.violations`, rich `repr`/HTML, and a histogram `.plot()`). CI tests are memoized across permutations; small graphs (≤ 7 nodes) enumerate all `n!` relabelings exactly, larger graphs sample.
- **`PathModel.falsify(...)`** — works on observed data before sampling, mirroring `test_implications()`.
- **Docs** — new example page `docs/examples/03-identification/dag_falsification.qmd` (with thumbnail + frozen outputs), API reference entries, and `SKILL.md` updates.
- **Tests** — `tests/test_falsify.py` with ~286 cases.

## Notable design decisions

- **Residual covariances (`~~`) are rejected** with a clear error. They encode bidirected/ADMG confounding whose testable implications require m-separation, which is out of scope for a DAG-only permutation test (dowhy is DAG-only too). Falsify the directed structure without `~~`, or use `test_implications()` / `sensitivity()` for confounded pairs.
- **Partial correlation uses effective rank** for degrees of freedom, so constant/collinear conditioners reduce to the correct test.
- **Non-numeric columns** are skipped (treated as unavailable), not crashed on.
- A tiny `significance_level` caps the defaulted `n_permutations` rather than erroring.

## Review process

Developed through an iterative adversarial review loop (adversarial-reviewer, bug-code-review, edge-case-tester subagents). Converged after 3 rounds with all reviewers reporting no actionable findings. Fixes along the way: the verdict boundary (`p_tpa == α`), `n!` enumeration guard, rank-based df, residual-covariance handling, non-numeric columns, and `n_permutations` capping/validation.

## Test plan

- [x] `uv run pytest tests/test_falsify.py` — 286 passing
- [x] `tests/test_identification.py`, `test_sensitivity.py`, `test_smoke.py` — no regressions (334 passing together)
- [x] `make lint` — clean (ruff, ruff-format, mypy, license)
- [x] `uv run great-docs build` — example + reference pages render
- [ ] Maintainer review of the residual-covariance rejection decision

## Note

The final commit (`ebf384d`) adds local `.cursor/` agent tooling per the request to "commit all". It is unrelated to the feature and can be dropped from this branch if you'd prefer a feature-only PR.


drbenvincent (commented):
Thanks, this is in good shape overall. I did a focused pass on the new docs page and a deeper review of `pathmc/falsify.py` against the DoWhy reference behavior.

A couple of non-blocking requests:

- Please make the prerequisite cross-link to the existing conditional-independence notebook more explicit near the first LMC / conditional-independence explanation. The relevant existing page is `docs/examples/03-identification/dag_testing.qmd` ("Testing Your DAG Against Data"), which teaches implied conditional independences and `test_implications()` in detail. This page can then stay focused on the whole-graph permutation baseline rather than reteaching that concept.

- Please reword the "not rejected only if informative" phrasing in the docs and API docs. The implementation correctly returns `falsified == False` for non-informative DAGs too; those DAGs are not rejected, but the conclusion is vacuous because `falsifiable == False`. The sentence appears in the notebook intro and similarly in `pathmc/falsify.py` / `PathModel.falsify()` prose. Suggested framing: an informative DAG that beats the LMC baseline is the positive "not contradicted and testable" case; a non-informative DAG is also not rejected, but should be interpreted as "not falsifiable" rather than evidence for the graph.

Implementation review notes: I did not find a blocking correctness issue in `pathmc/falsify.py`. The parental triple enumeration, symmetric duplicated unconditional triples, `violations / max(1, n_tests)` behavior for skipped tests, p-value inequalities, identity-including permutations, and the residual-covariance rejection all track the DoWhy design or are intentionally documented deviations. I also ran `uv run pytest tests/test_falsify.py -q` locally: `286 passed`.

For the low-resolution frozen PNG, I opened a separate project-level config PR here: https://github.com/pymc-labs/pathmc/pull/288. Once that lands or is rebased in, this PR should refresh and commit the freeze artifact so the histogram PNG is generated at retina resolution.

drbenvincent (commented):
A few additional docs polish requests, separate from the earlier review:

- Code cells that only `print()` values should probably hide the source with Quarto cell options (`#| echo: false`) so readers see the result without a code block that teaches little. The final p-value/verdict print cell is the main example here.

- Please define LMC and tPA in the main text before relying on the abbreviations, or add enough explanation in the relevant figure caption/table text. "Local Markov Condition" appears once, but the abbreviation carries a lot of the example, and "tPA" is especially opaque for readers who have not read the paper.

- For visual clarity, model specifications in the notebook should use multiline strings instead of escaped `\n` sequences. For example, prefer a triple-quoted spec block passed to `pathmc.model(...)` over `"B ~ A\nC ~ A\n..."`.

- Please add Eulig et al. (2023) to `docs/references.bib` and cite it with a proper Quarto citation rather than plain prose. That keeps the example consistent with the rest of the docs bibliography.

drbenvincent:
Add author metadata in the docs yaml header please

```
---
title: "Testing Your DAG Against Data"
author:
  - name: "Carlos"
---
```

PR #290, opened 2026-06-15
Theme-aware transparent logo and favicon

drbenvincent:
## Summary

- Replace the single off-white-background logo with **transparent** dark- and light-ink variants (`logo_light.png`, `logo_dark.png`), extracted from the source via a luminance-to-alpha ramp so the mark drops cleanly onto any background (including the mint navbar gradient).
- Make the logo **theme-sensitive** across all three surfaces:
  - **Navbar** via `great-docs.yml` `logo: {light, dark}`.
  - **Homepage hero** (`00-welcome.qmd`) via two `` tags toggled by the `html[data-bs-theme="dark"]` selector that Great Docs' dark-mode toggle sets.
  - **README** via a `<picture>` element with a `prefers-color-scheme: dark` source (GitHub-native theming).
- Add a **favicon** (`favicon.png`) cropped from a single smiley node onto a rounded white card, so it stays legible at 16–32px on both light and dark browser chrome. Great Docs auto-derives `favicon-16x16/32x32/.ico` from it.
- Remove the now-unused `logo.png` source.
- Refresh the homepage freeze cache (`_freeze/index`) for the new hero markup.

## Notes

- Both logo variants share one luminance-derived alpha mask; only the ink colour differs (charcoal `#262E36` for light mode, soft near-white for dark), so the shapes are pixel-aligned.
- `make lint` passes.

## Test plan

- [ ] `uv run great-docs preview`, toggle dark mode: navbar + hero logos swap; no broken images.
- [ ] Browser tab shows the smiley favicon on both light and dark tab strips.
- [ ] README renders the correct logo in GitHub light and dark themes.


daimon-pymclabs (commented):
Quick review — overall this is clean, tightly scoped, and CI is green (Build Docs, wheel/sdist build, prek all pass; PyPI/Pages jobs correctly skipping). The theming approach is right on all three surfaces: README `<picture>` + `prefers-color-scheme`, navbar `logo: {light, dark}`, and the homepage `html[data-bs-theme="dark"]` toggle. Mergeable.

Two things before merge:

1. **Confirm the deleted `logo.png` is fully de-referenced.** This PR removes `docs/assets/logo.png`, but the refreshed freeze cache (`_freeze/index/execute-results/html.json`) still contains `{width=75% ...}` in its cached markdown, even though source `00-welcome.qmd` now uses the two `` tags. Build Docs passing does **not** clear this — Quarto treats a missing image as a warning, not a build failure, so a dangling reference could ship as a broken hero image. Worth a `grep -rn 'logo\.png' ` across the repo + a glance at the rendered `_site` homepage to confirm it points at `logo_light/dark.png`. If the freeze cache still renders the old ``, it needs a clean re-render rather than just a hash bump.

2. **Test plan is unverified.** All three checkboxes (dark-mode preview, favicon at 16–32px, README in GitHub light/dark) are still unchecked. Since this is purely visual theming, that manual `great-docs preview` toggle is the real verification — CI can't catch a logo that swaps incorrectly or a favicon that's illegible at small sizes.

Neither is a hard blocker; (1) is the only one I'd actually confirm before merging.

PR #294, opened 2026-06-15
Apply review feedback for DAG falsification docs (#275)

cetagostini:
## Summary

Follow-up to the merged DAG-falsification PR #275, applying all of @drbenvincent's review comments (the original PR was merged before they were addressed).

- **Docs prose** (`docs/examples/03-identification/dag_falsification.qmd`): explicit prerequisite cross-link to `dag_testing.qmd` near the first LMC explanation; define **LMC** and **tPA** before the abbreviations carry the example; multiline triple-quoted model specs instead of escaped `\n`; `#| echo: false` on the print-only verdict cell; reworded the "not rejected only if informative" framing so a non-informative DAG reads as *not falsifiable* rather than evidence for the graph.
- **Author metadata**: credit Carlos Trujillo (was set to Ben Vincent in #289).
- **Bibliography**: added `eulig2023falsifying` to `docs/references.bib` and replaced plain-prose mentions with Quarto citations; wired up per-page `bibliography:` frontmatter.
- **Source docstrings**: same reworded framing applied to `pathmc/falsify.py` (module + `falsify_graph`) and `PathModel.falsify()` in `pathmc/model.py`.
- **Freeze cache**: refreshed so the histogram renders at retina resolution (unblocked by merged #288).

## Test plan

- [x] `make lint` — clean (ruff, ruff-format, mypy, license)
- [x] `uv run great-docs freeze docs/examples/03-identification/dag_falsification.qmd`
- [x] `uv run great-docs build` — page renders; "Eulig et al. (2023)" citation links to the References bibliography, author shows "Carlos Trujillo"


daimon-pymclabs (approved):
Reviewed against the two review batches on #275 — every requested change is present and correct.

**Ben's first pass:**
- ✅ Explicit prerequisite cross-link to `dag_testing.qmd` now opens the page, right before the first LMC explanation.
- ✅ "Not rejected only if informative" reworded in the notebook, `pathmc/falsify.py` (module + `falsify_graph`), and `PathModel.falsify()`: an informative DAG that beats the baseline is the positive *"not contradicted and testable"* case, while a non-informative DAG is framed as *not falsifiable* (`falsifiable == False`) rather than evidence for the graph. Matches the implementation's actual `falsified == False` semantics.
- ✅ Freeze artifact refreshed (hash `b78bbf4…` → `c9bf0e7…`) now that #288 landed, so the histogram renders at retina resolution.

**Ben's polish pass:**
- ✅ `# | echo: false` on the print-only verdict cell — freeze confirms the source block is now hidden (consistent with the `# |` convention used elsewhere in the docs).
- ✅ LMC defined ("every variable independent of its non-descendants given its parents") and tPA defined ("test of permutation adjacency") before the abbreviations carry the example.
- ✅ Multiline triple-quoted model specs replace every escaped `"…\n…"` spec.
- ✅ `eulig2023falsifying` added to `references.bib` with arXiv/DOI metadata; plain-prose mentions replaced with `[@eulig2023falsifying]` citations and per-page `bibliography:` frontmatter wired up (path `../../../docs/references.bib` resolves correctly from the example dir).

Author metadata corrected to Carlos Trujillo. No correctness concerns; the source-docstring rewording stays faithful to the decision rule. LGTM.

PR #296, opened 2026-06-17
Remove internal dev notes from public repo; trim AGENTS.md

drbenvincent:
## Summary

- Delete the `docs/dev/` directory (PRD, milestones, audits, brainstorms, and other candid internal notes) now that the repo is public. These move to the gitignored `.scratch/` space as developer-local notes.
- Slim down `AGENTS.md`: drop the stale milestone-driven workflow and all now-unavailable `docs/dev/` references, keeping the constraints that actually prevent breakage.
- Add a note that the package targets **PyMC 6+, ArviZ 1.1+, PyTensor 3+, NumPy 2+** so agents don't assume older APIs, and clarify that `make setup` installs the `prek` git hooks.
- Fix dangling `docs/dev/` comment references in `.github/workflows/docs.yml`.

## Notes

- Nothing sensitive (no secrets/credentials) was committed; this is internal dev documentation. History was intentionally **not** rewritten — the content is non-secret and the repo has been public, so a force-push purge would be disruptive with little benefit.

## Test plan

- [x] `make lint` passes.


drbenvincent:
@bugbot review

PR #297, opened 2026-06-17
Fix PyMC version guard and refresh module map

drbenvincent:
## Summary

- Align the runtime PyMC version check in `pathmc/__init__.py` with the declared `pyproject.toml` constraint (`pymc>=6.0,<7`). The previous guard checked `>=5.22.0`, which no longer matched the real floor, ignored the upper bound, and would crash on non-numeric (pre-release) version segments like `6.1.0rc1`. It now uses `packaging.version` for a robust range check with an accurate, actionable error message.
- Add `packaging` as an explicit dependency, since it is now imported directly (previously only transitive).
- Document the `Prior` re-export as deliberate with a short comment.
- Correct `AGENTS.md`: describe the actual public API (`model()`, `simulate()`, `add_lags()` deprecated, `Prior` re-export) instead of a nonexistent top-level `fit()`, and add the modules missing from the structure map (`falsify`, `sensitivity`, `priors`, `residuals`, `idata`) — several of which are imported directly by tests, so the "these paths are fixed" contract should list them.

## Test plan

- [x] `import pathmc` succeeds on PyMC 6.0.1
- [x] `ruff check` / `ruff format` clean; pre-commit hooks (ruff, mypy, license) pass
- [x] `uv lock --check` confirms the lockfile is consistent after adding `packaging`
- [ ] CI green

## Notes

- `main`'s `AGENTS.md` is the older, pre-cleanup version. The in-flight branch `chore/remove-internal-dev-notes` also edits `AGENTS.md` (trimming dev-notes references). The module-map fix here may conflict with that branch on merge; resolution is straightforward (both edits are additive to the same block).


drbenvincent:
@bugbot review

PR #298, opened 2026-06-17
Align ArviZ usage with 1.0 DataTree philosophy

drbenvincent:
## Summary

ArviZ 1.0 replaced the `InferenceData` class with `xarray.DataTree`: the `az.InferenceData` alias now emits a `MigrationWarning`, and accessing a group as a node attribute (`idata.posterior`) returns a `DataTree` rather than a `Dataset`. This PR aligns pathmc's small ArviZ surface (`idata.py` and its consumers) with that philosophy, while keeping pathmc's deliberate 0.94 HDI policy.

- **Type posteriors as `xarray.DataTree`/`Dataset`** instead of the deprecated `az.InferenceData` alias across `idata.py`, `effects.py`, `simulate.py`, and `model.py` (signatures and docstrings). Dropped the now-unused `import arviz as az` from `effects.py`/`simulate.py`; `model.py` keeps it for `az.summary`.
- **Route group access through `idata["posterior"].dataset`** so callers get a statically typed `Dataset` — the exact input `pm.compute_deterministics` is typed to accept — instead of relying on `DataTree`/`Dataset` API overlap that the migration guide warns is not guaranteed.
- **Refresh the `idata` module docstring** to describe its real ArviZ 1.0 role: typed group access plus pinning `az.hdi` at `prob=0.94` so pathmc stays stable against ArviZ 1.0's changed global defaults (`ci_prob` 0.94→0.89, `ci_kind` hdi→eti). The `hdi()`/`DEFAULT_HDI_PROB` wrapper is *more* warranted under 1.0, not less.
- **`DoResult.hdi`** now reuses `DEFAULT_HDI_PROB` instead of a duplicated hardcoded `0.94`.
- **`tests/test_idata.py`**: annotate the helper as `xr.DataTree` so it no longer evaluates the `az.InferenceData` alias and emits a `MigrationWarning`.

Behavior is unchanged (still a stable 0.94 HDI); this is an idiomatic/typing modernization.

## Test plan

- [x] `make test` passes locally (full suite, including MCMC).
- [x] `tests/test_idata.py` passes with `-W error::arviz.MigrationWarning` (warning fully eliminated).
- [x] `make lint` clean (ruff, ruff-format, mypy, license).
- [x] Verified `pm.compute_deterministics` is typed for `Dataset`, and that `DataTree.__contains__` checks data variables (existing `var in ppc.posterior_predictive` membership checks remain correct).

### DIFF
--- a/pathmc/effects.py
+++ b/pathmc/effects.py
@@ -21,10 +21,10 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
-import arviz as az
 import narwhals.stable.v1 as nw
 import numpy as np
 import pandas as pd
+import xarray as xr
 
 from pathmc.idata import DEFAULT_HDI_PROB, beta_draws, hdi
 from pathmc.parse import Spec
@@ -67,7 +67,7 @@ class EffectResult:
 
 def extract_labeled_draws(
     spec: Spec,
-    idata: az.InferenceData,
+    idata: xr.DataTree,
 ) -> dict[str, np.ndarray]:
     """Extract posterior draws for all labeled coefficients.
 
@@ -75,7 +75,7 @@ def extract_labeled_draws(
     ----------
     spec : Spec
         Parsed model specification with labeled terms.
-    idata : az.InferenceData
+    idata : xarray.DataTree
         Posterior samples from MCMC.
 
     Returns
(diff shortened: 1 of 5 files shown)


drbenvincent:
@bugbot review

Issue #253, opened 2026-06-13
Feature: Effect refutation suite (placebo, random common cause, data subset, dummy outcome)


daimon-pymclabs:
## Summary

Port dowhy's effect refutation suite (`dowhy.causal_refuters`) — a set of model-agnostic robustness checks that perturb the *problem* and confirm the estimated effect behaves as a valid pipeline should.

## What it does

After estimating an effect, run automated stress tests:

- **Placebo treatment** — replace the treatment with random noise; a valid pipeline should return ~0 effect.
- **Random common cause** — add an irrelevant confounder to the model; the estimate shouldn't move materially.
- **Data subset / bootstrap** — re-estimate on random subsamples; the effect should be stable.
- **Dummy outcome** — replace the outcome with noise; the effect should vanish.

## Why it merits a port

- pathmc has rich tooling to *estimate* effects (`effects.py`: ATE/ATT/ATU) but little to *attack* them. This is probably the single clearest gap.
- Each refuter is materially richer with a posterior: a placebo refuter should yield a posterior credibly **straddling zero**, not merely a point estimate near zero — a genuinely Bayesian upgrade over dowhy's point-estimate version.
- Low conceptual surface area, high trust payoff for client deliverables.

## Reference

- https://www.pywhy.org/dowhy/main/user_guide/refuting_causal_effects/index.html

## Notes

dowhy's **add-unobserved-common-cause** refuter overlaps the existing `sensitivity.py` — diff before adding that specific one. The four listed above are the high-value, non-overlapping ports.

PR #310, opened 2026-06-18
Add Bayesian placebo refuter (issue #253)

cetagostini:
## Summary

Ports dowhy's `placebo_treatment_refuter` to pathmc and upgrades it to a fully Bayesian form (issue #253, scoped to the placebo refuter). Instead of dowhy's `num_simulations` point estimates, it runs `J` treatment permutations, re-fits the PyMC model each time, and pools the per-permutation ATE posteriors through a hierarchical normal-normal random-effects null model (the calibration model from Bugaev & Trujillo, *Bayesian Design Analysis for Time-Series Quasi-Experiments*).

- Each permutation is a placebo "fold" contributing a posterior mean `m_j` and SD `s_j`.
- The hierarchical model separates a systematic placebo bias `mu_null` (which a sound pipeline keeps near zero) from the structural volatility `tau_het`.
- The null predictive `theta_new ~ Normal(mu_null, tau_het)` should straddle zero; the real ATE is then calibrated against it with `z_cal` and a one-sided tail probability `p_tail` — the post-estimation calibration step dowhy lacks.

### What's included
- `pathmc/refute.py`: `PlaceboRefutationResult` (verdicts, `repr`/`_repr_html_`/`plot`), `refute_placebo()`, `_permute_and_refit()`, `_fit_hierarchical_null()`.
- `pathmc/model.py`: `PathModel.refute_placebo()` + `_refit_permuted()`; `model()` records construction inputs for faithful refits; `fit()` gains a `compute_log_likelihood` opt-out (refits skip it).
- `pathmc/__init__.py`: exports `PlaceboRefutationResult`.
- Docs: a "Robustness checks" section in the causal-inference user guide + the `pathmc` skill.
- `tests/test_refute.py`.

### Verdicts
- `passes_placebo`: the placebo bias `mu_null` 94% HDI contains zero.
- `effect_survives`: the real ATE's calibrated `p_tail` is below `significance_level`.

### Design notes / deferred
- Default `J = 4` permutations (each a full MCMC re-fit; cost scales linearly).
- One-sided directional `p_tail`, the HDI-contains-zero verdict (no ROPE), and point-`delta_obs` calibration are faithful to the source paper and documented.
- The other dowhy refuters in #253 (dummy outcome, data subset, random common cause) are deferred to follow-ups.

### Reviewer note
`tests/test_public_api.py` is the explicit public-API ledger; it was updated to register `PlaceboRefutationResult` (and the `pathmc.refute` submodule), mirroring how `FalsificationResult` is registered. Flagging since AGENTS.md asks not to modify tests without review.

## Test plan
- [x] `uv run pytest tests/test_refute.py` (40 passed, incl. slow end-to-end)
- [x] `uv run pytest -m "not slow"` full fast suite (719 passed)
- [x] `make lint` (ruff, ruff-format, mypy, license) clean
- [ ] Reviewer: confirm the `test_public_api.py` registry edit
- [ ] Optional: refresh frozen docs (`uv run great-docs freeze docs/user_guide/13-causal-inference.qmd`) before publishing

Hardened through 5 adversarial review cycles (prior-scale collapse, set_priors staleness, non-finite guards, z_cal/p_tail coherence, unconverged-fold warning, etc.).


pathmc/refute.py
```diff
@@ -0,0 +1,732 @@
...
+
+    hier_kwargs = {**sample_kwargs, "random_seed": random_seed}
+    hier_kwargs.setdefault("progressbar", False)
+    if sys.platform == "darwin":
+        hier_kwargs.setdefault("mp_ctx", "forkserver")
+
+    with pm.Model(coords={"fold": np.arange(n_folds)}):
+        obs_sd = pm.Data("obs_sd", fold_sds, dims="fold")
+        mu = pm.Normal("mu_null", mu=0.0, sigma=2.0 * sigma_hat)
+        tau = pm.HalfNormal("tau_het", sigma=2.0 * sigma_hat)
```
daimon-pymclabs: Points 1 & 2: with `n_permutations=4`, `tau_het` is estimated from 4 folds and is prior-dominated — and this `HalfNormal(2 * sigma_hat)` prior is what sets the null spread that both verdicts depend on. The `2x` multiplier is an undocumented magic number doing real work at small J. Please add a one-line rationale for the `2x` (or expose it), and see the line comment on the docstring re: the default `n_permutations`.

pathmc/refute.py
```diff
@@ -0,0 +1,732 @@
...
+    outcome : str
+        Outcome variable name.
+    treatment : str
+        Treatment variable to replace with a placebo. Must be an observed
+        (non-latent) data column.
+    values : tuple[float, float]
+        ``(lo, hi)`` intervention values for the ATE contrast, matching
+        :meth:`pathmc.PathModel.ate` (default ``(0.0, 1.0)``).
+    n_permutations : int
+        Number of placebo permutations / folds (default 4, the practical
```
daimon-pymclabs: Point 1: "the practical minimum for identifying `tau_het`" overstates what 4 folds buy — a variance component from 4 points is identified in name only and the HDI / `p_tail` end up prior-driven. Suggest bumping the default to ~8-10, or rewording to "4 is a floor; the null spread is prior-dominated until ~8+ folds."

pathmc/refute.py
```diff
@@ -0,0 +1,732 @@
...
+    null_sd = float(np.std(theta_new_draws))
+    mean_within_var = float(np.mean(fold_sds**2))
+    sigma_pred = float(np.sqrt(null_sd**2 + mean_within_var))
+    observed_mean = float(np.mean(observed_draws))
+    z_cal = (observed_mean - null_mean) / sigma_pred if sigma_pred > 0.0 else 0.0
+
+    noise_rng = np.random.default_rng(hier_seed)
+    noise_sd = noise_rng.choice(fold_sds, size=theta_new_draws.size)
+    pred_null = theta_new_draws + noise_rng.normal(0.0, noise_sd)
+    if observed_mean >= null_mean:
```
daimon-pymclabs, line 834: Point 3: the tail side is selected by `sign(observed_mean - null_mean)`. Since `mu_null` straddles zero, that's effectively `sign(observed effect)` — a directional test in the effect's own direction. Fine and consistent with the calibration paper, but the docstring's "one-sided" wording doesn't flag that the side is chosen post-hoc. One clarifying sentence would prevent a reader mistaking it for a pre-registered side.

pathmc/_model.py
```diff
@@ -1465,6 +1604,15 @@ def model(
         latent=latent_set,
         priors=priors,
     )
+    path_model._construction = {
+        "spec_string": spec_string,
+        "families": families,
+        "panel": panel,
+        "pooling": pooling,
+        "latent": latent,
+        "priors": priors,
```
daimon-pymclabs, line 1637: Minor: this `"priors"` key is never read — `_refit_permuted` deliberately uses `self._priors` (the merged set, to honor `set_priors()`) instead. Either drop the key or add a comment that it's intentionally unused, so a future reader doesn't "fix" the refit to consume the stale original arg.

daimon-pymclabs (commented):
## Deep review — Bayesian placebo refuter

Strong, careful PR. The API mirrors `falsify()`/`sensitivity()` faithfully, the construction-replay is sound, and the guard/error surface and test coverage are thorough. I cross-checked every external API call against the current pathmc interfaces and found **no breaking bugs**: `DoResult.draws(var)` is a method, `ate(..., values=)` exists, `model()` accepts every replayed construction kwarg, and `merge_priors(defaults, self._priors)` is idempotent on permuted data (a permutation preserves marginals, so data-derived defaults are unchanged). CI build/docs/lint are green; the `test` job is still pending — confirm it passes before merge.

The substantive comments are statistical, not mechanical. Concrete requests below; inline anchors on the relevant lines.

### Requested changes (statistical)

**1. Reconsider `n_permutations=4` default, or soften the docstring claim.** A `HalfNormal` variance component (`tau_het`) estimated from 4 folds is prior-dominated. Both verdicts depend on it — `passes_placebo` reads the `mu_null` HDI (width set by `tau_het`) and `p_tail` convolves `null_sd` (≈ `tau_het`) — so at the default, `sigma_hat` plus the `2x` multiplier largely *are* the answer. The docstring calls 4 "the practical minimum for identifying `tau_het`"; I'd argue 4 identifies it in name only. **Request:** bump the default to ~8-10, or change the doc to "4 is a floor; the null spread is prior-dominated until ~8+ folds."

**2. Document (or expose) the `2 * sigma_hat` prior scale.** `sigma_hat = sqrt(between_var + within_var)` is computed from the folds and then used as the prior scale for the fit on those same folds. Standard empirical Bayes, and the *total*-variance choice is correct (the between-fold-collapse reasoning in the docstring is a genuinely nice catch). But the `2x` constant does real work at small J and is currently an undocumented magic number. **Request:** one-line rationale, or expose it as a kwarg.

**3. Clarify that `p_tail` is directional with a data-selected side.** The tail side is picked by `sign(observed_mean - null_mean)`. Since `mu_null` straddles zero by construction, that's effectively `sign(observed effect)` — a directional test in the estimated effect's own direction. Defensible and consistent with the calibration paper, so **not** a bug, but the docstring says "one-sided" without noting the side is chosen post-hoc. **Request:** one clarifying sentence so a reader doesn't mistake it for a pre-registered side.

**4. Note the calibration asymmetry in the user guide.** Only `observed_mean` enters `z_cal`/`p_tail`; the real effect's posterior spread is discarded (documented as faithful to the source paper). Worth surfacing for client work: the null gets a full predictive distribution while the real effect is reduced to a point, so a wide-but-large ATE survives on its mean alone. **Request:** a sentence in the Robustness-checks guide section.

### Minor / cleanup
- **Dead field:** `_construction["priors"]` is stored but never read (`_refit_permuted` deliberately uses `self._priors`). Drop it or comment why it's retained, so nobody "fixes" the refit to consume it.
- **`model()` silently drops `**kwargs`** and `_construction` doesn't capture them — pre-existing, but replay makes a model built with stray kwargs refit differently-than-built. Low priority; consider asserting-empty or documenting.
- **`_require_fitted` runs twice** (method wrapper + module function). Harmless redundancy.
- **Sequential refits** — J full MCMC fits in a Python loop; fine at J=4, but if the default rises (point 1), note the linear cost and that folds are embarrassingly parallel.
- **`cast("matplotlib.figure.Figure", ax.get_figure())`** can mask a `None` return in newer matplotlib. Cosmetic.

### Done well (so it isn't lost)
Non-centered parameterization; `theta_new` forward-sampled *after* `pm.sample` for the null predictive; `mean(s_j^2)` vs `(mean s_j)^2` coherence between `sigma_pred` and the bootstrap; the SD floor enforced in both caller and callee; the non-convergence outlier warning; and the AGENTS.md-flagged public-API ledger edit — all correct.

**Bottom line:** approve-with-nits. Nothing blocks merge except the pending `test` job. The one change I'd genuinely advocate for is the `n_permutations` default/claim (point 1); the rest is documentation and cleanup.

cetagostini:
Thanks for the careful review, @daimon-pymclabs — addressed in 4a585ee.

**Statistical**
1. **`n_permutations=4` claim** — fixed the wording everywhere (function + `PathModel.refute_placebo` docstrings and the user guide): 4 is now described as a *floor* where `tau_het` is a variance component estimated from a handful of points and is **prior-dominated**, with a recommendation to use 8+ for a data-driven null spread. I kept the **default at 4** (deliberate cost trade-off — each fold is a full MCMC re-fit), rather than bumping it, but the docs no longer oversell it. Happy to raise the default if you'd prefer that over the doc fix.
2. **`2 * sigma_hat` prior scale** — documented: the factor of 2 follows Röver (2021)'s weakly-informative heterogeneity prior (HalfNormal scale = 2× the empirical SD), with a note that it does real work at small J. Left it as a constant rather than a kwarg to keep the surface small; can expose it if you want it tunable.
3. **`p_tail` directional side** — docstring now states the side is selected *post hoc* from `sign(observed_ate - null_mean)` (≈ sign of the observed effect) and is **not** a pre-registered one-sided test.
4. **Calibration asymmetry** — added a sentence to the Robustness-checks guide section: full predictive null vs the real effect reduced to its posterior mean, so a wide-but-large ATE survives on its mean alone (inspect `observed_ate_hdi`).

**Cleanup**
- `_construction["priors"]` now carries a comment that it's intentionally unused (`_refit_permuted` rebuilds with the current merged `self._priors` to honor `set_priors()`).
- Left as-is per your "non-blocking": the double `_require_fitted` (harmless), `cast(ax.get_figure())` (matches the existing `falsify`/`sensitivity` convention), and `model()` dropping `**kwargs` (already a no-op).

Lint/build/docs green; full `tests/test_refute.py` (44) passes locally. Will confirm the `test` CI job before marking ready.

daimon-pymclabs:
Thanks @cetagostini — this all looks good, and the honest small-J framing is exactly what I was after.

On your two open questions:

1. **`n_permutations` default** — keep it at **4**. The cost argument (a full MCMC re-fit per fold) is the right call for a default, and now that the docstrings frame 4 as a prior-dominated floor with an explicit "use 8+ for a data-driven null spread" recommendation, the default no longer misleads. No need to bump it.

2. **`2 * sigma_hat` prior scale** — leaving it as a documented constant is fine. The Röver (2021) citation does the justification work, and keeping the surface small is worth more than the tunability here. If a real use-case ever needs to override it we can expose a kwarg then; no need to pre-empt that.

Agreed on leaving the non-blocking cleanups as-is. Once the `test` CI job is green this is good to mark ready from my side.

daimon-pymclabs:
## Follow-up review (2026-06-22) — prior comments addressed

I re-reviewed after the `Address PR review` commit (`4a585ee1`) and today's merge + `EstimandResult` accessor change (`1c845e9a`). **All four of my original inline comments and the calibration-asymmetry note are resolved:**

- `2×` prior scale — refactored to a named `prior_scale` with a Röver (2021) rationale comment. ✅
- `n_permutations=4` "practical minimum" claim — reworded everywhere to "4 is a floor; prior-dominated until ~8+ folds." ✅
- `p_tail` post-hoc direction — docstring now states the side is selected post hoc and is not a pre-registered one-sided test. ✅
- Dead `_construction["priors"]` key — comment added explaining it's intentionally unused on refit. ✅
- Calibration asymmetry (full null vs point real effect) — new user-guide paragraph pointing readers to `observed_ate_hdi`. ✅

I also verified the `.draws(outcome)` → `.draws()` change is behavior-preserving: `EstimandResult.draws(var=None)` resolves to `self._default_var`, set to `outcome` at construction (`simulate.py:234`, `:269`). No regression.

### Two small nits remaining (both non-blocking, documentation/cleanup)

**1. `passes_placebo` is biased toward PASS at small J — the flip side of the prior-domination you just documented.** The verdict is a pure interval rule (`mu_null` HDI contains 0). At J=4 the prior inflates that HDI wide, so it nearly always straddles zero, meaning a genuinely broken pipeline carrying a real placebo bias can be masked as PASS. The existing `.. note::` on `passes_placebo` covers only the false-*FAIL* direction (a tiny-but-precise bias failing the interval). A sentence noting the false-*PASS*-at-small-J risk would make the small-J caveat symmetric.

**2. `sigma_pred` formula is duplicated.** It's computed inline in `refute_placebo` (`refute.py:808`) for the stored `z_cal`, and again in the `sigma_pred` property (`refute.py:219`). They agree today, but it's a drift risk if one is edited — consider having the property be the single source the inline computation reuses.

### Otherwise

Test coverage is strong (44 tests: verdicts, every guard/raise path, plot kinds, the heterogeneous-fold warning, log-likelihood stripping, integration fits). Core mechanics are correct — permutation severs the treatment cleanly, merged priors are honored on refit, and per-fold seeds are independent.

**Bottom line: approve-with-nits**, both nits are docs/cleanup. Only open item is confirming the `test` CI job lands green (currently pending after today's push).

juanitorduz:
@cetagostini , I think once you address the small comments we could merge :)

drbenvincent:
~~I'll run a re-freeze to get updated docs and push to this branch - but will wait for @cetagostini's updates~~

Actually, I do that in a separate PR just before I we release 0.1.1

cetagostini:
Thanks @daimon-pymclabs, @juanitorduz — both follow-up nits addressed in 03c424e.

1. **`passes_placebo` false-PASS asymmetry** — extended the `.. note::` to cover the symmetric small-J failure mode: with few folds the prior dominates `tau_het` and inflates the `mu_null` HDI so it almost always straddles zero, which can mask a genuinely biased pipeline as PASS. It now points readers to use 8+ folds before trusting a PASS.
2. **Duplicated `sigma_pred`** — extracted a single `_predictive_null_sd(null_sd, fold_sds)` helper; both the inline `z_cal` calculation in `refute_placebo` and the `sigma_pred` property now call it, so they can't drift.

Lint/mypy green; `tests/test_refute.py` (44) passes locally. Ready from my side pending the `test` CI job.

cetagostini:
> passes_placebo false-PASS asymmetry — extended the .. note:: to cover the symmetric small-J failure mode: with few folds the prior dominates tau_het and inflates the mu_null HDI so it almost always straddles zero, which can mask a genuinely biased pipeline as PASS. It now points readers to use 8+ folds before trusting a PASS

I have a challenge for this not a blocker now. Will validate this with math but pretty sure is daimon being a bit to extreme. but not a blocker.

juanitorduz (approved):
Thanks @cetagostini !

drbenvincent (changes requested):
## Documentation Review

This PR introduces excellent functionality but has **two blocking issues** that must be addressed before merge:

### ❌ Blocking Issue 1: Missing API Reference Entry

`PlaceboRefutationResult` is exported in `pathmc/__init__.py` but is **not included** in the API documentation.

**Fix required:** Add to `great-docs.yml` lines 102-110 under "Result objects":

```yaml
  - title: Result objects
    desc: Result containers returned by PathModel methods for programmatic inspection.
    contents:
      - simulate.DoResult
      - simulate.EstimandResult
      - effects.EffectResult
      - identify.ImplicationTestResult
      - sensitivity.SensitivityResult
      - falsify.FalsificationResult
      - refute.PlaceboRefutationResult  # <-- ADD THIS LINE
```

After adding, rebuild docs with `uv run great-docs build` and verify `great-docs/_site/reference/refute.PlaceboRefutationResult.html` exists.

### ❌ Blocking Issue 2: Zero Examples

There are **zero demonstrations** of this functionality in any of the 20 example notebooks under `docs/examples/`.

**Action:** I'm creating a separate issue to track adding example(s). This is critical for user adoption but can be addressed post-merge via the issue.

---

### ✅ What's Good

1. **User guide documentation** (`docs/user_guide/13-causal-inference.qmd`) is clear, technically accurate, and well-structured
2. **SKILL.md** is properly updated
3. Code is well-documented with comprehensive docstrings in `refute.py`
4. Proper integration with existing robustness-check ecosystem (`falsify()`, `sensitivity()`)

### Optional Improvements (Non-blocking)

Minor wording suggestions in the user guide:
- Line 306: "dowhy's refuters" → "DoWhy's refuters" (consistent capitalization)
- Line 308: "Four folds" might confuse with cross-validation; consider "Four permutations"
- Line 310: "worth knowing for client work" is oddly specific; consider "Note:" or "Technical detail:"

---

**Resolution:** Fix the API reference entry (Issue 1), then this can merge. Examples (Issue 2) will be tracked separately.

cetagostini:
@drbenvincent I corrected all your points, example to come later on. Docs are there still.

cetagostini:
Thanks @drbenvincent — documentation review addressed.

**Blocking Issue 1 (API reference):** Added `refute.PlaceboRefutationResult` under "Result objects" in `great-docs.yml`. Rebuilt with `uv run great-docs build` and confirmed `great-docs/_site/reference/refute.PlaceboRefutationResult.html` is generated.

**Optional wording (user guide):** applied all three — "DoWhy's" capitalization, "Four permutations" (avoids cross-validation "folds" confusion), and softened the calibration-asymmetry sentence to "Note:".

**Blocking Issue 2 (examples):** understood — leaving for the separate issue you're tracking, to be added post-merge.

Also merged your `main` sync (#339) into the branch; lint green and `tests/test_refute.py` (44) + the new `test_partial_pooling_warning.py` pass locally. Branch is in sync.

drbenvincent:
FYI: Going to release 0.1.1 now. We can merge this to main _after_. So today, no more blockers from me other than getting 0.1.1 out

cetagostini:
> FYI: Going to release 0.1.1 now. We can merge this to main _after_. So today, no more blockers from me other than getting 0.1.1 out
> 

Can we merge now?

PR #318, opened 2026-06-19
fix: silently ignore explicit `1 +` intercept in spec formula

### LINKED ISSUE #314: does spec allow intercept
what is the right way to allow an intercept, the below generates a key error

spec = """
M ~ 1 + a*X
Y ~ 1 + b*M + c*X
indirect := a*b
"""

### LINKED ISSUE #317: Parser should handle explicit `1 +` intercept syntax gracefully
## Problem

Users coming from R/lavaan/statsmodels expect to write `Y ~ 1 + X` to include an intercept. In pathmc, intercepts are implicit (included by default), and the `1 +` syntax causes an unhelpful `KeyError` because the parser tries to look up `1` as a variable name.

Reported in #314.

## Expected behavior

The parser should detect `1` as a term and either:
- **Silently ignore it** (since the intercept is already included), or
- **Raise a helpful `ParseError`**: e.g. `"Intercepts are included by default in pathmc — remove '1 +' from 'Y ~ 1 + X'. To suppress the intercept, use '0 +'."`

The second option is preferable — it teaches the user the pathmc convention rather than silently accepting foreign syntax.

## Scope

Small change in `pathmc/parse.py`, in the term-parsing logic of `_parse_regression()`. Detect when a term is the literal string `"1"` and handle accordingly.

daimon-pymclabs:
## Summary

- Parser now silently skips `1` as a term in regression formulas, matching how `0` is already handled for intercept suppression
- Users from R/lavaan who write `Y ~ 1 + X` no longer get a `KeyError`
- Added parametrized tests covering `1 +` with plain variables, labeled coefficients, and multi-equation specs

Two-line fix in `parse.py`, 15-line parametrized test addition.

Closes #317
Closes #314

## Test plan

- [ ] Existing `test_default_intercept` still passes (implicit intercept unchanged)
- [ ] Existing `test_intercept_suppression` still passes (`0 +` still works)  
- [ ] New `test_explicit_intercept_ignored` passes for all three parametrized formulas
- [ ] `1` does not appear as a variable in any parsed terms

### DIFF
--- a/pathmc/parse.py
+++ b/pathmc/parse.py
@@ -218,6 +218,8 @@ def _parse_regression(stmt: str) -> Regression:
         if raw == "0":
             has_intercept = False
             continue
+        if raw == "1":
+            continue
         terms.append(_parse_term(raw))
 
     if not terms:
--- a/tests/test_parse.py
+++ b/tests/test_parse.py
@@ -61,6 +61,21 @@ class TestRegressionParsing:
         assert "X1" in variables
         assert "X2" in variables
 
+    @pytest.mark.parametrize(
+        "formula",
+        [
+            "Y ~ 1 + X1 + X2",
+            "Y ~ 1 + a*X",
+            "M ~ 1 + a*X\nY ~ 1 + b*M + c*X\nindirect := a*b",
+        ],
+    )
+    def test_explicit_intercept_ignored(self, formula):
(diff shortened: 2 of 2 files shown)


drbenvincent:
@bugbot review

drbenvincent (approved):

Issue #319, opened 2026-06-19
Add EstimandResult for ate/cate/att/atu and __repr__ for DoResult


daimon-pymclabs:
## Problem

`ate()`, `cate()`, `att()`, `atu()` all return `DoResult`, which has no `__repr__` and requires the user to re-specify the outcome variable to get a number (`result.mean("Y")`). This is confusing for new users (#315) and inconsistent with every other Result type in pathmc, all of which have `__repr__`.

### Two result types, one per intent

**`DoResult`** (returned by `do()`) — "what happens to the whole system under this intervention?" No single privileged outcome. Holds draws for all variables.

**`EstimandResult`** (new, returned by `ate()`, `cate()`, `att()`, `atu()`) — "what is the effect of X on Y?" Knows its outcome variable. Holds the same draws but provides a focused API.

Both hold `dict[str, np.ndarray]` internally. The difference is that `EstimandResult` knows what question was asked.

### What changes

**`DoResult` gets a `__repr__`:**
```
DoResult (2000 draws, 5 variables)
              mean   94% HDI
  X           1.00   [1.00, 1.00]
  M           3.45   [2.81, 4.12]
  Y           7.21   [5.93, 8.44]
```

**`EstimandResult` is a new class with:**

```python
result = m.ate("Y", "X", values=(0, 1))

# Notebook repr:
#   ATE of X on Y
#     Mean:    1.23
#     94% HDI: [0.87, 1.61]
#     P(> 0):  0.97
#     Draws:   2000

result.mean()              # defaults to outcome — no arg needed
result.hdi()               # defaults to outcome
result.draws()             # np.ndarray (n_samples,) for outcome
result.prob("> 0")         # P(ATE > 0)
result.summary()           # tidy DataFrame row
float(result)              # posterior mean via __float__

result.mean("M")           # still works for other variables
```

### Per-unit draws (stretch goal)

Currently, draws are averaged over units during `do()` computation. Instead, store the full tensor internally (xarray with named dims: `sample`, `unit`, optionally `time`) and summarize lazily. Default accessors give population-level summaries. An opt-in accessor exposes unit-level detail:

```python
result.draws()                  # (n_samples,) — population
result.draws(per_unit=True)     # (n_samples, n_units) — unit-level
```

### Implementation notes

- `prob()` uses an explicit method, not operator overloading — `result > 0` returning a float breaks Python `__gt__` semantics
[...]

Issue #315, opened 2026-06-19
m.ate("Y", "X", values=(0, 1)).mean('X')


dirknbr:
what is the right way to get a numerical answer to the ATE question

m.ate("Y", "X", values=(0, 1)) returns an object, not a number 

m.ate("Y", "X", values=(0, 1)).mean('X') seems to return 1 exactly 

m.ate("Y", "X", values=(0, 1)).mean('Y')  returns a large number which seems to be an over estimate of the effect

drbenvincent:
Thanks! This has highlighted the result you get back is not as user friendly as it could be. I'm going to redesign this, see #319.

daimon-pymclabs:
Thanks for raising this — the API here could definitely be clearer, and we're tracking improvements in #319.

To explain what you're seeing:

**`.mean('X')` returning 1 exactly** — this is expected. `ate()` computes `do(X=1) - do(X=0)`, so the result is a contrast. The `X` column in that contrast is literally `1 - 0 = 1`. It's not telling you anything about the effect — it's just echoing back the intervention gap.

**`.mean('Y')` is the value you want** — that's the posterior mean of the average treatment effect of `X` on `Y`. Whether the number is "too large" depends on your data and model. Without seeing the full spec, data, and model fit we can't tell whether there's a bug or whether the estimate is simply reflecting the model you specified. If you can share a minimal reproducible example (spec + data generation or a small dataset), we can take a look.

The general workflow is:

```python
result = m.ate("Y", "X", values=(0, 1))
result.mean("Y")    # posterior mean of the ATE
result.hdi("Y")     # 94% highest density interval
result.draws("Y")   # raw posterior draws if you want to plot/inspect
```

We agree the API is more verbose than it needs to be — you already told `ate()` that the outcome is `Y`, so having to repeat it in `.mean("Y")` is redundant. #319 tracks a redesign to make this simpler.

drbenvincent:
Closed by #319 I believe. If not, happy to re-open.

PR #320, opened 2026-06-19
Add EstimandResult for ate/cate/att/atu and __repr__ for DoResult

drbenvincent:
Part of #319 (does **not** fully close it — see Deferred below).

## Summary

- `ate()`, `cate()`, `att()`, `atu()` now return a new `EstimandResult` that knows its outcome variable, so `.mean()`, `.hdi()`, and `.prob()` need no argument; `float(result)` gives the posterior mean; printing shows a tidy summary. (Accessors still accept a variable name internally, but the docs now show the intended no-arg form.)
- `DoResult` gains a tabular `__repr__` (per-variable mean and 94% HDI), matching every other pathmc result type.
- New `EstimandResult.prob("> 0")` reports the posterior probability the estimand clears a threshold (e.g. `P(ATE > 0)`), distinct from the predictive `PathModel.prob()`.
- User guide + package skill updated to distinguish `DoResult` (whole system) from `EstimandResult` (focused effect) and to highlight `prob()`. `EstimandResult` added to the API reference; all result classes consolidated under one "Result objects" heading.
- **Example notebooks updated to the intended API**: every `ate()/cate()/att()/atu()` call now uses outcome-default accessors, and the foundational `causal_do_operator.qmd` showcases the repr, `float()`, `.prob("> 0")`, and `.summary()`. `do()` contrasts deliberately keep the explicit variable (no privileged outcome).

This addresses the new-user confusion in #315.

## Deferred (stretch goals from the issue — issue intentionally kept open)

The internal xarray named-dim refactor and full per-unit draw storage (`per_unit` / `by_unit()`) are **not** in this PR — they need deep changes to the `do()` engine, which currently pools draws across units. Recommend a follow-up issue. The `summary()` output stays pandas, matching `effects_summary()` / `standardized()` / `to_dataframe()` (narwhals is used for input data only).

## Test plan

- [x] `make test-fast` — 689 passed
- [x] Gate tests: `test_causal_queries.py`, `test_att_atu.py`, `test_public_api.py`
- [x] `make lint` clean
- [x] Manual smoke test of `__repr__` / `__float__` / `mean` / `hdi` / `prob` / `summary` / `__sub__`
- [ ] **Local docs re-freeze by @drbenvincent** — 11 example notebooks had executable cells edited (`causal_do_operator`, `ate_estimation`, `mediation`, `moderation`, `causal_identification`, `sensitivity`, `the_house`, `panel_data`, `mmm`, `saas_funnel`, `vaccine_surrogates`)

## Note for reviewers

`tests/test_public_api.py` is a snapshot of the public `__all__` lists; it was updated to include the new public `EstimandResult` class (done with maintainer sign-off, since AGENTS.md discourages editing tests).


drbenvincent:
@bugbot review

daimon-pymclabs (commented):
Reviewed the core changes (`simulate.py`, `model.py`, public API). Overall this is clean and the API direction is good — binding the outcome into `EstimandResult` so `.mean()`/`.hdi()`/`.prob()` no longer need it re-specified reads much more naturally than the old `DoResult` contrast.

The g-computation fix in `run_do_pymc` is the most important part and looks correct: applying the inverse link *per unit* before averaging within each draw is the right order, since `E[g⁻¹(μ)] ≠ g⁻¹(E[μ])` for non-identity links. Worth calling out in the changelog that this is a behavior change for cross-sectional `kind="mean"` — the contrast now collapses to `(n_samples,)` (one standardized effect per draw) rather than the old flattened `(n_samples × n_units,)`, so reported HDIs now reflect posterior uncertainty in the *average* effect rather than unit-level spread. That's the desired semantics for an estimand, just flagging it as user-visible.

A couple of smaller things:

1. **`EstimandResult.__sub__` drops the per-time data.** It rebuilds the result without passing `values_by_time`/`time_index`, so they default to `None`. That means subtracting two `simulate_over="time"` estimands silently loses `by_time()` (later `by_time()` calls then raise "Per-time data not available"). `DoResult.__sub__` preserves them — worth matching here for consistency, or intentionally documenting the difference.

2. **Accessor `KeyError` on unknown `var`.** `mean()`/`hdi()`/`draws()` index `self._values[...]` directly, so a typo'd variable name surfaces a bare `KeyError`. A small "unknown variable {var}; available: [...]" guard would be friendlier, matching the helpful messaging elsewhere in the class.

3. **`prob()` `eval`** — the locked-down namespace (`__builtins__: {}`) is reasonable, but only `SyntaxError` is caught; an expression like `prob("> foo")` raises a raw `NameError`. Minor, only if you want fully friendly errors.

None of these are blockers. Nice work.

PR #330, opened 2026-06-22
Fix pathmc.model shadowed by submodule (#329)

### LINKED ISSUE #329: pathmc.model shadowed by the pathmc.model submodule (top-level API resolves inconsistently)
## Summary

`pathmc.model` is **both** a module (`pathmc/model.py`) and a function (`model()` exported from it). Depending on import order, the package attribute `pathmc.model` sometimes resolves to the **submodule** instead of the **function**, so the documented top-level API `pathmc.model(spec, data=...)` intermittently raises:

```
AttributeError: module 'pathmc' has no attribute 'model'
```

(or, when the submodule wins, `TypeError: 'module' object is not callable`).

## Reproduction

The same code behaves differently depending on how Python is launched:

```console
$ python -c "import pathmc; print(type(pathmc.model))"
<class 'function'>          # function wins

$ printf 'import pathmc\nprint(type(pathmc.model))\n' > t.py && python t.py
AttributeError: module 'pathmc' has no attribute 'model'   # submodule shadow
```

It reproduced reliably for me when running a script from a file (vs `-c`), and while writing tests. It currently happens to resolve to the function under `pytest`, which is why the suite passes — but that's luck, not guaranteed.

## Cause

`pathmc/__init__.py`:

```python
from pathmc.model import PathModel, model, simulate
```

This binds the **function** `model` in the package namespace, but importing the submodule also sets `pathmc.model` to the **submodule** object. Whichever assignment happens last wins, and that order depends on what else imports `pathmc.model` (directly or transitively) during package initialization. The result is order-dependent and fragile.

## Suggested fix (pick one)

1. **Rename the module to remove the collision (preferred, durable).** Move `pathmc/model.py` → e.g. `pathmc/_model.py` (or `core.py`) and keep `from pathmc._model import PathModel, model, simulate` in `__init__.py`. Then `pathmc.model` is unambiguously the function, regardless of import order. Most libraries avoid a module and a public callable sharing a name for exactly this reason.
2. **Re-bind at the end of `__init__`.** Import the function under an alias and assign it last (`from pathmc.model import model as _model; ...; model = _model`). Lighter touch, but a later `import pathmc.model` anywhere can still clobber the attribute at runtime — option 1 is safer.

Either way, add a regression test:

```python
def test_pathmc_model_is_the_function():
    import pathmc
    assert callable(pathmc.model)
```

## Impact

Low severity (workaround: `from pathmc.model import model`), but it undermines the advertised top-level API and makes copy-pasted examples (including the #316 reproducer, which calls `pathmc.model(...)`) fail unpredictably. Worth a small standalone PR.

Found while building the graph-consistency harness (#328); unrelated to that change.

daimon-pymclabs:
Closes #329.

## Problem

`pathmc.model` was **both** a submodule (`pathmc/model.py`) and the public `model()` function exported from it. The package attribute `pathmc.model` resolved to whichever assignment ran last — the function binding from `__init__.py`, or the submodule object set by any (possibly transitive) `import pathmc.model`. That order is environment-dependent, so the documented top-level API `pathmc.model(spec, data=...)` intermittently raised `AttributeError: module 'pathmc' has no attribute 'model'` (or `TypeError: 'module' object is not callable` when the submodule won). The suite passed only by luck of import order.

## Fix

Option 1 from the issue (the durable one): rename `pathmc/model.py` → `pathmc/_model.py`. With no submodule named `model`, the public `model` attribute is unambiguously the function regardless of import order.

- `git mv pathmc/model.py pathmc/_model.py`
- Update the two internal importers: `pathmc/__init__.py` and `tests/test_priors.py` now import from `pathmc._model`.
- Drop the now-private `_model` from the public-submodule API test (`test_public_api.py`) — it's no longer part of the public surface.
- Add a regression test asserting `pathmc.model` is callable, including after a direct `import pathmc._model` (the old failure mode).

No public symbols change: `pathmc.model`, `pathmc.PathModel`, `pathmc.simulate` and `__all__` are all unchanged. No docs/mkdocs referenced the module path.

## Verification

Reproduced the original inconsistency, confirmed it's resolved (function in every import order), and ran the affected tests:

- `tests/test_public_api.py` + `tests/test_priors.py`: **41 passed**
- Full suite collects cleanly (896 tests); the only collection error is an unrelated missing optional dep (`polars`).


drbenvincent:
@bugbot review

drbenvincent (approved):

PR #331, opened 2026-06-22
fix: eliminate exog-lag scan carry to fix panel lag(x) logp bug

drbenvincent:
## Summary

Fixes the root-cause bug behind #316: panel models using `lag(x)` (exogenous lag) returned zero posteriors for the lag coefficient because `pm.sample`'s gradient computation evaluated against the wrong mu.

## Root Cause

PyTensor's scan-merge optimizer incorrectly permuted sit_sot carry channels when `pytensor.function` compiled `[mu_det, logp]` together. This happened because the exogenous lag carry had a **trivially-echo update** (`carry[t+1] = sequence[t]`), making two structurally-identical scan nodes (one from `replace_rvs_by_values`, one from `model.logp()`) candidates for merging. The merge incorrectly swapped the spend-carry with the sales-carry, producing:

```
mu_sales[t≥1] = intercept + beta * mu_sales[t-1]   # WRONG
```
instead of:
```
mu_sales[t≥1] = intercept + beta * spend[t-1]       # CORRECT
```

## Fix

Pre-compute lagged exogenous sequences as tensors **before** the scan:
```python
lagged_seq = pt.concatenate([init_row[None, :], exog_data[:-1]], axis=0)
```

These tensors are derived from `pm.Data` nodes, so `pm.set_data()` and `do()` interventions continue to propagate. With no trivial-echo carry, the optimizer has nothing to incorrectly merge.

## Testing

The 4 `xfail(strict=True)` red tests installed by #328 now pass unconditionally:

- `test_model_consistent[panel-lag(x)-complete]` ✅
- `test_model_consistent[panel-lag(x)-partial]` ✅
- `test_gaussian_likelihood_matches_hand_oracle[panel-lag(x)-complete]` ✅
- `test_gaussian_likelihood_matches_hand_oracle[panel-lag(x)-partial]` ✅

Full suite: 772 passed, 0 failures. Lint: clean.

Closes #316
Related: #326, #327

Issue #316, opened 2026-06-19
getting 0 posterior for lag effect


dirknbr:
I simulate data as 

``` 
ngeo = 5
ntime = 10
data = pd.DataFrame()
for g in range(ngeo):
  spend = np.random.normal(10, 1, ntime)
  sales = np.ones(ntime) * 20
  sales[1:] = 10 + spend[:-1] + np.random.normal(0, 1, ntime-1)
  df = pd.DataFrame({"spend": spend, "sales": sales})
  df['week'] = np.arange(ntime)
  df['geo'] = g
  data = pd.concat((data, df))
```

I then estimate

``` 
model = pathmc.model(
        "sales ~ lag(spend)",
        data=data,
        panel={"unit": "geo", "time": "week"},
        pooling="partial",
    )
model.fit(draws=1000, chains=2)
``` 

But I get a posterior 0 for lag(spend), why? True value is 1.

drbenvincent:
Thanks for these. I'll take a look. Most likely next week because of weekend family duties.

daimon-pymclabs:
## Diagnosis: `lag()` panel models silently optimize a corrupted likelihood

I dug into this. It reproduces, and it's a **real correctness bug in the scan-compiled temporal panel path**, not a sampling/identifiability problem. The model that `pm.sample` actually optimizes uses a `mu` for the likelihood that is **inconsistent with the model's own generative `mu`**, so NUTS correctly drives `beta[lag(spend)]` to ~0 against a corrupted objective.

### Evidence

**1. Reproduced.** `beta_sales[lag(spend)]` ≈ 0.019 ± 0.058; true value is 1.

**2. It is not a funnel / sampling issue.** Refitting with `pooling=None` (complete pooling, no random intercept, no funnel) samples cleanly — **0 divergences, good r_hat** — and `beta` is *still* 0.019 ± 0.065. A clean sampler on a correctly-specified linear model with an aligned, varying regressor would recover ≈1.

> Aside: under `pooling="partial"` there are also ~38 divergences, but those come from a *separate* redundant-intercept problem (the fixed `Intercept` term **plus** the random-intercept mean `mu_alpha` — their sum is identified, the pieces are not). That's worth a follow-up issue, but it is not what zeros the slope.

[...]

PR #332, opened 2026-06-22
fix: guard exog-lag sequence build against missing base data node

daimon-pymclabs:
Follow-up to #331, targeting its branch. Addresses the one correctness regression and the two typos found in review.

## KeyError regression (`compile.py`)

#331 builds the lagged exogenous sequence with a direct index:
```python
lagged_exog_sequences[base] = pt.concatenate(
    [init_row, exog_data_nodes[base][:-1]], axis=0
)
```
But the two key sets are not guaranteed equal:
- `exog_lag_bases` filters only on `base not in endo_set` (line 1329)
- `exog_data_nodes` **also** requires `base in data_sorted.columns` (line 1316)

So a `lag(x)` base whose contemporaneous column is absent from the data is in `exog_lag_bases` but not in `exog_data_nodes`, and the direct index raises `KeyError`. The adjacent `init_exog_lag` `else` branch (lines 1353-1354) already treats "base absent from data" as a reachable state, falling back to zeros — and the **removed** carry path tolerated it via `exog_t.get(k, pt.zeros(n_units))`, resolving such bases to the init row at `t=0` and zeros for `t>=1`.

This PR restores that behavior: when the base has no data node, build the lagged sequence as `[init_row, zeros((n_times-1, n_units))]`. This faithfully reproduces the old semantics, including the case where `init_exog_lag[base]` was seeded from a `{base}_lag1` column even though the contemporaneous `base` column is missing.

## AGENTS.md typos
- "comple issues" -> "complex issues"
- "obsolte ones" -> "obsolete ones"

## Not addressed here
- The `n_units=1` broadcastable concern raised in review (the `_init_carry` `pytensor.shared(..., broadcastable=(False,))` workaround is carry-specific; `init_row` is now a plain constant in a sequence). This is a *verification* item, not a confirmed bug — worth a targeted test rather than a speculative code change. No test files were modified, per the AGENTS.md guardrail.

## Validation
`py_compile` clean; `ruff check` and `ruff format --check` pass. Full suite not run here (no pymc/pytensor in the review sandbox).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,11 +2,11 @@
 
 `pathmc` is a Python package for Bayesian path analysis (observed-variable SEM). It compiles a lavaan-inspired formula DSL into PyMC models, provides introspection, and supports a `do()` operator for interventional simulation.
 
-Remember, when running into bugs or comple issues, it is not impossible we are finding bugs in external packages. In which case, flag potential issues to the user, suggesting the formation of an issue in the relevant upstream repo.
+Remember, when running into bugs or complex issues, it is not impossible we are finding bugs in external packages. In which case, flag potential issues to the user, suggesting the formation of an issue in the relevant upstream repo.
 
 ## Workflow
 
-1. Make your changes. Do not modify test files — if a test seems wrong, flag it for human review. Exceptions are for adding new tests or removing obsolte ones, but never for changing the expected behavior of existing tests.
+1. Make your changes. Do not modify test files — if a test seems wrong, flag it for human review. Exceptions are for adding new tests or removing obsolete ones, but never for changing the expected behavior of existing tests.
 2. Run the relevant gate tests, e.g. `uv run pytest tests/test_<module>.py -x -v`. `make test-fast` skips MCMC sampling; `make test` runs everything.
 3. Run `make lint` before considering work done or committing. It runs `prek run --all-files` (ruff, ruff-format, mypy, YAML/TOML, license checks).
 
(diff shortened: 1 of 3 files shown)


daimon-pymclabs:
Added `tests/test_exog_lag_missing_base.py` (4 tests) exercising the missing-base `else` branch directly, via the public `pathmc.model(...)` API — a `sales ~ lag(spend)` panel model where the contemporaneous `spend` column is absent from the data.

**Reachability confirmed:** all four tests raise `KeyError: 'spend'` at `compile.py:1418` against the pre-guard code and pass with the guard, so the branch is reachable through the public API (not dead code).

The tests pin:
- compiles without `KeyError`;
- genuinely takes the missing-base branch (the base is not materialized as a `pm.Data` node);
- stays graph-consistent under the no-sampling #316 battery (`assert_model_consistent` → mu context-invariance + finite logp/grad);
- **contributes exactly zero** — perturbing *only* the lag coefficient (holding the default `Intercept` fixed) leaves `mu_sales` bit-for-bit unchanged. (Worth noting: `sales ~ lag(spend)` adds an `Intercept` by default, so the lag is `beta_sales[1]`, not the whole vector.);
- the precomputed-`spend_lag1`-only variant (non-zero init row, no contemporaneous column) also compiles and stays consistent.

Validation in the sandbox: `ruff check` + `ruff format --check` clean; the 4 new tests pass; the 6 existing `panel-lag` consistency cells pass. mypy could not be run (the sandbox tmpfs fills before it completes) — flagging so a human runs `make lint` before merge.

Note: the pre-#331 carry implementation could not even compile this minimal missing-base model (empty `sequences` → scan can't infer `n_steps`), so #331's sequence approach actually *widens* what compiles here — the guard just keeps that path from crashing on the `KeyError`.

drbenvincent:
@bugbot review

juanitorduz:
@bugbot review

drbenvincent (approved):

PR #337, opened 2026-06-22
Fix nutpie scan crash for no-intercept lag models

### LINKED ISSUE #336: 'sales ~ 0 + lag(spend)' crashes with Reshape scan error on fit()
More issues uncovered by #316

## Summary

A scan-compiled panel model with no fixed intercept (`~ 0 + lag(x)`) builds successfully but crashes at `fit()` with a `pytensor` scan Reshape error. This blocks the natural fix for the partial-pooling redundant-intercept issue.

## Reproduction

```python
import pandas as pd, numpy as np, pathmc

frames = []
for g in range(5):
    spend = rng.normal(10, 1, 10)
    sales = np.ones(10) * 20
    sales[1:] = 10 + spend[:-1] + rng.normal(0, 1, 9)
    frames.append(pd.DataFrame({'spend': spend, 'sales': sales, 'week': np.arange(10), 'geo': g}))
data = pd.concat(frames, ignore_index=True)

model = pathmc.model(
    'sales ~ 0 + lag(spend)',   # no fixed intercept
    data=data,
    panel={'unit': 'geo', 'time': 'week'},
    pooling='partial',
)
# Model builds fine; free RVs look correct: beta_sales[1] (just the lag coefficient)
model.fit(draws=100, tune=100, chains=2)
```

## Error

```
ValueError: Argument Reshape{1}.0 given to the scan node is not compatible
with its corresponding loop function variable i3
```

## Context

- The model builds correctly: `beta_sales` has shape `[1]` (only `lag(spend)`, no intercept). The random intercept is properly captured by `mu_alpha_sales + alpha_sales[unit]`.
- The crash occurs in `nutpie.compile_pymc_model` → `clone_replace` → inside `pytensor.scan`'s `make_node`.
- Without the fix for this, the only way to avoid the partial-pooling redundant-intercept problem is `pooling=None`.
- This was first noted in the #316 diagnosis comment as 'the obvious workaround is also broken', and is listed in #327 item 11 as a crash vector.
- The fix in #331 (eliminating the exog-lag carry) did not resolve this crash — it's a different scan shape/broadcast issue.

## Related

- #327 item 11 — intercept removal as a crash vector for lag models
- Partial pooling redundant intercept issue (filed separately)
- #331/#332 — lag logp-corruption fix (separate issue, resolved)

drbenvincent:
## Summary

Fixes #336 by avoiding a PyMC/PyTensor raveled-input scan edge case for scan-compiled panel models with no fixed intercept and exactly one free temporal predictor, such as `sales ~ 0 + lag(spend)`.

The public `beta_<lhs>` RV, predictor dims, and coords are preserved, but `_compile_scan_panel()` now passes scalar beta components into `pytensor.scan` as non-sequences instead of passing the dimensioned beta vector directly. This keeps pathmc's public API stable while avoiding the length-one dimensioned-vector reshape mismatch that PyMC and nutpie hit during raveled-input replacement.

## Tests

- `uv run pytest tests/test_nutpie_scan_compile.py tests/test_graph_consistency.py -x -v`
- `uv run pytest tests/test_lag_syntax.py tests/test_panel_smoke.py tests/test_panel_engines.py -x -v -m "not slow"`
- manual issue path: `model.fit(..., nuts_sampler="nutpie")` for `sales ~ 0 + lag(spend)`, `pooling="partial"`
- `make test-fast`
- `make lint`

## Upstream note

I isolated a minimal pure-PyMC/PyTensor reproducer showing that PyMC's raveled NUTS setup fails when a length-one beta vector with named `dims` is used inside `pytensor.scan`. The upstream bug report is:

- pymc-devs/pymc#8337

This PR provides a local graph-shape workaround so pathmc users are not blocked on an upstream release. Follow-up cleanup is tracked in:

- #338

### DIFF
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -1466,12 +1466,17 @@ def _compile_scan_panel(
         # Non-sequences: all parameters
         non_seq_list: list[Any] = []
         non_seq_names: list[str] = []
+        beta_component_names: dict[str, list[str]] = {}
         non_seq_list.append(use_observed_carry)
         non_seq_names.append("_use_observed_carry")
         for var in endo_keys:
             if beta_rvs[var] is not None:
-                non_seq_list.append(beta_rvs[var])
-                non_seq_names.append(f"beta_{var}")
+                beta_component_names[var] = []
+                for idx in range(len(get_free_predictor_columns(reg_by_lhs[var]))):
+                    component_name = f"beta_{var}__{idx}"
+                    non_seq_list.append(beta_rvs[var][idx])
+                    non_seq_names.append(component_name)
+                    beta_component_names[var].append(component_name)
         for name, rv in tparam_rvs.items():
             non_seq_list.append(rv)
             non_seq_names.append(name)
@@ -1542,7 +1547,12 @@ def _compile_scan_panel(
             carry_mu: dict[str, Any] = {}
 
             for var in endo_keys:
-                beta = ns_map.get(f"beta_{var}")
+                beta_names = beta_component_names.get(var)
+                beta = (
+                    [ns_map[name] for name in beta_names]
+                    if beta_names is not None
+                    else None
+                )
 
                 resolver = _make_scan_resolver(
                     exog_t,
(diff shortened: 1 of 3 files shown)


drbenvincent:
@bugbot review

PR #339, opened 2026-06-22
Warn when partial pooling is combined with formula intercepts

### LINKED ISSUE #335: Partial pooling panel models have a redundant intercept causing divergences
More issues uncovered by #316 

## Summary

All `pooling='partial'` panel models have a non-identifiable parameterisation: the model includes both `beta_<y>[Intercept]` (a fixed global intercept from the regression coefficients) and `mu_alpha_<y>` (the mean of the random intercepts). Only their sum is identified by the data; individually they are unconstrained, creating a ridge that NUTS diverges on.

## Reproduction

```python
import pandas as pd, numpy as np, pathmc, arviz as az

frames = []
for g in range(5):
    x = rng.normal(0, 1, 10)
    y = 5 + 2*x + rng.normal(0, 1, 10)
    frames.append(pd.DataFrame({'x': x, 'y': y, 'week': np.arange(10), 'geo': g}))
data = pd.concat(frames, ignore_index=True)

model = pathmc.model('y ~ x', data=data, panel={'unit': 'geo', 'time': 'week'}, pooling='partial')
idata = model.fit(draws=500, tune=500, chains=4)
# → ~85 divergences
```

Free RVs: `sigma_alpha_y, mu_alpha_y, alpha_y[5], beta_y[2], sigma_y`

The total intercept for unit g is `beta_y[Intercept] + mu_alpha_y + alpha_y[g]`. Since `beta_y[Intercept]` and `mu_alpha_y` both have `Normal(0, 10)` priors and the data only sees their sum, they can trade off freely. This is not a lag-specific issue — it reproduces with any `pooling='partial'` model.

## Fix options

1. **Remove `beta[Intercept]` when `pooling='partial'`** — the intercept is already encoded in `mu_alpha`. The user-facing formula keeps `~ x` but the compiler drops the fixed intercept in favour of the hierarchical mean.
2. **Centre the random effects** — fix `mu_alpha = 0` and let the fixed intercept carry the mean. Equivalent but different convention.
3. **Warn the user** that `~ x` with `pooling='partial'` implies a redundant intercept and they should use `~ 0 + x` to remove it — but option B below shows that currently crashes.

## Related

- Discovered during the #316 investigation (lag models surface it more visibly but it's general)
- #316, #331, #332 — lag logp-corruption fix (separate issue, resolved)
- See also the `'sales ~ 0 + lag(spend)'` crash issue (filed separately)

drbenvincent:
Closes #335

## Summary

Adds a prominent `UserWarning` when partial pooling is combined with formula intercepts. This alerts users to a non-identifiable parameterization that causes sampling divergences, and provides a clear solution.

## Problem

Partial pooling models (`pooling='partial'`) with formula intercepts create a non-identifiable parameterization where both `beta[Intercept]` (fixed global intercept) and `mu_alpha` (hierarchical mean of random intercepts) compete for the same variance. Only their sum is identified by the data, causing ~85 divergences during sampling.

## Solution

Implemented **Option 3** from the issue discussion: Add a prominent warning that:
- Explains the problem (non-identifiable parameterization)
- Shows the exact fix: change `y ~ x` to `y ~ 0 + x`
- States that `mu_alpha` serves as the effective intercept
- Is raised in both cross-sectional and scan compilation paths

## Example Warning

```
==============================================================================
PARTIAL POOLING WITH REDUNDANT INTERCEPT
==============================================================================

Your model uses pooling='partial' (random intercepts) but the following
equations include a formula intercept: 'y'.

This creates a NON-IDENTIFIABLE parameterization:
  • beta[Intercept] (fixed global intercept)
  • mu_alpha (mean of random intercepts)
Only their sum is identified by the data. This causes sampling divergences.

SOLUTION: Remove the intercept from your formula(s):

  y ~ 0 + x

The hierarchical mean mu_alpha will serve as the effective intercept.
==============================================================================
```

## Changes

- `pathmc/compile.py`: Added `_warn_partial_pooling_intercept()` validation function
- `tests/test_partial_pooling_warning.py`: Comprehensive test suite (9 test cases)

## Testing

- ✅ All new tests pass (9/9)
- ✅ Existing panel tests pass with expected warnings
- ✅ Lint and type checks pass
- ✅ Verified warning in both cross-sectional and scan paths
- ✅ Confirmed warning message content and actionability

### DIFF
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -362,6 +362,7 @@ def compile_to_pymc(
         graph_info = build_graph(spec, latent=latent)
 
     _validate_residual_cov_families(spec, families)
+    _warn_partial_pooling_intercept(spec, pooling)
 
     if panel_info is not None and _has_temporal_deps(spec, graph_info):
         _validate_scan_non_gaussian_intermediaries(spec, families, latent)
@@ -993,6 +994,61 @@ def _validate_residual_cov_families(spec: Spec, families: dict[str, str]) -> Non
                 )
 
 
+def _warn_partial_pooling_intercept(spec: Spec, pooling: str | dict | None) -> None:
+    """Warn if partial pooling is combined with formula intercepts.
+
+    Partial pooling models include both a fixed intercept (beta[Intercept])
+    and a hierarchical mean (mu_alpha). Only their sum is identified by the
+    data, creating a non-identifiable parameterization that causes divergences.
+
+    Raises
+    ------
+    UserWarning
+        When any equation has an intercept and pooling requests random intercepts.
+    """
+    import warnings
+
+    if not _has_random_intercepts(pooling):
+        return
+
+    equations_with_intercepts = [
+        reg.lhs for reg in spec.regressions if reg.has_intercept
+    ]
+
+    if not equations_with_intercepts:
+        return
+
+    var_list = ", ".join(f"'{v}'" for v in equations_with_intercepts)
+    formulas = "\n".join(
+        f"  {reg.lhs} ~ 0 + {' + '.join(t.variable for t in reg.terms)}"
+        for reg in spec.regressions
+        if reg.has_intercept
+    )
+
+    warnings.warn(
+        f"\n{'=' * 78}\n"
+        f"PARTIAL POOLING WITH REDUNDANT INTERCEPT\n"
+        f"{'=' * 78}\n"
+        f"\n"
+        f"Your model uses pooling='partial' (random intercepts) but the following\n"
+        f"equations include a formula intercept: {var_list}.\n"
+        f"\n"
+        f"This creates a NON-IDENTIFIABLE parameterization:\n"
+        f"  • beta[Intercept] (fixed global intercept)\n"
+        f"  • mu_alpha (mean of random intercepts)\n"
+        f"Only their sum is identified by the data. This causes sampling divergences.\n"
+        f"\n"
+        f"SOLUTION: Remove the intercept from your formula(s):\n"
+        f"\n"
+        f"{formulas}\n"
+        f"\n"
+        f"The hierarchical mean mu_alpha will serve as the effective intercept.\n"
+        f"{'=' * 78}\n",
+        UserWarning,
+        stacklevel=4,
+    )
+
+
 # ---------------------------------------------------------------------------
 # Temporal dependency detection
 # ---------------------------------------------------------------------------
@@ -1185,6 +1241,8 @@ def _compile_scan_panel(
 
     if priors is None:
         priors = default_priors(spec, families, pooling, latent)
+
+    _warn_partial_pooling_intercept(spec, pooling)
 
     reg_by_lhs = {r.lhs: r for r in spec.regressions}
     transform_map = _build_transform_map(spec)
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -994,6 +994,37 @@ def _validate_residual_cov_families(spec: Spec, families: dict[str, str]) -> Non
                 )
 
 
+def _render_term_for_formula(term: Term) -> str:
+    """Reconstruct the original term string from a parsed Term.
+
+    Handles transforms, interactions, lags, and labeled coefficients.
+    """
+    prefix = ""
+    if term.label is not None:
+        prefix = f"{term.label}*"
+    elif term.fixed_value is not None:
+        prefix = f"{term.fixed_value}*"
+
+    if term.transform is not None:
+        return prefix + _render_transform_call(term.transform)
+    # Interactions and lags already have the full string in .variable
+    return prefix + term.variable
+
+
+def _render_transform_call(tc: TransformCall) -> str:
+    """Recursively render a TransformCall back to its original syntax."""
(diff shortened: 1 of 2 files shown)


drbenvincent:
Thoughts on this @juanitorduz, @cetagostini? Particularly this bit:

> The hierarchical mean mu_alpha will serve as the effective intercept.

I think it's reasonable, but wanted to get your input

juanitorduz (approved):

drbenvincent:
minor updates incoming

daimon-pymclabs:
## Review — conceptual call is right; two implementation defects worth fixing before 0.1.1

**On the design question (mu_alpha as effective intercept, ZeroSumNormal):** the wording is correct and I'd keep it. The random intercept is built centered — `alpha = pm.Normal(f"alpha_{var}", mu=mu_alpha, sigma=sigma_alpha, dims="unit")` — so once `beta[Intercept]` is dropped via `~ 0 + x`, `alpha[g] ~ Normal(mu_alpha, sigma_alpha)` and `mu_alpha` *is* the population-level mean intercept that units shrink toward. That is exactly an effective intercept.

ZeroSumNormal is **not** needed for the route this PR takes. The naive alternative (keep `beta[Intercept]`, set `alpha ~ Normal(0, sigma_alpha)`) does *not* fix identifiability — the realized sample mean of the alphas is still free to drift and trades off against `beta[Intercept]` exactly as `mu_alpha` does now; `Normal(0, sigma)` pins it only softly, leaving a milder funnel/ridge. `pm.ZeroSumNormal` is what closes that gap (hard `sum(alpha)=0`), but it's only relevant if you *retain* the formula intercept and use genuine deflections. The current approach has a single location parameter and is fully identified. So ZeroSumNormal is an alternative design, not a missing piece — I wouldn't block on it. (A nice future enhancement: when a user insists on `~ x` with partial pooling, auto-reparameterize to `beta[Intercept]` + `ZeroSumNormal` deflections rather than warn — issue #335's Option 1 done cleanly.)

**Is the warning robust?** Mostly, but two concrete defects the 9 tests don't catch (they only exercise `y ~ x`, a non-temporal cross-sectional path):

### 1. The suggested-fix formula silently drops transforms
The reconstruction uses `t.variable`:
```python
formulas = "\n".join(f"  {reg.lhs} ~ 0 + {' + '.join(t.variable for t in reg.terms)}" ...)
```
But for a transformed term `Term.variable` is the *leaf* (`parse.py:263` — `log(x)` → `variable="x"`, the transform lives in a separate attribute). So a model written `y ~ log(x)` is told to write `y ~ 0 + x`, dropping the log entirely. Interactions (`a:b`) and lags (`lag(spend)`) survive because they keep the full string, but transforms and labeled coefficients (`label*x`) do not. The advice is wrong precisely for the non-toy models. Fix: reconstruct from the original formula string, or render each term faithfully.

### 2. The warning double-fires on the scan/lag path
`_warn_partial_pooling_intercept` is called at `compile.py:365`, then line 368 dispatches to `_compile_scan_panel`, which calls it *again*. So any panel model with temporal deps (the lag models — exactly the ones where #335 surfaced this) emits the warning twice. The `len(w) == 1` test passes only because `y ~ x` has no temporal deps and never enters the scan path. Fix: hoist the call to a single site, or guard it.

### One thing to confirm
The warning steers users to `~ 0 + x`, and #335 noted `sales ~ 0 + lag(spend)` was crashing. That looks resolved by #337 (merged), but worth confirming the recommended escape hatch doesn't lead a lag-model user into a different error — otherwise the warning is sound advice into a wall.

Net: conceptual call is right, warning text is good, but defects 1 and 2 bite exactly the transformed/lag models this feature protects. Happy to push a commit fixing the double-fire and the formula reconstruction.

drbenvincent:
Thanks for the detailed review @daimon-pymclabs! I've addressed both defects:

## Fixed: Transform dropping (Defect #1)
Added `_render_term_for_formula()` and `_render_transform_call()` to properly reconstruct term strings. Now:
- `y ~ logistic_saturation(x, lam=lam)` suggests `y ~ 0 + logistic_saturation(x, lam=lam)`
- `y ~ 2*x` suggests `y ~ 0 + 2*x` 
- `y ~ a:b` suggests `y ~ 0 + a:b`
- `y ~ lag(y)` suggests `y ~ 0 + lag(y)`

## Fixed: Double-firing (Defect #2)
Removed duplicate call from `_compile_scan_panel()`. Warning now fires exactly once from `compile_to_pymc()` before dispatch.

## Test coverage expanded
Added 5 new tests (14 total, up from 9):
- `test_no_double_fire_on_lag_models` - catches the double-fire bug
- `test_warning_preserves_log_transform` - labeled coefficients
- `test_warning_preserves_interaction` - interaction terms
- `test_warning_preserves_lag` - lag terms  
- `test_no_intercept_lag_model_works` - confirms `~ 0 + lag(...)` compiles (fix from #337)

All tests pass, including the originally-failing transform/lag scenarios you identified.

## Re: ZeroSumNormal
Agreed — that's a different design (Option 1 from #335) worth considering for a future enhancement. This PR implements Option 3 (warn + suggest removal) which is fully identified as-is.

Issue #309, opened 2026-06-18
Bayesian model averaging of effects across candidate DAGs (structure uncertainty)


daimon-pymclabs:
Part of #306. Depends on the discovery port and the DAG→spec translator.

## Goal

Propagate **structural uncertainty** into the effect posteriors: when discovery returns several equally-plausible graphs, don't pick one — fit each and **Bayesian-model-average** the resulting ATE/CATE.

## Scope

- Fit each candidate spec emitted by the translator.
- Combine effect posteriors with appropriate weights (marginal likelihood / stacking / LOO-based weights — evaluate options).
- Report both the **averaged** estimand and the **per-graph spread**, so users see how much of their uncertainty is structural vs parametric.
- Stretch: a single summary that says "the effect is robust across the equivalence class" vs "the sign flips depending on graph X→Y orientation."

This is the piece that makes pathmc uncategorizable by the competition — full Bayesian uncertainty over *structure*, not just parameters. Related: #166 (graph-aware priors).

Issue #306, opened 2026-06-18
[Epic] Causal discovery → structure uncertainty (learn-the-DAG, model-average the effects)


daimon-pymclabs:
## Summary

Today pathmc is strictly **"DAG-in"**: the user supplies a single graph and we estimate/identify/falsify against it (causal discovery is a hard ❌ in the comparison doc). This epic tracks the work to flip that into **"learn-a-DAG (or a *set* of plausible DAGs) → model → propagate structural uncertainty into the effect posteriors."**

This is the flagship stream-2 ("expand what pathmc fundamentally is") bet. The differentiator almost nobody owns: we don't pretend the graph is known — we quantify uncertainty *over the structure itself* and carry it all the way through to the ATE/CATE. It also maps directly onto how our EAP clients actually work (co-owning a set of equally-plausible DAGs and cutting impossible edges by business assumption).

## Layering

1. **Discovery** — port the causal-discovery tools from PyMC-Marketing. Output is generally a *set* of equally-plausible graphs (a CPDAG / Markov-equivalence class), not one. (#307)
2. **DAG → spec translator** — turn a discovered graph into a pathmc DSL spec. Must accept a *list* of graphs from the start, not a single one. (#308)
3. **Structure uncertainty / Bayesian model averaging** — fit each candidate spec and model-average the effect posteriors, so structural uncertainty propagates into the final estimand. (#309)

Related prior issues: #166 (graph-aware priors for structure learning), #168 (expressibility boundaries).

## Owner

Discovery port + DAG→model: @cetagostini (Carlos). Layering/roadmap: Ben.

## Open design questions

- Discovery method being ported: constraint-based (PC), score-based (GES), or the correlation/CI approach used in EAP sessions? Determines whether output is a single DAG or an equivalence class.
- Translator target: emit the lavaan-style DSL string (cleanest, given the `parse`/`compile` split) vs. build the PyMC model directly?

PR #349, opened 2026-06-23
Add causal-discovery front end (TBFPC, CPDAG, DAG→model)

### LINKED ISSUE #307: Port causal-discovery tools from PyMC-Marketing into pathmc
Part of #306.

## Goal

Port the causal-discovery tooling from PyMC-Marketing into pathmc, giving pathmc a **discovery front end** (currently a hard ❌ — pathmc only consumes a graph it's handed).

## Scope

- Bring over the discovery algorithm(s) from PyMC-Marketing.
- Return type should preserve the **equivalence class** where the method produces one (CPDAG / set of DAGs), not collapse to a single arbitrary DAG — this is what downstream model-averaging (#306) needs.
- Expose orientation constraints / background-knowledge hooks so users (and EAP clients) can forbid impossible edges by business assumption.

## Open question

Which method is being ported — constraint-based (PC), score-based (GES), or the correlation/conditional-independence approach Carlos uses in EAP sessions? This determines the output type (single DAG vs equivalence class).

Owner: @cetagostini

cetagostini:
## Summary

Ports the causal-discovery tooling from PyMC-Marketing into pathmc, giving pathmc a **discovery front end**. pathmc was previously strictly *DAG-in* (you hand it a graph); it can now **learn candidate structure from data** and turn a discovered graph into a model.

Closes #307. Part of the structure-uncertainty epic #306.

## What's new (public API)

- **`discovery.TBFPC`** — Target-first Bayes Factor PC discovery (a target-oriented PC variant using a ΔBIC Bayes-factor CI test).
  - Returns a **CPDAG** — the whole Markov equivalence class — not a single arbitrary DAG. `get_all_cdags_from_cpdag()` enumerates the acyclic orientations, which is what downstream model-averaging (#309) needs.
  - **Background-knowledge hooks:** `forbidden_edges` / `required_edges` let users (and EAP clients) forbid impossible edges and force known ones.
  - `discovery.TestResult` records per-test ΔBIC statistics.
- **`cpdag.same_markov_equivalence_class`** — skeleton + v-structure equality across DOT strings, `graphviz` `.source` objects, and `networkx.DiGraph`, via a dependency-free DOT reader.
- **`dag.dag_to_spec`** — translate a DAG (DOT / `"A->B"` edge list / `networkx.DiGraph`) into a pathmc lavaan-style DSL string.
- **`dag.BuildModelFromDAG`** — build a model from a DAG in two styles:
  - `style="digraph"` — the **fully linear** Gaussian reading of the graph (one slope per edge, dims/coords aligned via xarray).
  - `style="native"` — DAG → DSL → `pathmc.model()`, unlocking families, transforms, latent mediators, custom priors, and the full intervention API.

## Design notes

- **No new dependencies.** Implemented with pathmc's existing core deps (networkx/pymc/pytensor/numpy/pandas). **No `dowhy`, no `pydot`** — pathmc keeps its own native backdoor identification (`identify.py`); `CausalGraphModel` is intentionally *not* ported.
- Adaptations from the PyMC-Marketing original: pydantic `validate_call` → pathmc-style manual validation; PyTensor 3 `pt.linalg.pinv`; the TBFPC independence test now decides in **log space** so an overwhelmingly dependent pair can't overflow `float64`; `BuildModelFromDAG.build()` **reindexes** pivoted data to the declared coord order so unsorted coords can't silently mislabel inference.

## Testing

- **159 new tests** (`tests/test_cpdag.py`, `tests/test_discovery.py`, `tests/test_dag.py`), covering regular + corner cases and the discovery→model seam.
- Hardened by an adversarial multi-lens review pass (18 findings fixed, with regression guards added).
- `mypy`, `ruff`, license checks all pass.

### DIFF
--- a/docs/user_guide/18-causal-discovery.qmd
+++ b/docs/user_guide/18-causal-discovery.qmd
@@ -1,0 +1,214 @@
+---
+title: "Causal Discovery"
+guide-section: "Concepts"
+---
+
+Most of pathmc assumes you already have a DAG: you write the structural equations, and pathmc estimates, identifies, and falsifies against that graph.
+Causal **discovery** flips the starting point — it learns *candidate* structure from data, so you can use it when the graph is unknown or only partially known.
+The discovery front end is `TBFPC`; the resulting graph feeds into `dag_to_spec()` and `BuildModelFromDAG`, which turn it into a model you can fit and intervene on with the rest of pathmc. See the API Reference for the full signatures of each.
+
+::: {.callout-important}
+## Discovery is not a substitute for domain knowledge
+
+Structure learning recovers an *equivalence class* of graphs that are statistically indistinguishable, not a single ground truth. Treat its output as a set of hypotheses to combine with what you already know — forbid edges that are impossible by domain assumption, and keep the conclusions you draw honest about the structural uncertainty that remains. `TBFPC` is **experimental** and emits a warning on construction; its API may change.
+:::
+
+## Discovery returns an equivalence class, not one DAG
+
+A key idea: data alone usually cannot orient every edge.
+The chain `A → B → C`, the other chain `A ← B ← C`, and the fork `A ← B → C` all imply exactly the same conditional independencies, so no test on observational data can tell them apart — they form a *Markov equivalence class*.
+Discovery therefore returns a **CPDAG** (completed partially directed acyclic graph): some edges are oriented (they point the same way in every member of the class), and the rest are left undirected (they may point either way).
+pathmc preserves this distinction end to end rather than collapsing to one arbitrary DAG, because downstream model-averaging needs the whole set.
+
+## Learning structure with `TBFPC`
+
+`TBFPC` (Target-first Bayes Factor PC) is a target-oriented variant of the PC algorithm.
+It first tests which candidate drivers connect to your outcome, then builds the skeleton among the drivers, using a Bayes-factor (ΔBIC) test for conditional independence at each step.
+Standardizing the columns first (zero mean, unit variance) keeps that test well scaled.
+
+```python
+import numpy as np
+import pandas as pd
+from pathmc import TBFPC
+
+rng = np.random.default_rng(7)
+n = 2000
+C = rng.gamma(2, 1, n)
+A = 0.7 * C + rng.gamma(2, 1, n)
+D = 0.5 * C + rng.gamma(2, 1, n)
+B = 0.8 * A + rng.gamma(2, 1, n)
+Y = 0.9 * B + 0.6 * D + 0.7 * C + rng.gamma(2, 1, n)
+
+df = pd.DataFrame({"A": A, "B": B, "C": C, "D": D, "Y": Y})
+df = (df - df.mean()) / df.std()  # recommended scaling
+
+model = TBFPC(target="Y", target_edge_rule="fullS")
+model.fit(df, drivers=["A", "B", "C", "D"])
+
+model.get_directed_edges()  # e.g. [("B", "Y"), ("C", "Y"), ("D", "Y")]
+model.get_undirected_edges()  # e.g. [("A", "B"), ("A", "C"), ("C", "D")]
+print(model.summary())
+```
+
+The fitted model exposes its result several ways:
+
+| Method | Returns |
+|---|---|
+| `get_directed_edges()` | Oriented edges `(u, v)`, sorted |
+| `get_undirected_edges()` | Unoriented adjacencies `(u, v)`, sorted |
+| `summary()` | A text report of directed/undirected edges and the number of CI tests run |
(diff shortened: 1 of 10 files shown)


drbenvincent:
Excellent. I'm currently pretty sick, so I probably can't get to it until the end of the week. If I'm well enough I might trigger some directed agent review

juanitorduz:
@cetagostini after we merge this, do you still wanna keep these methods in PyMC-Marketing?

cetagostini:
> @cetagostini after we merge this, do you still wanna keep these methods in PyMC-Marketing?

Nope, ideally after merge then we add pathmc as optional dependency, as we have dowhy (we replace) and then we keep all examples an notebook very marketing driven there, but here the more open applications.

@juanitorduz

cetagostini:
## Follow-up on the review of #350: the actual fix is now on `causal-discovery` (`22cd41a`)

The review of #350 surfaced a deeper problem than the one #350 set out to fix, so I addressed the root cause here instead.

### Root cause

`TBFPC` collected separating sets but never used them. After the skeleton phase it left **all** driver–driver edges undirected, so its output was not a valid CPDAG — it was an under-specified PDAG that straddled multiple Markov equivalence classes. That made `get_all_cdags_from_cpdag` ill-posed:

- The original code over-generated, mixing DAGs from *different* equivalence classes (the bug #350 spotted).
- But #350's filter-only fix (drop any orientation that adds a new v-structure) goes too far in the other direction: on a genuine collider it **silently discards the identifiable answer**. With data generated from `A→B←C` (so `A ⟂ C` marginally, i.e. `B ∉ sepset(A,C)`), discovery returns 3 non-collider DAGs and the true collider is never among them.

The "artificial cut" instinct in the review was right: collider-vs-non-collider is *observationally identifiable*, so it should be **decided from the data**, not silently assumed away.

### The fix (option a)

`fit()` now runs the PC orientation phase:

1. **Unshielded colliders** — for each driver triple `x — z — y` (x,y non-adjacent), orient `x→z←y` iff `z ∉ sepset(x,y)`, using the separating sets TBFPC already records.
2. **Meek rules R1–R3** propagate the compelled edges to a fixpoint.

Only edges that are compelled (identical across the whole class) become directed; genuinely reversible edges stay undirected. `get_all_cdags_from_cpdag` keeps the *correct* part of #350 — an orientation survives only if it is acyclic **and** introduces no new v-structure, which is exactly the equivalence-class membership test.

Result:

| Data | `fit()` | `get_all_cdags` |
|---|---|---|
| collider `A→B←C`, `B→Y` | `A→B, C→B, B→Y`; no undirected | **1 DAG = the true collider** |
| chain `A→B→C`, `B→Y` | `B→Y`; `A—B, B—C` undirected | 3 DAGs (CPDAG survives) |

So the filter does **not** kill CPDAGs — it removes only the *spurious* ambiguity (identifiable colliders) and preserves genuine within-class ambiguity (chains, forks, cliques).

### Secondary review items (kept, since they remain relevant)

- `TBFPC.max_conditioning_set_size` (default 3), validated + documented, replacing the silent `min(3, …)` cap. It now also bounds the sepsets the v-structure step relies on, so it is correctness-relevant, not just a perf knob.
- `BuildModelFromDAG.time_dim` (default `"date"`), replacing the hardcoded `"date"` magic dim — including the warning message and docstring that #350 left hardcoded.
- `build()` hoists `set_index`/`to_xarray` out of the node loop and raises a clear error when reindexing to declared coords would inject NaN.

### Tests / checks

New regression tests cover collider recovery, reversible-CPDAG survival, equivalence-class membership of every enumerated DAG, `max_conditioning_set_size` validation + search bound, `time_dim`, and the reindex-NaN guard. No existing test was modified — the pinned `synthetic_df` recovery tests still hold because orientation is a no-op on that fixture. `make test-fast` (997 passed) and `make lint` are clean.

The user guide is updated to describe the v-structure/Meek orientation and the now-accurate equivalence-class enumeration.

cetagostini:
@juanitorduz the @daimon-pymclabs review was nice but issues was upstream. All should be good now!

daimon-pymclabs:
Thanks @cetagostini — agreed, this is the right place to fix it. `22cd41a` orienting v-structures in TBFPC so discovery emits a proper CPDAG addresses the actual root cause (the unused separating sets) rather than the symptom my #350 was patching, and the new cases in `test_dag.py` / `test_discovery.py` cover both the over-generation and the dropped-collider failure modes. Happy to close #350 as superseded by this. 👍

cetagostini:
@drbenvincent @juanitorduz Daimon is happy, I would suggest merge and continue iterations!

drbenvincent:
I'm OOO tomorrow. Happy to merge if Juan gives approval. But also happy to review when I'm back on Wednesday

juanitorduz (approved):

PR #357, opened 2026-06-26
Stay in xarray in do() producers; simplify with native broadcasting

### LINKED ISSUE #356: Stay in xarray in do() producers; delete numpy re-wrap machinery
## Goal (read this first)

After #353, producers already *store* labelled `xr.Dataset`s, but they still
**convert PyMC/ArviZ DataArrays to numpy and back** before storage. The point
of this issue is to stop that.

**Producers should assemble `xr.DataArray`s end-to-end.** The only place we
materialize flat numpy sample vectors is the public API boundary:
`_stack_sample` / `_stack_sample_time` (used by `.draws()` / `.by_time()`).

Do **not** treat “extract private helpers” as success if `simulate.py` net
grows and `.to_numpy()` / `.values` / `_da_chain_draw*` calls in producers
are unchanged.

## Context

#353 (PR #355) moved `DoResult` / `EstimandResult` construction into
`run_do_pymc` and `run_do_panel_unified`. Production no longer uses flat
`dict[str, np.ndarray]` or `_build_dataset()`.

What remains is **adapter machinery** left over from the flat-dict era:
`.to_numpy()` → reshape → `_da_chain_draw` / `_da_chain_draw_unit` /
`_da_time_chain_draw`. That machinery should shrink or disappear as producers
stay in xarray.

### 1. Eliminate producer round-trips (main work)

For each endogenous variable in `run_do_pymc` and `run_do_panel_unified`:

| Path | Today (anti-pattern) | Target |
|------|----------------------|--------|
| Cross-sectional `kind="predictive"` | `ppc[...].to_numpy()` → `_da_chain_draw(_unit)` | `isel` / `rename` / `transpose` on the source `DataArray`; no numpy |
| Cross-sectional `kind="mean"` | `det[...].to_numpy()` → numpy slice/mean → `_da_chain_draw` | xarray slice/mean; inverse link on `DataArray` (e.g. `xr.apply_ufunc`); return labelled `(chain, draw)` without re-wrap |
| Panel mean/predictive | `source.values.reshape(...)` → numpy mean → `_da_time_chain_draw` | xarray reshape/mean/transpose with explicit dims where possible; avoid `.values` unless dims are genuinely unknown |

**Constants** (set / exogenous fills): build with `xr.full` + coords, not
`np.full` → `_da_chain_draw`.

**Inverse link:** extend `_apply_inverse_link` (or add a thin xarray wrapper)
so the mean path does not call `np.asarray(da.values)` mid-pipeline.

### 2. Delete machinery that exists only for re-wrapping

As round-trips go away, remove or narrow:

- `_da_chain_draw_unit` (if still present) — only existed for numpy re-label
- Calls to `_da_chain_draw` / `_da_time_chain_draw` **inside producers** (they
  may remain for edge cases or move to a single “ensure coords” helper)
- Any helper whose sole job is “numpy array → reshape → DataArray with chain/draw
  coords” after we already had a DataArray

Keep `_stack_sample` / `_stack_sample_time` — those are the intentional
`.draws()` boundary.

### 3. Deduplicate only where it falls out naturally

OK to extract helpers for logic that is **identical** and **stays in xarray**:

- Missing-RV dummy fill (twice in `run_do_pymc`)
- Predictive source lookup (`ppc`, `mu_*`, `extra_det`)
- Panel unit-mean → `(time, chain, draw)` (mean and predictive)

Do **not** add a large “producer helper” layer if it duplicates the same
numpy bridges under new names.

### 4. Small cleanups (only if trivial)

- Remove unused `panel_info` from `run_do_panel_unified` (or wire it for time labels)
- `_model.py`: use public `.draws()` instead of `_draw()`
- Repr sample counts from dim sizes, not `len(self.draws(...))`
- Shrink `tests/_draw_fixtures.py`; tests can build `xr.Dataset` directly
- Remove stale “planned xarray migration” docstrings

## Acceptance criteria (agent must verify)

1. **Grep gate — producers only** (functions `run_do_pymc`, `run_do_panel_unified`,
   and any helpers they call *except* `_stack_sample*`):
   - No `.to_numpy()` on PyMC/ArviZ outputs
   - No `np.asarray(...values)` used to re-wrap after xarray ops
   - `_da_chain_draw` / `_da_time_chain_draw` not called on arrays that were
     just extracted from a `DataArray` (constants/`xr.full` paths are fine)

2. **Behaviour:** existing tests pass, especially:
   - `tests/test_xarray_storage.py`
   - `tests/test_estimand_result_api.py`, `tests/test_reprs.py`
   - `tests/test_panel_by_time.py`
   - `tests/test_causal_queries.py`, `tests/test_att_atu.py`

3. **Size:** `simulate.py` should be **net smaller or neutral** vs the #353
   branch — not +100 lines of private helpers. If helpers are added, more
   numpy-bridge code must come out.

4. **Readability:** `run_do_pymc` / `run_do_panel_unified` read as “get
   labelled DataArray from PyMC → slice/mean/transpose → assign to Dataset”,
   not “convert to numpy and rebuild.”

## Out of scope

- Changing public `draws()` / `mean()` / `hdi()` numpy contract
- `by_unit()` accessor
- Merging `run_do_pymc` and `run_do_panel_unified` into one function
- `EffectResult` xarray migration

## Depends on

- #353 (merged via PR #355)

## Anti-patterns (do not do this)

- Extract 10+ private helpers that re-wrap the same numpy pipelines
- Claim success because duplication dropped while `.to_numpy()` count is unchanged
- Optimize for “~80–120 lines removed” by moving code into helper functions
  without removing numpy bridges

drbenvincent:
## Summary

Closes #356.

The `do()` producers (`run_do_pymc`, `run_do_panel_unified`) now assemble labelled `xr.DataArray`s end-to-end. The leftover `.to_numpy()` → reshape → `_da_chain_draw*` round-trips from the flat-dict era are gone, and so is the adapter machinery that existed only to re-wrap numpy back into labelled arrays.

Rather than swap the old numpy wrappers for new single-use ones, the producers lean on xarray's native broadcasting and named reductions so they read as label algebra:

- **Constants by broadcasting.** A coord-carrying `ones = _chain_draw_ones(post)` turns every intervention/exogenous fill into `ones * value` — no manual `dims`/`coords` or `np.full`/`np.broadcast_to`.
- **Mean path** inlines `inverse_link → isel(subgroup) → .mean(obs_dims)`; the g-computation standardization is one expression.
- **Predictive path** just renames the observation dim to `unit`; `_stack_sample` flattens it at the public `.draws()` boundary.
- **Panel** uses `_panel_unit_mean` (average units → `(time, chain, draw)`) and `_spread_over_time` for scalar/per-time interventions.
- **`_apply_inverse_link`** dropped its dead numpy branch and dispatches `np.exp`/`pt.exp` on the input type, operating directly on `DataArray`s (no `xr.apply_ufunc`).

Helpers removed: `_da_chain_draw`, `_da_chain_draw_unit`, `_da_time_chain_draw`, `_const_chain_draw`, `_const_time_chain_draw`, `_select_subgroup`, `_mean_to_chain_draw`, `_predictive_to_chain_draw`, `_panel_to_time_chain_draw`, `_extra_dims`, `_chain_draw_coords`. Kept/added only small label-only helpers used across multiple call sites: `_obs_dims`, `_chain_draw_ones`, `_spread_over_time`, `_panel_unit_mean`, `_fill_missing_posterior_rvs`, `_predictive_source`, `_exog_value`.

Small cleanups: `_model.py` `prob()` / `sensitivity()` use public `.draws()`; repr sample counts come from dim sizes.

**Size:** `simulate.py` is 79 lines smaller than `main` (985 vs 1064), with no numpy bridges remaining in the producers.

## Test plan

- [x] `uv run pytest tests/test_xarray_storage.py tests/test_estimand_result_api.py tests/test_reprs.py tests/test_panel_by_time.py tests/test_causal_queries.py tests/test_att_atu.py`
- [x] Full `make test`: all pass, 87.8% coverage
- [x] `make lint` (ruff, ruff-format, mypy, license checks)
- [x] Grep gate: no `.to_numpy()` / numpy re-wrap on PyMC/ArviZ outputs in producers
- [x] Added coverage for time-varying (array-valued) panel interventions — the `(n_times,)` `set` path through `run_do_panel_unified` was previously exercised only with scalar values. New tests assert the intervened variable's per-time draws track the supplied array exactly.

## Notes for reviewers

- No tests referenced the deleted helpers or the old xr→np→xr machinery, so none were removed; the inverse-link mean path (bernoulli/poisson) is covered by `test_bernoulli.py` / `test_families.py`.
- `_panel_unit_mean` and the predictive path identify the observation dims positionally (`obs[0]` = time, `obs[1]` = unit) because the scan compiler names them generically (`mu_sales_dim_0` / `_dim_1`). This `(time, unit)` ordering was confirmed against a live scan model; if the compiler ever changes it, this is the spot that would need a coord-name lookup instead.

### DIFF
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -1461,7 +1461,7 @@ class PathModel:
             raise ValueError(f"n_grid must be >= 2, got {n_grid}.")
 
         ate_result = self.ate(outcome, treatment, values=values, **do_kwargs)
-        ate_draws = ate_result._draw(outcome)
+        ate_draws = ate_result.draws(outcome)
 
         return compute_sensitivity(
             observed_ate_draws=ate_draws,
@@ -1503,7 +1503,7 @@ class PathModel:
         self._require_data("prob")
         result = self.do(set=set, kind=kind, **do_kwargs)
         namespace: dict[str, Any] = {
-            str(var): result._draw(str(var)) for var in result.dataset.data_vars
+            str(var): result.draws(str(var)) for var in result.dataset.data_vars
         }
         namespace["np"] = np
         namespace["__builtins__"] = {}
--- a/pathmc/simulate.py
+++ b/pathmc/simulate.py
@@ -94,56 +94,147 @@ def _chain_draw_coords(n_chains: int, n_draws: int) -> dict[str, np.ndarray]:
     return {"chain": np.arange(n_chains), "draw": np.arange(n_draws)}
 
 
-def _da_chain_draw(
-    arr: np.ndarray,
-    n_chains: int,
-    n_draws: int,
+def _extra_dims(da: xr.DataArray) -> list[str]:
+    """Dims other than ``chain`` / ``draw`` on a PyMC output array."""
+    return [d for d in da.dims if d not in ("chain", "draw")]
+
+
+def _select_subgroup(
+    da: xr.DataArray, subgroup_indices: np.ndarray | None
 ) -> xr.DataArray:
-    """Wrap a ``(chain, draw)`` array with labelled coords."""
-    arr = np.asarray(arr).reshape(n_chains, n_draws)
+    """Restrict observation rows to *subgroup_indices* (first extra dim)."""
+    if subgroup_indices is None:
+        return da
+    extra = _extra_dims(da)
+    if extra:
+        return da.isel({extra[0]: subgroup_indices})
+    return da
+
+
+def _const_chain_draw(
+    fill: float | np.ndarray, n_chains: int, n_draws: int
+) -> xr.DataArray:
+    """Constant ``(chain, draw)`` array with integer coords."""
+    scalar = float(np.mean(fill)) if isinstance(fill, np.ndarray) else float(fill)
+    coords = _chain_draw_coords(n_chains, n_draws)
     return xr.DataArray(
-        arr,
+        np.broadcast_to(scalar, (n_chains, n_draws)),
         dims=("chain", "draw"),
-        coords=_chain_draw_coords(n_chains, n_draws),
+        coords=coords,
     )
 
 
-def _da_chain_draw_unit(
-    arr: np.ndarray,
-    n_chains: int,
-    n_draws: int,
-) -> xr.DataArray:
-    """Wrap a ``(chain, draw, unit)`` array with labelled coords."""
-    arr = np.asarray(arr)
-    n_units = arr.shape[-1]
-    arr = arr.reshape(n_chains, n_draws, n_units)
-    return xr.DataArray(
-        arr,
-        dims=("chain", "draw", "unit"),
-        coords={
-            **_chain_draw_coords(n_chains, n_draws),
-            "unit": np.arange(n_units),
-        },
-    )
-
-
-def _da_time_chain_draw(
-    arr: np.ndarray,
+def _const_time_chain_draw(
+    val: float | np.ndarray,
     n_times: int,
     n_chains: int,
     n_draws: int,
     time_index: np.ndarray,
 ) -> xr.DataArray:
-    """Wrap a ``(time, chain, draw)`` array with labelled coords."""
-    arr = np.asarray(arr).reshape(n_times, n_chains, n_draws)
+    """Constant or time-varying ``(time, chain, draw)`` array."""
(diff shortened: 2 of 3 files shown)


drbenvincent:
@bugbot review

Issue #360, opened 2026-07-02
Feature: Hilbert Space Gaussian Process (HSGP) term for smooth nonparametric effects


daimon-pymclabs:
## Summary

pathmc currently has **no Gaussian Process support of any kind** — a grep for `hsgp`, `gaussian process`, `pm.gp`, `Hilbert` returns nothing across code, docs, and tests. This issue scopes adding a **Hilbert Space Gaussian Process (HSGP)** term to the DSL, giving pathmc a nonparametric smooth analogous to Bambi's `hsgp()` formula term ([bambi HSGP 1D notebook](https://bambinos.github.io/bambi/notebooks/hsgp_1d.html)).

Prior art worth reviewing for design cues:
- Bambi's closed HSGP PRs: https://github.com/bambinos/bambi/pulls?q=is%3Apr+is%3Aclosed+hsgp
- Bambi issue framing the feature: https://github.com/bambinos/bambi/issues/982

Related pathmc issue: #98 (time-varying coefficients for panel models) — an HSGP over time is one way to express smoothly-varying effects, so these should stay design-aware of each other.

## Motivation

pathmc models are strictly linear-in-coefficients along each path. Any nonlinear-but-smooth dependence (a smooth trend over time, a smooth dose–response that isn't captured by `logistic_saturation`, a smooth confounder adjustment) currently has no expression. HSGP is the standard scalable way to add such a smooth: it approximates a GP with a finite Laplacian eigenfunction basis, so it reduces to a linear model in a fixed basis matrix — a natural fit for a library already built around design matrices and `pm.do()`.

## Why it isn't a drop-in today

The closest existing machinery is the **transform system** (`pathmc/transforms.py`): a registry (`adstock`, `logistic_saturation`) of parametrized, estimable, composable functions with automatic priors, panel/scan-awareness, and `pm.do()` compatibility. That's the right conceptual neighbor, **but not a drop-in**:

- Transforms map **one input column → one output tensor**.
- HSGP must expand **one input column → an M-column basis matrix** that multiplies M free coefficients.

So the bottleneck is the design-matrix layer, not the kernel math. `build_design_matrix` in `compile.py` emits one column per term, and the two resolvers (`_make_cross_sectional_resolver`, `_make_scan_resolver`) dispatch on `slot.kind`. HSGP needs a new slot kind plus basis expansion at that layer.

[...]

PR #362, opened 2026-07-06
Add hsgp() term for 1-D cross-sectional nonparametric smooths

daimon-pymclabs:
Implements **Phase 1** of the HSGP scaffolding — a Bambi-style `hsgp()` term for 1-D cross-sectional nonparametric smooths. Executes the design in #361 / #360.

```python
model = pathmc.model("y ~ hsgp(x, m=20, c=1.5)", data=df)
idata = model.fit()
model.do(set={"x": 12.0})   # basis recomputes through the intervention
```

## What's included

- **`parse.py`** — `HSGPCall` AST node, `Term.hsgp` field, and `_parse_hsgp_expr` with literal coercion (`m`→int, `c`/`L`→float, `cov`→lowercased str, `centered`→bool) and guardrails: missing `m`, both/neither `c`/`L`, unknown kwargs (`by=`, `iso`, …), multi-dimensional input, coefficient prefix, transform nesting (`adstock(hsgp(...))`), interaction use, and duplicate `(lhs, variable)` smooths.
- **`hsgp.py`** (new) — `make_cov_func` (ExpQuad / Matern52 / Matern32), `hsgp_basis` (takes a required `cov_func` so `sqrt_psd` stays consistent with the estimated `eta`/`ell` — the BugBot finding from #361), and `assemble_hsgp_term` for both centered and non-centered parametrizations.
- **`compile.py`** — `kind="hsgp"` predictor slot with an **inert `coeff_type` marker** (not `"free"`, per review); the `build_mu` HSGP branch sits **ahead of** coefficient bookkeeping so `free_idx` stays aligned with `beta`; `get_free_predictor_columns` skips HSGP; weights coord registration; resolver gains `lhs`/`priors`; a **model-level panel guard** (catches panel models even without temporal deps), residual-block rejection, and a scan-resolver guard.
- **`priors.py`** — `_collect_hsgp_defaults`: `ell` `InverseGamma(3,1)`, `eta` `HalfNormal(1)`, `beta_hsgp` `Normal(0,1)` with weights dims; flat override-able keys `ell_{lhs}_{var}` etc.
- **`introspect.py`** — `f_hsgp(x)` equation + LaTeX rendering and prior listing.
- **`_model.py`** — `simulate()` rejects `hsgp()` terms.

## Verification

Verified against the real **pymc 6.0.1** in-env: `prior_linearized` centers `X` internally and derives the boundary from `c` at build time, so raw non-zero-mean inputs work and the basis recomputes under `do()` within the fitted support.

- New tests (47): `test_hsgp_parse`, `test_hsgp` (kernel/basis/assemble units), `test_hsgp_compile`, `test_hsgp_priors`, and slow `test_hsgp_do` — including a **non-zero-mean recovery test** (`x ∈ [10,20]`, corr > 0.9) as a centering guard and a `do()` basis-recomputation test.
- Existing suites for the touched modules pass unchanged (133).
- `ruff check`, `ruff format`, and `mypy` clean on all changed files.

## Follow-ups (out of Phase 1 scope)

Panel/scan models, multi-dimensional inputs, and grouped (`by=`) GPs each raise a clear error today and are left as follow-ups. The docs example `docs/examples/01-foundations/hsgp_1d.qmd` is added but its `_freeze/` output needs regenerating with `great-docs` (not available in the dev sandbox).

Related to #360.


juanitorduz:
@bugbot review

pathmc/parse.py
```diff
@@ -395,6 +455,133 @@ def _parse_transform_expr(raw: str) -> TransformCall:
...
+                )
+            positional.append(arg.strip())
+
+    if len(positional) > 1:
+        joined = ", ".join(positional)
+        raise ParseError(
+            f"Multi-dimensional hsgp({joined}) is not supported yet "
+            "(see follow-up). Use a single input variable."
+        )
+    variable = positional[0]
```
daimon-pymclabs, line 561: Bug: this raises `IndexError` (not `ParseError`) when the call is keyword-only with no positional input, e.g. `hsgp(m=20, c=1.5)`. The guard at line 486 only checks that `args[0]` is non-empty, not that any positional arg exists — with all-kwargs `positional` is `[]`, so `positional[0]` throws.

Confirmed by running the parser in isolation:
```
>>> parse_spec('y ~ hsgp(m=20, c=1.5)')
IndexError: list index out of range
```

Suggest adding a guard before line 520:
```python
if not positional:
    raise ParseError(
        "hsgp(...) requires an input variable as its first argument. "
        "Example: hsgp(x, m=20, c=1.5)."
    )
variable = positional[0]
```
and a case in `test_guardrails_raise_parse_error`, e.g. `"y ~ hsgp(m=8, c=1.5)"  # no positional input`.

pathmc/hsgp.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    weights_dim = f"{lhs}_{var}_hsgp"
+    beta_name = f"beta_hsgp_{lhs}_{var}"
+
+    if call.centered:
+        # Coefficient scale is data-derived (sqrt_psd), so beta is built
+        # directly rather than from the Prior config; a user override on
+        # beta_hsgp has no effect in centered mode (tune ell/eta instead).
+        beta = pm.Normal(beta_name, mu=0.0, sigma=sqrt_psd, dims=weights_dim)
```
daimon-pymclabs, line 232: The comment correctly notes a user override on `beta_hsgp` has no effect in centered mode — but `_collect_hsgp_defaults` and `build_priors` still register/list `beta_hsgp_{lhs}_{var}` unconditionally, so `model.priors()` advertises it as tunable in centered mode and `merge_priors` accepts an override for it silently. That's a confusing no-op for the user.

Suggest one of: (a) skip registering `beta_hsgp` when `centered=True` (so an override raises the usual "Unknown prior key"), or (b) `warnings.warn(...)` if an override is supplied in centered mode. (a) is the cleaner signal. Either way, `introspect`/`priors()` should reflect that in centered mode the tunable knobs are `ell`/`eta`, not `beta_hsgp`.

pathmc/parse.py
```diff
@@ -395,6 +455,133 @@ def _parse_transform_expr(raw: str) -> TransformCall:
...
+        )
+
+    if "m" not in kwargs:
+        raise ParseError(
+            "hsgp(...) requires m=<int> (number of basis vectors). "
+            "Example: hsgp(x, m=20, c=1.5)."
+        )
+    try:
+        m = int(kwargs["m"])
+    except ValueError:
```
daimon-pymclabs, line 581: Nit: `raise ParseError(...)` inside `except ValueError` will chain the coercion `ValueError` into the traceback ("During handling of the above exception..."). Add `from None` (or `from err`) to keep the user-facing message clean. Same pattern at lines 299, 555, 560. If ruff's `B904` is enabled it would flag these.

daimon-pymclabs (commented):
Reviewed Phase 1 of the HSGP scaffolding. Overall this is clean, well-scoped, and the layering (kernel/basis math isolated in `hsgp.py`, compiler dispatch, prior defaults, introspection, and the panel/scan/residual-block/simulate guards) is coherent. Test coverage is strong across parse, compile, priors, unit, and slow do()/recovery layers, including the non-zero-mean centering guard.

One real bug and a couple of smaller UX/polish points below. Nothing blocks the design; the parser bug is worth fixing before merge since it contradicts the stated "guardrail for every malformed input" goal.

**1. Uncaught `IndexError` on keyword-only `hsgp()` (bug).** `hsgp(m=20, c=1.5)` — i.e. no positional input variable — raises `IndexError: list index out of range` instead of a clean `ParseError`. Verified in isolation. Every other malformed form in `test_guardrails_raise_parse_error` raises `ParseError`; this one leaks a raw `IndexError`. Fix + a test case are noted inline.

**2. Centered-mode `beta_hsgp` override is silently ignored.** In centered mode `beta` is built directly from `sqrt_psd`, but `_collect_hsgp_defaults`/`build_priors` still register and list `beta_hsgp_{lhs}_{var}` as an overridable prior. A user who overrides it gets no effect and no warning. Detail inline.

**3. `raise ParseError(...)` inside `except ValueError` blocks** loses exception-chaining intent — add `from None` (or `from err`) so the traceback doesn't append the coercion `ValueError`. Minor. Applies at parse.py:299, 540, 555, 560.

**Nits (no change required):** `assemble_hsgp_term`'s `assert n_basis == call.m` is stripped under `python -O`; it's an internal invariant so likely fine, but a raised error would be more robust. `make_cov_func`/`hsgp_basis` type `cov_func` as `object` — `pm.gp.cov.Covariance` would document intent better if mypy allows.

pathmc/parse.py
daimon-pymclabs: Fixed in 147a484. Added an explicit `if not positional:` guard before `variable = positional[0]` in `_parse_hsgp_expr`, raising the ParseError shown above, plus the guardrail case `"y ~ hsgp(m=8, c=1.5)"` in `test_guardrails_raise_parse_error`. Verified the keyword-only forms now raise `ParseError` and valid forms still parse.

pathmc/hsgp.py
daimon-pymclabs: Fixed in 147a484 (option a). `beta_hsgp_{lhs}_{var}` is now registered only when `centered=False`: guarded with `if not call.centered` in `_collect_hsgp_defaults`, and the matching entry in `build_priors`/introspect is likewise gated on `not term.hsgp.centered`. So in centered mode `priors()` no longer advertises `beta_hsgp`, and an override on it raises the usual "Unknown prior key" instead of being a silent no-op — `ell`/`eta` stay the tunable knobs. Added `test_centered_hsgp_does_not_register_beta_prior` and `test_centered_hsgp_beta_override_raises_unknown_key`.

pathmc/parse.py
daimon-pymclabs: Fixed in 147a484. The three `raise ParseError(...)` inside `except ValueError` for `m`/`c`/`L` coercion now use `... from None`. Note on line 299: that one is not a re-raise inside `except` — it is a fallback assignment (`float()` fails → treat the token as a label), so it is intentionally left as-is; no chaining issue there.

pathmc/hsgp.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Returns
+    -------
+    tuple
+        ``(phi, sqrt_psd, n_basis)`` -- ``phi`` of shape ``(n, m)``,
+        ``sqrt_psd`` of shape ``(m,)``, and ``n_basis == m``.
+    """
+    boundary = [call.L] if call.L is not None else None
+    gp = pm.gp.HSGP(m=[call.m], L=boundary, c=call.c, cov_func=cov_func)
+    phi, sqrt_psd = gp.prior_linearized(x)
```
drbenvincent, line 120: **P1 — reject endogenous HSGP inputs in Phase 1.** Here `x` may resolve to an upstream endogenous RV. PyMC `prior_linearized()` calls `.eval()` on `X` to freeze its midpoint and boundary, so `m ~ x; y ~ hsgp(m, ...)` builds a basis from an unobserved random draw. I built that model twice, then fixed `m`, `ell`, `eta`, and the HSGP weights; `f_y_m` still differed by up to 2.1. The fitted model therefore depends on compilation-time RNG. Reject terms whose input is endogenous (or make centering and boundaries deterministic) and add a regression test.

pathmc/parse.py
```diff
@@ -395,6 +455,148 @@ def _parse_transform_expr(raw: str) -> TransformCall:
...
+        raise ParseError(f"hsgp(...) m must be >= 1, got {m}.")
+
+    has_c = "c" in kwargs
+    has_l = "L" in kwargs
+    if has_c == has_l:
+        raise ParseError("hsgp(...) needs exactly one of c=<float> or L=<float>.")
+
+    c: float | None = None
+    boundary_l: float | None = None
+    if has_c:
```
drbenvincent, line 595: **P1 — validate HSGP boundary literals at the parser boundary.** `c=0`, `c=-1`, `L=0`, and `L=-1` all parse and compile, but drawing `f_y_x` produces non-finite values (including divide-by-zero warnings). `nan` is also accepted. Reject non-finite and non-positive `c`/`L` values before constructing the PyMC graph, with guardrail tests for each invalid class.

docs/examples/01-foundations/hsgp_1d.qmd
```diff
@@ -0,0 +1,98 @@
+---
```
drbenvincent, line 1: **P1 — this documentation page cannot ship with frozen execution.** There is no `_freeze/.../hsgp_1d/execute-results/html.json` in this branch, while `freeze: true` means documentation builds never execute a new page. Readers will not receive the promised fitted output and figure. Regenerate and commit the freeze cache, or leave this page out of the Phase 1 PR.

docs/examples/01-foundations/hsgp_1d.qmd
```diff
@@ -0,0 +1,98 @@
...
+
+```{python}
+idata = model.fit(random_seed=0)
+f_mean = idata.posterior["f_y_x"].mean(("chain", "draw")).values
+np.corrcoef(f_mean - f_mean.mean(), true - true.mean())[0, 1]
+```
+
+The posterior-mean smooth tracks the true function closely.
+
+```{python}
```
drbenvincent, line 60: **P2 — make this a Bayesian teaching figure.** Posterior draws are available, but this renders an ad-hoc recentered `f_mean` line without uncertainty. Plot `mu_y` posterior mean with a 94% HDI, followed by the true curve; that handles the intercept directly and avoids implying certain recovery. Add the Quarto figure label/caption and fold the plotting cell as well.

drbenvincent (changes requested):
## GPT-5.6 Terra High — request changes

The exogenous 1-D compile and intervention path is cleanly separated and the happy-path coverage is substantial. I ran the new HSGP tests locally: **47 fast + 3 slow passed**.

However, the three inline blockers make this unsafe to merge as Phase 1: an endogenous input makes the basis depend on a compilation-time random draw, invalid boundary literals produce non-finite graphs, and the new frozen documentation page has no committed execution cache.

The documentation is not final-ready yet. The additional inline note covers the minimal figure changes needed to make the example communicate posterior uncertainty correctly.

— GPT-5.6 Terra High

daimon-pymclabs:
Thanks — all three P1 blockers and the P2 figure note are addressed in 98d2908.

**P1 — endogenous HSGP inputs.** Confirmed the mechanism: `prior_linearized()` evaluates `X` to freeze the centering midpoint and boundary, so an endogenous input builds the basis from a prior draw. Rejected at spec level in `_reject_endogenous_hsgp_inputs` (compile.py), which fires alongside the existing panel/residual-block guards — before any graph construction, so the user gets a clean `NotImplementedError` naming the offending variable and equation rather than a silently RNG-dependent fit. Two tests: `test_endogenous_hsgp_input_raises`, and `test_exogenous_hsgp_input_alongside_other_equations_allowed` to pin that the guard does not over-fire on a multi-equation spec whose smooth input is exogenous.

I went with rejection rather than making the centering/boundary deterministic. Deterministic centering would need the midpoint and `L` to come from something other than the input's realized values, which is a real design question (fit-time data statistics? user-supplied?) and worth its own PR rather than a rushed Phase 1 answer.

**P1 — boundary literals.** `_validate_hsgp_boundary` in parse.py now rejects non-finite and non-positive `c`/`L` before the PyMC graph is built. Guardrail tests cover `0`, negative int, negative float, `nan`, `inf`, `-inf` for both `c` and `L` (11 new cases).

**P1 — freeze cache.** You were right that it was missing. Regenerated with `great-docs freeze docs/examples/01-foundations/hsgp_1d.qmd`; `_freeze/examples/01-foundations/hsgp_1d/` is now committed, including `figure-html/fig-hsgp-recovery-output-1.png`. Note the path is `_freeze/examples/...`, not `_freeze/docs/examples/...` — that is what the tool emits, matching the other example pages. Re-ran the freeze after the prose edit so the hash matches the final source.

**P2 — teaching figure.** Replaced the recentered `f_mean` line with the `mu_y` posterior mean plus a 94% HDI band, then the true curve on top. That carries the intercept directly, so no manual recentering, and it stops implying certain recovery. Added `label: fig-hsgp-recovery`, a caption, and `code-fold: true`. One note: this repo is on arviz 1.1.0 where the kwarg is `prob=`, not `hdi_prob=` — the first freeze attempt failed on exactly that, so I matched the existing `az.hdi(..., prob=0.94)` usage in `custom_transforms.qmd`.

Also documented the exogenous-input restriction in the page's Scope section, since it is now a user-visible constraint rather than an implementation detail.

**Verification:** 63 HSGP tests pass (fast + slow). Full suite: 1106 passed, 6 failed — all six are `ModuleNotFoundError: No module named pyarrow` in the polars-backend tests, an artifact of my local env, untouched by this change. ruff check, ruff format, and mypy all clean.

drbenvincent:
## Re-review summary

**Technically ready, but not merge-ready yet** — CI is green and the earlier blockers look addressed, but GitHub still shows `CHANGES_REQUESTED` with no re-approval after the fix commit (`98d2908`). Someone with approval rights needs to re-review and approve before merge.

---

### What's in good shape

**CI / automation:** All checks pass — full test suite, `prek`, docs build, Bugbot.

**Architecture:** Clean Phase 1 split:
- `hsgp.py` — kernel/basis math, `cov_func` threaded through so `sqrt_psd` stays consistent with estimated `eta`/`ell`
- `compile.py` — `kind="hsgp"` with inert `coeff_type="hsgp"` marker, assembled before `beta` indexing
- Parse guardrails are thorough (duplicate smooths, transform nesting, invalid kwargs, boundary validation)

**Review fixes all landed:**

| Issue | Status |
|---|---|
| Endogenous HSGP inputs freeze basis at compile-time RNG | `_reject_endogenous_hsgp_inputs` + tests |
| Invalid `c`/`L` (0, negative, nan, inf) compile to non-finite graphs | `_validate_hsgp_boundary` + 11 parse tests |
| Missing `_freeze/` for docs page | Committed with figure |
| Teaching figure without uncertainty | `mu_y` mean + 94% HDI |
| Keyword-only `hsgp(m=8, c=1.5)` → `IndexError` | Fixed + test |
| Centered mode `beta_hsgp` silently ignored | Prior only registered when `centered=False` |

**Tests:** 60 fast tests pass locally; 3 slow tests cover `do()` basis recomputation and smooth recovery (including non-zero-mean `x ∈ [10, 20]`).

---

### Minor concerns (non-blocking)

1. **Docs example has 133 divergences** in the frozen output. The figure is fine, but the sampler struggled. Consider bumping `target_accept` in the example or a one-line note that HSGP hyperparameters can be stiff — purely polish.

2. **No user-guide page** — only `docs/examples/01-foundations/hsgp_1d.qmd`. Fine for Phase 1, but HSGP won't be as discoverable as transforms/panel topics until a guide lands.

3. **Documented limitations are real** (and correctly guarded):
   - Exogenous inputs only
   - No panel/scan, residual blocks, or `simulate()`
   - `do()` interventions should stay within fitted `x` support when using `c`

4. **`assert n_basis == call.m`** was upgraded to a raised `ValueError` — good fix for the earlier nit.

---

### Recommendation

@drbenvincent — could you re-review? The substantive concerns from the Jul 31 review appear resolved in `98d2908`, and the author documented the fixes clearly. I wouldn't hold this for new code changes unless we want the divergences note in the example.

pathmc/hsgp.py
juanitorduz: Fixed in 98d2908. `_reject_endogenous_hsgp_inputs` now runs at compile time next to the panel/residual-block guards and raises `NotImplementedError` when an `hsgp()` input is the LHS of another equation, so the basis can no longer be frozen from a compilation-time prior draw. Latent inputs are covered by the same guard: pathmc requires every latent variable to be a regression LHS, so they fall in the same set. Regression tests: `test_endogenous_hsgp_input_raises` (your exact repro shape) and `test_exogenous_hsgp_input_alongside_other_equations_allowed` to pin the guard's non-firing case. The page's Scope section now documents the restriction.

pathmc/parse.py
juanitorduz: Fixed in 98d2908. `_validate_hsgp_boundary` in the parser now rejects non-positive and non-finite `c`/`L` with a `ParseError` before any PyMC graph is built — `c=0`, `c=-1`, `L=0`, `L=-1`, `nan`, `inf`, and `-inf` are all covered. Parametrized guardrail tests per invalid class: `test_non_positive_boundary_rejected` and `test_non_finite_boundary_rejected`.

docs/examples/01-foundations/hsgp_1d.qmd
juanitorduz: Fixed in 98d2908 and refreshed in b692d5a. The executed freeze cache is now committed at `_freeze/examples/01-foundations/hsgp_1d/execute-results/html.json` (plus the rendered figure PNG), so the page ships with its fitted output under `freeze: true` without CI ever spawning a kernel. Verified locally with `great-docs build`: the page renders with all cell outputs from the cache.

docs/examples/01-foundations/hsgp_1d.qmd
juanitorduz: Fixed in 98d2908. The figure now plots the `mu_y` posterior mean with a 94% HDI band (`az.hdi`) against the observed data and the true curve — `mu_y` carries the intercept, so the ad-hoc `f_mean` recentering is gone. The cell has a Quarto label (`fig-hsgp-recovery`), a caption noting the band widens at the boundaries, and `code-fold: true`. b692d5a adds the examples-index thumbnail rendered from the same figure.

juanitorduz:
## Review round 2 — summary of changes

All four items from @drbenvincent's review are addressed on the branch (commits 98d2908 and b692d5a); each inline thread has a detailed reply.

**P1 — endogenous HSGP inputs are now rejected** (`pathmc/compile.py`). `_reject_endogenous_hsgp_inputs` runs at compile time alongside the panel/residual-block guards and raises `NotImplementedError` when an `hsgp()` input is the LHS of another equation. `prior_linearized()` evaluates its input to fix the basis midpoint and boundary, so an endogenous input froze the basis at a compilation-time prior draw. Latent inputs fall under the same guard, since every latent must be a regression LHS. Tests cover the failing repro and the non-firing exogenous case; the docs page's Scope section documents the restriction.

**P1 — boundary literals validated at the parser** (`pathmc/parse.py`). `_validate_hsgp_boundary` rejects non-positive and non-finite `c`/`L` (`0`, negatives, `nan`, `±inf`) with a `ParseError` before any PyMC graph is built, with parametrized guardrail tests per invalid class.

**P1 — freeze cache committed.** `_freeze/examples/01-foundations/hsgp_1d/execute-results/html.json` (plus the rendered figure) is now on the branch, so the page ships with its fitted output under `freeze: true` and CI never spawns a kernel. Verified locally with `great-docs build`.

**P2 — Bayesian teaching figure.** The plot now shows the `mu_y` posterior mean with a 94% HDI band against the observed data and the true curve — `mu_y` carries the intercept, so the ad-hoc `f_mean` recentering is gone. The cell has a Quarto label/caption and `code-fold: true`.

**Page polish** (b692d5a): examples-index thumbnail rendered from the recovery figure + `image:` frontmatter, icons for the new Gaussian Process / HSGP / Nonparametric tags in `great-docs.yml`, and prose unwrapped to one paragraph per line per AGENTS.md, with the freeze cache refreshed to match.

Verification: lint clean, 1118 fast tests pass, slow HSGP do()/recovery tests pass, docs site builds from the committed cache with the labeled figure and thumbnail card.

daimon-pymclabs:
## Final review at `b692d5a` — one blocker, two follow-ups

Ran the branch locally rather than reading it: all 60 fast HSGP tests pass, and they pass on **pymc 6.2.0** as well as the 6.0.1 this was developed against, which is a useful data point given the `>=6.0,<7` pin.

The earlier review points are addressed and I verified them:

- **`beta` index alignment holds.** `y ~ hsgp(x, m=10, c=1.5) + z` gives `y_predictors = [Intercept, z]` with the HSGP slot correctly excluded, and after fitting `beta_z` recovers `0.517` against a true `0.5`. Mixed free/fixed/hsgp renders as `mu_y = 1 + f_hsgp(x) + z + 0.5*w`.
- **Centered/non-centered asymmetry is correct and non-silent.** With `centered=true`, overriding `beta_hsgp_y_x` raises `Unknown prior key(s): ['beta_hsgp_y_x']. Valid keys: ['beta_y', 'ell_y_x', 'eta_y_x', 'sigma_y']`.
- **Parser guards are solid.** I threw 24 malformed forms at it; every one produces a clean, specific `ParseError` except the case in item 3 below.
- `predict()` is keyword-only in-sample with no new-data argument, so there is no unguarded out-of-sample basis rebuild. Not a gap.

The architecture is sound and nothing below touches it.

---

### 1. Blocking: `do()` past the frozen boundary returns plausible, silently wrong numbers

With `c` (rather than an explicit `L`), `prior_linearized` freezes the boundary from the concrete `x` at build time. Fitting on `x ∈ [0.03, 9.97]` with `c=1.5` gives `L = 7.46` about a midpoint of `5.0`, so the valid region is `[-2.46, 12.46]`.

Because the eigenfunctions are sinusoids, an intervention outside that region **aliases** rather than diverging. `do(set={"x": 1000})` returns `-0.897`: finite, inside the plausible range of the observed data, no `nan`, no warning. There is no way for a user to tell that answer apart from a real one.

`do()` is the core of this library, so a counterfactual that looks answered and isn't is the worst available failure mode. The `.qmd` documents the limitation in prose (line 81), but prose doesn't fire at runtime.

Suggested fix, roughly ten lines: at intervention time, compare the requested values against `[mid - L, mid + L]` and warn (or raise) when any fall outside. Worth noting the docs statement is also slightly conservative — `x = 11` is outside the fitted data yet still recovers fine, so the real threshold is the **boundary**, not the data range, and the warning should say so.

### 2. Default priors are stiff enough to be a day-one user problem

On `x ∈ [0, 10]` with `m=25, c=1.5` at default `target_accept`: **59/600 divergences** on one run and **273/600** on an identical re-run, with r-hat > 1.01 and ESS < 100 both times. At `target_accept=0.95` it still warns. Posterior `ell` mean is `1.79` against an `InverseGamma(3, 1)` prior whose mean is `0.5` — the prior is pulling against the data.

This is the unstandardized design matrix (#230) surfacing, and it is the same thing producing the 133 divergences in the frozen docs example. Minimum before merge: bump `target_accept` in the example and add a sentence about the input scale. The real fix is scaling the `ell` default off the input range, which can be a follow-up.

### 3. The interaction guard only covers one side

`hsgp(x, m=5, c=1.5):z` raises a clean `ParseError`. But the other ordering does not:

```
z:hsgp(x, m=5, c=1.5)
  -> ValueError: Unknown transform 'z:hsgp'. Available transforms: adstock, logistic_saturation
```

Same for a capitalization slip: `HSGP(x, ...)` gives `Unknown transform 'HSGP'`. Non-silent, so minor, but the error names the wrong concept and won't help anyone debug it. Fine as a follow-up.

---

**Verdict:** item 1 is worth the ten lines before merge; item 2 needs at least the docs note; item 3 can land later.

juanitorduz:
## Re: final review — all three items addressed

**1. Out-of-boundary `do()` (blocker)** — fixed in 89b0e61, as a `raise` rather than a warn, since no valid answer exists past the boundary. `hsgp_intervention_bounds()` (`pathmc/hsgp.py`) recomputes the frozen support per input — midpoint `(max+min)/2`, `L` as given or `c·(max−min)/2`, mirroring `set_boundary` under the `>=6,<7` pin — and `do()` raises `ValueError` when any `set` value (scalar or array) falls outside `[mid − L, mid + L]`. The existing softer data-range warning still fires for values outside the data but inside the boundary, matching your point that the boundary, not the data range, is the real threshold; the docs prose now says the same. A fast test pins the recomputed bounds against `gp._X_center`/`gp.L` from an actual `prior_linearized` call, and slow tests cover both the raise and the warn-only band. (`shift` is still reserved/not implemented, so `set` is the only intervention path.)

**2. Divergences** — minimum done in c2d77ec: the example samples at `target_accept=0.99` (the refrozen run shows 0 divergences on all four chains, vs 33 before) and the priors section notes that the `InverseGamma(3, 1)` lengthscale default is in raw input units. The range-scaled default is tracked in #375, which carries your divergence measurements.

**3. Parser naming** — took the ten minutes now rather than deferring (89b0e61): `z:hsgp(...)` raises a parse-time `ParseError` naming the interaction restriction, and `HSGP(...)`/`Hsgp(...)` raise `Unknown function 'HSGP'. Did you mean hsgp(...)?` — neither reaches the transform registry anymore.

Verification: lint clean (ruff, mypy), 1127 fast tests, all 5 slow HSGP do-tests, docs build green, freeze cache regenerated after the source edits.

daimon-pymclabs:
## Re-check at `c2d77ec` — all three items land

Reviewed the diff this round rather than re-running the suite (no local PyMC 6 env available to me at the moment), so treat the below as a read of the code plus the tests you added, not a fresh execution.

**1. Out-of-boundary `do()` — resolved, and `raise` was the right call.** `hsgp_intervention_bounds` reconstructs the same midpoint/`L` that `prior_linearized` freezes (`mid = (max+min)/2`, `L = c·(max−min)/2` or the explicit `L`), and `test_hsgp_intervention_bounds_matches_pymc_frozen_boundary` pins it against `gp._X_center`/`gp.L` from a real call, which is the check that keeps the reimplementation honest. Two things I chased down and am satisfied with:

- The bounds are recomputed from `self._data`, and the compile path feeds `prior_linearized` the same raw column via `_resolve_var` → `pm.Data` with no intervening transform, so the two agree by construction. The parser rejecting `hsgp()` inside interactions is what keeps that true.
- The guard sits in `PathModel.do()`, and every other intervention entry point (`effect`, the contrast helpers, `run_do_panel_unified`) routes through it — there is no path to `pm.do()` in `simulate.py` that bypasses the check. Panel + hsgp is rejected at compile anyway.

The intersection semantics for a variable smoothed by several terms are the conservative choice and the right one.

**2. Divergences — the minimum is met.** `target_accept=0.99` in the example plus the raw-input-units note on `InverseGamma(3, 1)` covers the day-one failure mode, and #375 carries the real fix. Good.

**3. Parser naming — fixed.** `z:hsgp(...)` and `HSGP(...)` now raise hsgp-specific `ParseError`s before the transform registry, so the error names the right concept.

### One residual nit, follow-up only

`hsgp_intervention_bounds` is a second implementation of `set_boundary` rather than a read of the boundary actually frozen into the graph. The pinning test catches drift within the `pymc >= 6, < 7` pin, which is the right mitigation for now, but the moment the pin widens the two can diverge silently in exactly the direction that reintroduces this bug. If the compiled `gp` object is reachable from the model, reading `_X_center`/`L` off it at `do()` time would remove the duplication entirely. Not worth holding the PR.

The blocker I raised is cleared from my side. @drbenvincent still holds `CHANGES_REQUESTED`, so this needs your re-review to merge.

PR #368, opened 2026-07-31
Add unit-level counterfactual API

### LINKED ISSUE #29: Unit-level counterfactual API
## Context

The `do()` operator answers population-level interventional questions: "what happens on average if we set X to x?" Counterfactuals answer individual-level questions: "what would have happened to *this specific person* if X had been x, given what we've already observed about them?"

The difference comes down to the exogenous U terms. `do()` uses population means (U ≈ 0); counterfactuals first recover the individual's U values from their observed data (abduction), then intervene while holding those U values fixed.

**This issue is an API wrapper, not new functionality.** The three-step procedure already works manually in the docs example (see below). The goal is to replace ~20 lines of fragile, model-specific notebook code with `model.counterfactual(...)`.

## Existing documentation — do NOT create a new example notebook

> **For implementers:** The pedagogical example already exists. **Do not create a new docs notebook.** Resolve this issue by implementing the API and updating the existing page.

The counterfactual example was added in **#30** (merged PR #30):

- **Source:** `docs/examples/02-effect-estimation/counterfactual.qmd`
- **Live docs:** https://pathmc.pymc-labs.com/examples/effect-estimation/counterfactual.html
- **Title:** "Counterfactuals: From Population to Individual"

That notebook already:

1. Explains Pearl's three-step procedure (abduction → action → prediction)
2. Contrasts population-level `do()` with individual counterfactuals
3. Implements Joe's counterfactual manually using internal parameter naming (`beta_H`, `H_predictors`, etc.)

The manual implementation works but requires the user to:

1. Know the internal parameter naming convention
2. Extract coefficient and intercept draws manually via `.sel()` on the right coords
3. Implement the abduction step (algebraic inversion of structural equations)
4. Implement the prediction step (forward-propagating with modified equations)

That's ~20 lines of fragile, model-specific code. A dedicated API would make this accessible without knowledge of pathmc internals.

**Definition of done** includes updating this existing notebook to demonstrate the new API (see [Documentation updates](#documentation-updates) below). The manual section should be **kept** as "what's happening under the hood" teaching material.

## Proposed API

```python
result = model.counterfactual(
    evidence={"X": 0.5, "H": 1.0, "Y": 1.5},
    do={"H": 2.0},
)
result.mean("Y")           # posterior mean of Y_{H=2} for this individual
result.hdi("Y", prob=0.94) # HDI over the counterfactual
```

**Parameters:**

- `evidence`: dict mapping observed variable names to their values for this specific individual. Used in the abduction step to infer their U values.
- `do`: dict mapping variable names to intervention values (same format as `model.do(set=...)`).

**Returns:** `DoResult` (or similar) with the same `.mean()`, `.hdi()`, subtraction interface.

### What happens under the hood

For each posterior draw, the method performs Pearl's three-step procedure:

1. **Abduction**: Walk the DAG in topological order. For each endogenous variable V where evidence is provided, compute `U_V = evidence[V] - intercept_V - sum(beta_V[parent] * evidence_or_U[parent])`. This inverts the structural equation to recover the individual's exogenous value. For variables not in `evidence`, use U_V = 0 (population mean).

2. **Action**: Mark the `do` variables as intervened (their equations are replaced with constants).

3. **Prediction**: Walk the DAG in topological order again. For each non-intervened endogenous variable V, compute `Y_V = intercept_V + sum(beta_V[parent] * value[parent]) + U_V`, using the individual's U values from step 1 (not zeros). For intervened variables, use the `do` value directly.

The key difference from `do()`: in `run_do_pymc`, all U values are implicitly zero (population-level). In `counterfactual()`, U values are inferred from the individual's data and carried into the counterfactual world.

## Why this is NOT `pm.observe` + `pm.do`

This is a subtle but critical point. At first glance it seems like counterfactuals could be implemented by combining PyMC's built-in operations:

- `pm.observe(Y=1.5, H=1.0, X=0.5)` to condition on Joe's data
- `pm.do(H=2.0)` to intervene

**This does not work.** The fundamental problem is that counterfactuals operate across **two worlds** that share exogenous variables but differ in their structural equations:

| | Factual world | Counterfactual world |
|---|---|---|
| **Purpose** | Recover Joe's U values | Predict Joe's outcome |
| **H equation** | H = a·X + U_H | H = 2.0 (severed) |
| **Y equation** | Y = b·X + c·H + U_Y | Y = b·X + c·2 + U_Y |
| **U values** | Inferred from evidence | Carried over from factual |
| **Y** | Observed (1.5) — used as *input* | Computed — this is the *output* |

The problem is that Y plays opposite roles in the two worlds. In the factual world, Y = 1.5 is evidence (an input to abduction). In the counterfactual world, Y is the unknown quantity we want to compute (an output). `pm.observe(Y=1.5)` would condition Y in the generative model, but then `pm.do(H=2.0)` would try to compute Y from the modified equation — and you can't both observe and compute the same variable in the same model.

More precisely:

1. **`pm.observe` conditions the wrong thing.** `pm.observe(Y=1.5)` adds a likelihood term for Y in the *generative* model. But we don't want to constrain Y in the counterfactual world — we want to *predict* Y. What we actually need is to condition on Y in the *factual* world (to infer U_Y), then release Y in the counterfactual world.

2. **`pm.do` severs the wrong equations.** `pm.do(H=2.0)` replaces H's equation in the model. But in the abduction step, we need H's *original* equation to recover U_H from the evidence. We need the original model for abduction and the modified model for prediction — two passes, not one.

3. **The U values are not explicit RVs.** In pathmc's compiled model, U_V is implicit — it's the difference between the observed value and the predicted mean (`sigma * epsilon`, effectively). There's no named `U_Y` variable to condition on or carry between worlds. The abduction step must be computed externally from the posterior draws.

**In summary**: counterfactuals require a two-pass procedure (abduction on the original model, then prediction on the modified model) that cannot be expressed as a single `pm.observe` + `pm.do` operation. The `counterfactual()` method implements this two-pass logic using posterior draws directly, bypassing PyMC's graph surgery entirely.

### What about non-linear models? (future)

For linear Gaussian models, abduction is algebraic — simple subtraction, no sampling required. For non-linear or non-Gaussian models (Bernoulli, Poisson), algebraic inversion may not be possible. In that case, the abduction step would require approximate inference over U values (e.g. MCMC conditioning on the individual's evidence). This is substantially more complex and out of scope for the initial implementation.

### Core implementation

The logic lives in `simulate.py` (alongside `run_do_pymc`), called from a new `PathModel.counterfactual()` method in `_model.py`.

```python
def run_counterfactual(
    graph_info: GraphInfo,
    idata: az.InferenceData,
    evidence: dict[str, float],
    do: dict[str, float],
) -> DoResult:
    """Three-step counterfactual using posterior draws."""

    topo_order = graph_info.topological_order
    draws = _flatten_posterior(idata)  # shape (n_draws,) for each param

    # Step 1: Abduction — recover U values for each draw
    u_values = {}  # var -> array of shape (n_draws,)
    for var in topo_order:
        if var in graph_info.exogenous:
            u_values[var] = evidence.get(var, 0.0)
            continue
        if var not in evidence:
            u_values[var] = 0.0  # no evidence → population mean
            continue
        # Invert: U_var = evidence[var] - intercept - sum(beta * parent_value)
        intercept = draws[f"beta_{var}"].sel(Intercept)
        parent_contrib = sum(
            draws[f"beta_{var}"].sel(parent) * evidence[parent]
            for parent in structural_parents(var)
        )
        u_values[var] = evidence[var] - intercept - parent_contrib

    # Step 2 + 3: Action + Prediction — forward-propagate with U values
    predicted = {}
    for var in topo_order:
        if var in do:
            predicted[var] = do[var]
            continue
        if var in graph_info.exogenous:
            predicted[var] = evidence.get(var, 0.0)
            continue
        # Y_var = intercept + sum(beta * predicted[parent]) + U_var
        intercept = draws[f"beta_{var}"].sel(Intercept)
        parent_contrib = sum(
            draws[f"beta_{var}"].sel(parent) * predicted[parent]
            for parent in structural_parents(var)
        )
        predicted[var] = intercept + parent_contrib + u_values[var]

    return DoResult(predicted)
```

This is ~40 lines of core logic. The structural parent lookup uses `graph_info` (already available). The coefficient extraction mirrors what `effects.py` already does.

### Edge cases and validation

- **`do` variable also in `evidence`**: Allowed. The evidence for that variable is used in abduction (to infer U values of downstream variables in the factual world), but its value is replaced by the `do` value in the prediction step. This is the standard case (Joe's H is in both evidence and do).
- **Partial evidence**: Variables not in `evidence` get U = 0 (population mean). The method should issue a warning if the `do` target or any of its ancestors lack evidence, since the counterfactual is then mixing individual and population information.
- **Pre-sample check**: Must have called `.fit()` first. Same guard as `.do()`.
- **Exogenous variables in evidence**: Stored as-is (exogenous variables have no structural equation, so U_X = X by definition).
- **Interaction terms**: If the model has `Y ~ X + X:Z`, the parent contribution must handle interaction columns in the design matrix. Use the same coefficient/design logic as `effects.py`.

### Must have

- [ ] `model.counterfactual(evidence=..., do=...)` method on `PathModel`
- [ ] Works for linear Gaussian models (the standard case)
- [ ] Returns a `DoResult`-compatible object
- [ ] Validates that `evidence` keys are variables in the model
- [ ] Validates that `do` keys are endogenous variables with structural equations
- [ ] Warns if evidence is missing for ancestors of the `do` target
- [ ] **Update the existing** `docs/examples/02-effect-estimation/counterfactual.qmd` to demonstrate the new API (see below) — **not** a new notebook

### Should have

- [ ] Support for partial evidence (not all variables observed — use population mean for unobserved U values)
- [ ] Clear error message if a non-Gaussian family is used (with guidance that it's not yet supported)
- [ ] Verify API output matches the manual implementation in the existing notebook (Joe's counterfactual ≈ 1.90)

### Out of scope (future issues)

- [ ] Non-Gaussian families (Bernoulli, Poisson — abduction requires approximate inference)
- [ ] Panel/longitudinal counterfactuals
- [ ] `counterfactual_contrast()` convenience for comparing two counterfactual scenarios for the same individual
- [ ] **A new example notebook** — the existing counterfactual page is sufficient

## Documentation updates

> **Do not create a new example notebook.** Update the existing page from #30.

The existing `docs/examples/02-effect-estimation/counterfactual.qmd` is the sole docs deliverable for demonstrating counterfactuals. It already has excellent pedagogical content. The update would:

1. **Add a new section** after the manual implementation that shows the same computation using the new API:

   ```python
   result = model.counterfactual(
       evidence={"X": 0.5, "H": 1.0, "Y": 1.5},
       do={"H": 2.0},
   )
   print(f"E[Y_{{H=2}}]:  {result.mean('Y'):.3f}")
   print(f"94% HDI:       {result.hdi('Y', prob=0.94)}")
   ```

2. **Keep the manual implementation** as a "What's happening under the hood" section — it's excellent teaching material and should stay.

3. **Add a callout** explicitly noting that this is not `pm.observe` + `pm.do` and why (the two-worlds argument).

4. **Optionally** update `docs/concepts/causal_inference.qmd` to add a brief "Counterfactuals" subsection distinguishing `do()` from `counterfactual()`, linking to the example.

5. Re-freeze the notebook cache after edits (`uv run great-docs freeze docs/examples/02-effect-estimation/counterfactual.qmd`).

## References

- Pearl, Glymour, Jewell (2016), *Causal Inference in Statistics: A Primer*, §4.2.3–4.2.4
- #30 — existing counterfactual docs example (merged)
- `docs/examples/02-effect-estimation/counterfactual.qmd` — working manual implementation and pedagogical explanation
- https://pathmc.pymc-labs.com/examples/effect-estimation/counterfactual.html — live docs page

drbenvincent:
## Summary

- Add `model.counterfactual(evidence=..., do=...)` implementing Pearl's three-step procedure (abduction → action → prediction) on posterior draws for linear Gaussian models
- Return a `DoResult` with the same `.mean()`, `.hdi()`, and contrast interface as `do()`
- Update the existing counterfactual docs example to demonstrate the API while keeping the manual implementation as teaching material

Closes #29

## Test plan

- [x] `uv run pytest tests/test_counterfactual.py -v`
- [x] `make lint`
- [x] Re-freeze `docs/examples/02-effect-estimation/counterfactual.qmd` cache
- [ ] CI green


drbenvincent (commented):
Should this new API be used in any other existing docs?

daimon-pymclabs (changes requested):
I ran this branch on a dev session and probed it rather than reading only. `tests/test_counterfactual.py` is 10/10 green.

The three-step math is correct for the case the docs demonstrate: fully observed, additive, Gaussian, cross-sectional. The strongest invariant holds there — `do(H = observed H)` returns exactly the observed `Y`. Outside that case the implementation tends to return a plausible number rather than raise, and for an API whose whole selling point is "this is not `do()`", the specific way it degrades is the problem.

## 1. Incomplete `evidence` silently gives wrong answers

`_slot_column_value` falls back to `values.get(slot.name, 0.0)`. For a plain additive term the zero-fill cancels between the abduction and the prediction pass, so the answer survives by luck. It does not cancel for interaction terms, and it does not cancel when the outcome itself is missing from `evidence` — verified both on the branch. In the missing-outcome case you get the population `do()` answer wearing a counterfactual label, which is exactly the confusion the new callout box exists to prevent.

The ancestor warning only covers `ancestors(do_target) ∪ {do_target}`, so a moderator that isn't an ancestor of the intervention target never triggers it. It should cover the ancestor sets of the outcomes too. More strongly: a missing variable should raise by default, with partial evidence as an explicit opt-in. A `.get(name, 0.0)` in a causal kernel is a footgun; a `KeyError` here is a feature.

## 2. `do=` rejects exogenous variables, ruling out the most natural counterfactual

In `H ~ X; Y ~ X + H`, "what if this unit's `X` had been 1.5?" raises `Cannot intervene on 'X': it is not endogenous`. The restriction isn't load-bearing — `if var in do` is the first branch of the prediction loop and `predicted[var] = np.full(n_draws, do[var])` never touches a structural equation, so deleting the check just works. It's also a divergence from `do()`, which has no such limit.

## 3. Latent-variable models return wrong numbers instead of erroring

A latent variable can never appear in `evidence`, so it always reads as 0.0 during abduction. On `M ~ X; Y ~ M` with `latent=["M"]`, the implied `u_Y` at `do(M=0.4)` came back as exactly the observed `Y`, confirming `M` was treated as 0; the latent path is then double-counted in prediction. The warning that does fire tells the user to add `M` to `evidence`, which is impossible. This wants a `NotImplementedError` on `graph_info.latent`, beside the panel guard.

## 4. No extrapolation warning

`do(set={"H": 500})` warns that 500 is outside the observed range. `counterfactual(do={"H": 500})` is silent. Same hazard, and worse for a unit-level claim.

## 5. Interactions bypass the lag/transform guard

`_slot_column_value` raises for `kind in ("lag", "transform")`, but the `interaction` branch multiplies `values.get(part, 0.0)` over `interaction_parts` without checking whether those parts are themselves transforms or lags. Zeroed instead of rejected.

## 6. Duplicated linear-predictor evaluation

`_linear_predictor_draws` re-implements in numpy what `build_mu` does in pytensor. Any future `PredictorSlot.kind` will silently produce wrong counterfactuals rather than fail to compile, because the fallback is a `.get()` rather than an exhaustive match. Make the slot dispatch exhaustive with `else: raise`.

## Tests

The fixture overwrites the posterior betas with the true values, so the two `1e-10` assertions compare the arithmetic against itself rather than against inference. That's reasonable as a gate test, worth a comment saying so. The gaps that matter are the ones above, plus a `do(V = observed V)` identity test — one-liner, and it currently passes.

## Reusability elsewhere in the docs

Answering the review question in the PR description: no, the API can't be used in any other docs page as it stands. `did.qmd`, `mmm.qmd`, and `mmm_awareness.qmd` are panel models blocked by the panel guard, and `ate_estimation.qmd` is population-level g-computation. It applies to exactly one page, which is itself a scope signal.

## Nits

- `ancestors = nx.ancestors(...)` at `pathmc/simulate.py:709` shadows the module-level `from pytensor.graph.traversal import ancestors`.
- The warning can name a variable as its own ancestor, and says "which are" for a single variable.
- The new import block is unsorted; ruff's `I` rules aren't selected so lint stays green.
- The signature says `dict[str, float]` but the helpers handle `np.ndarray` — either tighten it or TODO it.
- A `DoResult` from `counterfactual()` will happily `__sub__` with one from `do()` to produce a meaningless contrast.

## Verdict

I'd want 1 and 3 fixed before merge. 2 is a small change with a large usability payoff; 4 and 5 are cheap.

drbenvincent (commented):
Implemented in a063f58.

- Evidence is now complete-by-default; `allow_partial_evidence=True` is the explicit, warning-emitting opt-in.
- Exogenous interventions are allowed; latent, panel, non-Gaussian, lagged, and transformed cases now reject clearly where unsupported.
- Counterfactuals share `do()` extrapolation warnings, use exhaustive predictor dispatch, and reject meaningless counterfactual-vs-population contrasts.
- Added identity, interaction, latent, exogenous-intervention, extrapolation, and partial-evidence regression coverage.

On the documentation question: no other existing example is appropriate today. The related MMM/DID pages are panel models, and ATE estimation is population-level; the existing counterfactual page remains the sole applicable example.

Validated with `make lint` and 128 passing focused regression tests.

drbenvincent (commented):
Review by GPT-5.6 Terra High

I reviewed final head a063f58, including the updated example, implementation, and focused tests. The prior correctness blockers are addressed: complete evidence is the default; unsupported latent/panel/non-Gaussian/lag/transform cases reject; exogenous interventions, interactions, extrapolation warnings, and contrast provenance are handled. The teaching flow is clear: concrete individual question -> manual three-step derivation -> minimal API call. The figures have captions, stable labels, semantic colors, and a direct comparison.

Requested documentation follow-up: add one concise support-boundary note beside the new API call. It should say that this release supports fully observed, cross-sectional Gaussian linear SEMs, and name the rejected categories (panel, latent, non-Gaussian, transform/lag). The example says the method removes the need to know internals, but currently never tells a reader whether their model is eligible. This can remain a one-sentence note or compact warning callout. Also qualify same contrast interface: contrasts are valid only between counterfactuals for identical evidence, not between a counterfactual and a population do result.

Test assessment: the suite is substantially more than a happy-path check (identity, partial evidence, exogenous intervention, interaction, latent/non-Gaussian, extrapolation, pre-fit, and provenance cases). Please add direct regression tests for the documented lag/transform rejections, plus the positive same-evidence counterfactual contrast. They protect the stated support boundary and ensure contrast metadata remains preserved.

CI note: docs and prek are green; the test job was still pending when I reviewed.

drbenvincent:
Follow-up for the remaining review requests, pushed in 08ca7ce:

- Added a concise support-boundary note beside the counterfactual API example: fully observed, cross-sectional Gaussian linear SEMs are supported; panel, latent, non-Gaussian, transform, and lag models are not yet supported.
- Clarified that counterfactual contrasts require identical evidence and cannot be mixed with population-level `do()` results.
- Added regression coverage for same-evidence counterfactual contrasts and rejected transform/lagged-panel models.

Validation:
- `uv run python -m pytest tests/test_counterfactual.py -x -q` — 19 passed
- `make lint` — passed

drbenvincent:
Maybe needs an eyeball test of the docs page

PR #369, opened 2026-07-31
Run pytest explicitly in the uv dev environment

drbenvincent:
## Summary
- Make `test-fast` and `test` explicitly install the `dev` extra and invoke pytest via the uv-managed Python interpreter.
- Update `AGENTS.md` and `CONTRIBUTING.md` so targeted test commands cannot silently fall back to a global pytest executable.

## Test plan
- [x] `make test-fast` — 889 passed, 176 deselected
- [x] `make lint`


daimon-pymclabs (commented):
Reviewed the diff and checked the surrounding config (`pyproject.toml`, `Makefile`, `.github/workflows/ci.yml`) plus uv's documented sync semantics. The change is correct and the failure mode it targets is real, but the fix is applied to 2 of ~6 equally-exposed call sites, and there's a config-level fix that covers all of them.

**Confirmed non-issue first**, since it was my initial worry: `uv run --extra dev` will *not* uninstall the `samplers` extra. Per the [uv docs](https://docs.astral.sh/uv/concepts/projects/sync/), `uv run` uses *inexact* syncing by default ("ensuring that all required packages are installed but not removing extraneous packages"), while `uv sync` is exact. So `tests/test_nutpie_scan_compile.py` (which does `pytest.importorskip("nutpie")`) still runs for anyone who has run `make setup`. No silent-skip regression.

### 1. The same hole is still open in `lint`, `check_lint`, and `docs`

If `uv run pytest` can silently fall back to a global `pytest`, then so can every other tool invocation in the `Makefile`, because they all come from the same extras:

- `lint:` → `uv run prek` (`dev` extra)
- `check_lint:` → `uv run ruff`, `uv run mypy` (`dev` extra)
- `docs:` → `uv run great-docs` (`docs` extra)

A stale global `ruff` or `mypy` silently passing a lint gate is arguably a worse outcome than a stale global `pytest`, since the failure is a false green rather than an import error. Hardening only the two test targets leaves the more dangerous cases untouched.

### 2. A config-level fix would cover all of them with no per-command flags

uv has no `[tool.uv] default-extras` (I checked the settings reference — only `default-groups` exists, for PEP 735 dependency groups). But moving the dev tooling from an extra to a dependency group gets exactly the behaviour you want:

```toml
[dependency-groups]
dev = ["build", "ipykernel", "mypy", "polars", "prek", "pytest", "pytest-cov", "ruff"]
```

uv installs the `dev` group by default on every `uv run` and `uv sync`, so `uv run pytest`, `uv run ruff`, and `uv run mypy` are all guaranteed to resolve inside `.venv/` with no flag anywhere. The `Makefile` and both docs files then revert to their current, simpler wording.

The tradeoff: `pip install pathmc[dev]` stops working, since dependency groups aren't exposed as extras to pip. For a dev-tooling extra on a library that's unlikely to matter, but worth a deliberate decision rather than an accident. If you want to keep it, a thin `dev = []` extra plus the group is possible, though it's not obviously worth the duplication.

### 3. CI and the `Makefile` now diverge

`.github/workflows/ci.yml:29` still runs `uv run pytest -x -v --cov=pathmc --cov-report=term-missing` directly, not `make test`. After this PR, CI and local `make test` are no longer the same command. That's the specific kind of drift that makes "works locally, fails in CI" hard to debug. Either point CI at `make test` (which also keeps them in sync automatically going forward) or apply the same flag there.

### 4. The docs now carry a two-rule split

`AGENTS.md` and `CONTRIBUTING.md` go from one rule ("prefix with `uv run`") to a conditional one ("for tests, use `uv run --extra dev python -m pytest`; for scripts, use `uv run python -c ...`"). That's harder to internalise and more likely to be misremembered than the thing it replaces, especially in `AGENTS.md` where the whole point is an unambiguous instruction. The dependency-group fix removes the need for the split entirely.

### 5. Minor: `python -m pytest` puts the CWD on `sys.path`

`python -m pytest` prepends the working directory to `sys.path`; the `pytest` console script doesn't. With pathmc's flat layout (`pathmc/` at the repo root) and an editable install, `import pathmc` resolves to the same working-tree directory either way, so there's no practical difference today. Flagging it only because it makes the test run marginally less faithful to what an installed user gets, and it's the sort of thing that bites later if the layout ever moves to `src/`.

### Verdict

Nothing here is wrong, and if you want it merged as-is it's a strict improvement over the status quo. My suggestion is to spend the same effort on the dependency-group migration instead: it's a smaller diff in the `Makefile` and docs (both revert), it fixes `lint`/`check_lint`/`docs` too, and it removes the conditional instruction from `AGENTS.md` rather than adding one.

daimon-pymclabs:
Clarification on point (2), since "move dev tooling to a dependency group" can read as though it would push dev dependencies onto end users. It does the opposite — dependency groups are *less* exposed to end users than extras, not more.

[PEP 735](https://peps.python.org/pep-0735/) is explicit: "Build backends MUST NOT include Dependency Group data in built distributions as package metadata." Groups never reach the wheel, so `pip install pathmc` from PyPI cannot see or install them under any invocation. They are a source-tree-only concept, resolved by tooling operating on the project directory (uv, here). Extras are the published half of this: `[project.optional-dependencies]` *does* land in wheel metadata, which is precisely what makes `pip install pathmc[dev]` work today.

What actually changes:

| | today (`dev` extra) | after (`dev` group) |
| --- | --- | --- |
| `pip install pathmc` | no dev deps | no dev deps |
| `pip install pathmc[dev]` | installs them | fails, extra no longer exists |
| `uv run` / `uv sync` in a checkout | requires `--extra dev` | automatic |

So a plain `pip install pathmc` never pulls in ruff and mypy either way. The only thing that breaks is the deliberate `pathmc[dev]` opt-in, which is the tradeoff noted in the original review. That matters only if something currently depends on it — a downstream CI job, or a contributor not using uv. The pip 25.1+ equivalent is `pip install --group dev`, but it only works against a source checkout, not against the published package.

Net effect on end users: today the dev tooling is published metadata that anyone can deliberately pull down; after the migration it isn't published at all.

drbenvincent:
Addressed the review feedback in commit 2cdfa01: moved development tooling from the published dev extra to uv's default dev dependency group, restored the simple uv-run commands across the Makefile and contributor guidance, and changed CI to run the same make test target used locally. The lockfile is regenerated and uv lock --check passes.

Issue #282, opened 2026-06-15
Support residual covariance (~~) in falsify() via canonical-DAG m-separation


cetagostini:
### Summary

`falsify()` currently **rejects** models that declare residual covariances (`~~`) with a clear error (PR #275), because a permutation test over a plain DAG cannot correctly handle bidirected/ADMG confounding. This issue tracks adding proper support.

### Why the simple approach is wrong

An earlier attempt that merely *excluded the literally-declared `~~` pair* from the implied independences was shown (in review) to be incorrect: a confounder with descendants opens dependencies the exclusion misses (e.g. `X ~~ Y`, `X -> Z`, `Y -> T` still wrongly tests `Y ⊥ Z` and `X ⊥ T`), and the permutation/MEC oracle could misclassify equivalence. It also crashed on `~~` endpoints that aren't directed-DAG nodes.

### Correct approach

Use the **latent-projection (canonical) DAG**: replace each `~~` pair `A <-> B` with a latent common cause `L -> A`, `L -> B`, then test implied independences with **m-separation** (equivalently, d-separation on the canonical DAG with latents never in conditioning sets). This must be applied consistently to:

- the implied-independence (LMC) enumeration for the given graph,
- the permutation baseline (relabel the `~~` structure alongside directed edges), and
- the tPA d-separation oracle (reference and permuted graphs both use their canonical forms).

### Scope / care

- The permutation-test semantics over ADMGs are not something dowhy validated — design and justify the baseline carefully (what counts as an MEC under `~~`).
- Ideally share the ADMG-aware machinery with the `test_implications` `~~` fix (#277).
- Handle `~~` endpoints that are not directed-DAG nodes (the case that previously KeyError'd).

### Acceptance criteria

- [ ] `falsify()` accepts `~~` models and computes correct implied independences (verified on confounder-with-descendants cases).
- [ ] Permutation baseline + MEC counting are correct and documented for the ADMG case.
- [ ] The current rejection error is removed only once the correct path is in place.
- [ ] Tests covering direct `~~`, confounder-with-descendants, multi-pair blocks, and dangling endpoints.

### Context

- Current rejection: `pathmc/falsify.py` (guard on `graph_info.residual_blocks`), PR #275.
- Related edge-by-edge limitation: #277.

drbenvincent:
**Composer 2.5 automated comment**

Overlap check against existing issues: meaningful overlap with #14 (open) and coordination need with #277.

[...]

Issue #279, opened 2026-06-15
Consistent significance-parameter naming across the causal-testing API


cetagostini:
### Summary

The causal-testing surface uses two different vocabularies for the same concept:

- `PathModel.test_implications(alpha=0.05)`
- `PathModel.falsify(significance_level=0.05, significance_ci=0.05)` (the latter mirror dowhy)

Users must learn both. We should pick a convention and document the mapping.

### Options

1. **Standardize on `alpha`** everywhere, keeping `significance_level`/`significance_ci` as documented aliases on `falsify` (deprecate slowly). Pro: internal consistency. Con: diverges from dowhy's published names.
2. **Keep `significance_level`/`significance_ci` for `falsify`** (faithful to the ported method) and add a prominent note in both docstrings + the user guide explaining that `alpha` ≡ `significance_ci` (the CI-test threshold) and `significance_level` is the *permutation-test* threshold (no analogue in `test_implications`).

### Recommendation

Option 2 with clear cross-referencing is lowest-risk; `falsify` genuinely has *two* thresholds (per-CI vs. whole-graph permutation), so collapsing them onto a single `alpha` would lose meaning. The key deliverable is documentation that makes the relationship obvious.

### Acceptance criteria

- [ ] Decision recorded (issue comment or short ADR).
- [ ] Both docstrings cross-reference the equivalence.
- [ ] User-guide/identification docs explain the two thresholds in one place.
- [ ] If aliases are added, they are tested and documented.

Issue #277, opened 2026-06-15
implied_independences()/test_implications() ignore residual covariance (~~), mis-flagging declared confounding


cetagostini:
### Summary

`implied_independences()` enumerates testable conditional independences from the **directed** edges only; it never consults the residual-covariance (`~~`) structure. A pair declared as confounded with `X ~~ Y` is therefore emitted as an implied independence and reported as a **violation** by `test_implications()` — even though the user explicitly stated the two variables share unobserved common causes and are *not* expected to be independent.

### Where

- `pathmc/identify.py:462` — `implied_independences` (uses `graph_info.contemporaneous_dag`, checks `dag.has_edge`, never reads `graph_info.residual_blocks`)
- consumed by `pathmc/identify.py` `test_implications` and `PathModel.test_implications`

### Why it matters

`~~` encodes a bidirected/ADMG edge (latent common cause). Such a pair is not d/m-separated by any observed conditioning set, so it must be excluded from the implied-independence basis set. Today, a perfectly-specified model with correlated residuals will show spurious "violations".

### Reproduction

```python
import numpy as np, pandas as pd, pathmc
n = 1000; rng = np.random.default_rng(31)
U = rng.normal(size=n)
X = 0.9*U + rng.normal(scale=.4, size=n)
Y = 0.9*U + rng.normal(scale=.4, size=n)
W = 0.5*X + 0.5*Y + rng.normal(scale=.5, size=n)
df = pd.DataFrame({"X": X, "Y": Y, "W": W})

m = pathmc.model("W ~ X + Y\nX ~~ Y", data=df)
print(m.test_implications())   # reports X ⊥ Y as a violation, despite X ~~ Y
```

### Proposed approach

- Exclude any pair that lies in the same residual-covariance block from the implied-independence enumeration.
- For full correctness with confounders that have *descendants* (e.g. `X ~~ Y`, `X -> Z`, `Y -> T`), the dependence opened through the latent must also be respected — this requires m-separation over the latent-projection (canonical) DAG, not just dropping the literal pair. See the related falsification ADMG issue; ideally both features share one ADMG-aware helper.
- Decide and document the scope: either (a) full ADMG/m-separation support, or (b) a clear limitation + warning when `~~` is present.

### Acceptance criteria

- [ ] A correctly-specified model with `~~` does not produce spurious independence violations for the confounded pair.
- [ ] Behavior for confounders with descendants is correct (m-separation) or explicitly documented as unsupported.
[...]

PR #374, opened 2026-08-05
Unify the partial-correlation CI engine on a shared rank-aware core (#276, #278)

### LINKED ISSUE #276: Partial-correlation CI test uses wrong degrees of freedom for constant/collinear conditioners
### Summary

`identify._partial_correlation_test` computes the t-test degrees of freedom from the *number* of conditioning variables (`df = n - k - 2`) rather than the **effective rank** of the conditioning design. When a conditioning variable is constant, or two conditioners are collinear/duplicated, this over-deflates `df` and can change a borderline `significant`/not-significant decision in `test_implications()`.

### Where

- `pathmc/identify.py:587` — `_partial_correlation_test`
- `pathmc/identify.py:627` — `df = n - k - 2`

### Why it's wrong

Conditioning on a constant column should be equivalent to the marginal test (the constant collapses into the intercept), and a duplicated conditioner adds no information. With `df = n - k - 2`, both cases use too few degrees of freedom, inflating the t-statistic's tail and shifting the p-value.

### Reproduction

```python
import numpy as np, pandas as pd, narwhals.stable.v1 as nw
from pathmc.identify import _partial_correlation_test

rng = np.random.default_rng(0); n = 60
x = rng.normal(size=n); y = 0.27 * x + rng.normal(size=n); z = np.ones(n)
df = nw.from_native(pd.DataFrame({"X": x, "Y": y, "Z": z}), eager_only=True)

r_m, p_m, _ = _partial_correlation_test(df, "X", "Y", [])        # marginal
r_c, p_c, _ = _partial_correlation_test(df, "X", "Y", ["Z"])     # conditioned on a constant
# p_m and p_c should be (essentially) identical; today they differ.
```

### Proposed fix

Mirror the fix already shipped in `pathmc/falsify.py:428` (`_PartialCorrelationTester._compute`):

```python
z_with_intercept = np.column_stack([np.ones(n), z_mat])
rank = int(np.linalg.matrix_rank(z_with_intercept))
df = n - rank - 1
if df <= 0:
    return np.nan, np.nan, n
```

At full rank this reproduces the current `n - k - 2`; for constant/collinear conditioners it gives the correct, larger `df`.

### Notes

- The same logic was independently fixed in the new `falsify` module (PR #275); this issue is to bring `identify` in line. See also the "unify the CI engine" improvement, which would resolve both in one place.
- **Do not** edit test files; add new tests under `tests/test_identification.py`-adjacent coverage if needed (confirm with maintainers).

### Acceptance criteria

- [ ] `test_implications()` returns identical p-values for a marginal test and the same test conditioned on a constant column.
- [ ] Duplicated/collinear conditioners do not change the p-value vs. the unique conditioner.
- [ ] No regression in existing `test_identification.py`.

### LINKED ISSUE #278: Unify the partial-correlation CI engine shared by identify and falsify
### Summary

There are now two independent partial-correlation conditional-independence implementations:

- `pathmc/identify.py:587` — `_partial_correlation_test` (used by `test_implications`)
- `pathmc/falsify.py` — `_PartialCorrelationTester` (used by `falsify`, added in PR #275)

The `falsify` version has guards the `identify` version lacks: zero-variance columns → skip, perfect collinearity → p=0, NaN handling, and **effective-rank degrees of freedom**. The two can (and already do) drift, and the same bug had to be fixed twice.

### Goal

Extract one well-tested, guarded partial-correlation CI function (e.g. in a small `pathmc/_ci.py` or as a shared helper) and have both `identify` and `falsify` call it. Resolving this also fixes #276 (df bug) in one place.

### Design notes

- Keep the public behavior of `test_implications()` stable (it returns `(r, p, n)`); the shared core can return the p-value with the callers adapting.
- Preserve `falsify`'s memoization (cache keyed by `frozenset({x, y})` + `frozenset(Z)`), which relies on symmetry of CI tests.
- Centralize the guards: `n < 3`, `df <= 0`, zero residual variance, `|r| >= 1 → p = 0`, NaN.

### Acceptance criteria

- [ ] One implementation; `identify` and `falsify` both delegate to it.
- [ ] Effective-rank df (closes #276) lives in the shared code.
- [ ] No behavior change for well-conditioned inputs; existing suites pass.
- [ ] Unit tests for the shared helper cover marginal, conditional, constant, collinear, NaN, and degenerate cases.

### Dependencies

- Closes/субsumes #276. Coordinate with #277 if a shared ADMG-aware layer is introduced.

thomaspinder:
## Summary

Closes #276. Closes #278.

pathmc had two independent partial-correlation CI implementations — `identify._partial_correlation_test` (behind `test_implications()`) and `falsify._PartialCorrelationTester` (behind `falsify()`) — which had drifted: the falsify version gained rank-aware degrees of freedom and degenerate-input guards that identify lacked. `test_implications()` could therefore return shifted p-values for constant or collinear conditioners (#276) and NaN/`RuntimeWarning` noise on degenerate input. This PR extracts one hardened engine into a new private module `pathmc/_ci.py` and has both callers delegate to it.

## Changes

- **`pathmc/_ci.py` (new)** — `partial_correlation_ci(x, y, z) -> CIResult`: array-level core with NaN-row dropping, effective-rank degrees of freedom (`n - rank([1, Z]) - 1`), and named skip reasons on a frozen result. `_PartialCorrelationTester` (the memoizing front-end) moves here from `falsify.py`; its `_compute` now delegates to the core.
- **`pathmc/identify.py`** — `_partial_correlation_test` becomes a thin adapter over the core, preserving the `(r, p, n)` contract and the `test_implications()` result schema; skipped tests surface as `(nan, nan, n)` rows as before.
- **`pathmc/falsify.py`** — ~95 duplicated lines removed; behavior unchanged.
- **`tests/test_ci_engine.py` (new)** — characterization pins recorded from both implementations *before* the refactor (first commit), unit tests for every guard and skip reason including the #276 reproducer, and a permanent identify↔falsify parity test so the engines cannot silently drift again. (P-value pins use `pytest.approx(..., abs=0.0)`: the default absolute tolerance of 1e-12 would trivially accept any value at the ~1e-34 magnitudes involved.)

## Intentional behavior changes (identify side only)

1. **Rank-aware df** — conditioning on a constant/duplicated/collinear column now behaves as if the redundant column were absent (the #276 fix).
2. **Perfect correlation (r² ≥ 1)** yields `p = 0.0` (a flagged violation) instead of NaN via divide-by-zero.
3. **Small-sample refusal** is rank-aware (`df ≤ 0`) rather than nominal (`n < k + 3`), so rank-deficient conditioning sets no longer forfeit observations.
4. Degenerate inputs no longer emit numpy `RuntimeWarning`s.

`falsify()` behavior is unchanged; the delegated `_compute` reproduces the previous semantics case-for-case, including the exact-zero variance guards.

## Verification

- Characterization tests pinning pre-refactor p-values from both implementations pass unchanged after the refactor.
- Full local suite: 1277 passed, including the slow MCMC tier.
- #278 acceptance criteria: one implementation; rank-aware df in shared code (closes #276); no behavior change for well-conditioned inputs (demonstrated by the recorded pins); unit tests cover marginal, conditional, constant, collinear, NaN, and degenerate cases.

## Out of scope

- `alpha` vs `significance_*` naming and the `<` vs `<=` violation convention (#279).
- `~~`/ADMG-aware implication enumeration (#277 / #282) — the core is pure statistics, so that work can layer on top.

### DIFF
--- a/pathmc/_ci.py
+++ b/pathmc/_ci.py
@@ -25,10 +25,13 @@ named skip instead of a NaN or a runtime warning.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Literal
+from typing import TYPE_CHECKING, Literal
 
 import numpy as np
 from scipy import stats
+
+if TYPE_CHECKING:
+    import narwhals.stable.v1 as nw
 
 CISkipReason = Literal[
     "insufficient_observations",
@@ -135,3 +138,63 @@ def partial_correlation_ci(
     t_stat = r * np.sqrt(df) / np.sqrt(1.0 - r * r)
     p = float(2.0 * stats.t.sf(np.abs(t_stat), df))
     return CIResult(r=r, p=p, n=n, df=df, skip_reason=None)
+
+
+class _PartialCorrelationTester:
+    """Memoized partial-correlation conditional independence tester.
+
+    Holds a numeric matrix extracted once from the data and caches each
+    ``X ⊥⊥ Y | Z`` p-value. Independence is symmetric in ``X`` and ``Y``,
+    so the cache key normalizes their order, mirroring dowhy's
+    ``_PValuesMemory`` and avoiding recomputation across the many
+    permuted graphs that re-test the same triples.
+    """
+
+    def __init__(self, data: nw.DataFrame, variables: list[str]) -> None:
+        columns = set(data.columns)
+        numeric: list[str] = []
+        arrays: list[np.ndarray] = []
+        for var in variables:
+            if var not in columns:
+                continue
+            try:
+                col = data.select(var).to_numpy().astype(float).ravel()
+            except (ValueError, TypeError):
+                # Non-numeric (string/categorical) columns cannot enter a
+                # partial-correlation test; treat them as unavailable so the
+                # affected triples are skipped, exactly like a missing column.
+                continue
+            numeric.append(var)
+            arrays.append(col)
+        if arrays:
+            matrix = np.column_stack(arrays)
+        else:
+            matrix = np.empty((0, 0), dtype=float)
+        self._matrix = matrix
+        self._col_idx = {name: i for i, name in enumerate(numeric)}
+        self._cache: dict[tuple[frozenset[str], frozenset[str]], float | None] = {}
+
+    def p_value(self, x: str, y: str, z_vars: tuple[str, ...]) -> float | None:
+        """Return the CI test p-value, or ``None`` if it cannot be run.
+
+        ``None`` signals a skipped test: a required variable has no data
+        column, there are too few complete observations, or the test is
+        otherwise degenerate (see :class:`CIResult`).
+        """
+        key = (frozenset((x, y)), frozenset(z_vars))
+        if key in self._cache:
+            return self._cache[key]
+        result = self._compute(x, y, z_vars)
+        self._cache[key] = result
+        return result
+
+    def _compute(self, x: str, y: str, z_vars: tuple[str, ...]) -> float | None:
+        needed = [x, y, *z_vars]
+        if any(v not in self._col_idx for v in needed):
+            return None
+        cols = [self._col_idx[v] for v in needed]
+        arr = self._matrix[:, cols]
+        result = partial_correlation_ci(arr[:, 0], arr[:, 1], arr[:, 2:])
+        if result.skip_reason is not None:
+            return None
+        return result.p
--- a/pathmc/_ci.py
+++ b/pathmc/_ci.py
@@ -1,0 +1,137 @@
+#   Copyright 2025 - 2026 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Shared partial-correlation conditional-independence engine.
+
+Single source of truth for the CI test behind both
+:func:`pathmc.identify.test_implications` (edge-by-edge implication
+testing) and :func:`pathmc.falsify.falsify_graph` (whole-graph
+falsification). Degrees of freedom follow the *effective rank* of the
+conditioning design, so constant, duplicated, or collinear conditioners
+behave as if they were absent, and every degenerate input maps to a
+named skip instead of a NaN or a runtime warning.
+"""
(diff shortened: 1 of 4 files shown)


daimon-pymclabs (commented):
Non-blocking review. I checked this empirically rather than by eye, and it looks sound to me — happy for it to merge as-is. Everything below is for your judgment, @thomaspinder; none of it blocks.

**What I verified.** I diffed the pre-PR `falsify._compute` and pre-PR `identify._partial_correlation_test` line-by-line against the new shared core, then ran both old implementations alongside the new one through every executable docs page that touches these APIs. 22 identify calls and 123 falsify calls, zero numeric disagreements.

**Docs.** No refreeze needed and no prose changes required. Only `dag_testing.qmd`, `dag_falsification.qmd` and `the_house.qmd` have executable `{python}` chunks hitting these APIs, and all three produced bit-identical results under old and new code. The `test_implications`/`falsify` mentions in `docs/user_guide/00-welcome`, `01-how-it-works`, `02-comparison`, `13-causal-inference` and `18-causal-discovery` are static ```python blocks Quarto never executes, so there is no frozen output to invalidate. Nothing in the qmd files or the `test_implications` docstring states the small-sample rule (`n < k + 3`) or the df convention, so no existing text is now wrong.

**Three behaviour changes worth a deliberate call:**

1. *Perfect correlation now flags a violation.* In identify, `r² >= 1` previously yielded NaN (not significant); it now yields `p = 0.0`, so `significant = True`. Defensible either way — the question is whether a perfectly collinear pair should read as a DAG violation or as a data-quality problem surfaced separately.

2. *Rank collapse is silent.* When a conditioning set is rank-deficient the test proceeds as though the redundant column were absent. Correct for df, but the user is never told their intended adjustment set was effectively smaller, which matters more here than it would in a generic stats library. `CIResult` already carries `df`; the identify adapter drops it and `test_implications()` has no column for it. Surfacing `df` or an `effective_k` looks like a cheap follow-up issue.

3. *Small-sample refusal is looser.* `df <= 0` replaces `n < k + 3`, so rank-deficient sets now return a p-value where they previously returned NaN. That is the intended #276 fix — just noting that some previously-silent tests now emit numbers computed off very few observations.

**Two minor observations, neither a regression:**

- `np.linalg.matrix_rank` runs on the unscaled `[1, Z]` design. I measured it: conditioners up to 1e12 are fine, but at >= 1e14 the rank collapses from 3 to 1 and the p-value moves from 0.392 to 0.211. Standardising columns before the rank check removes it. This is pre-existing in falsify and well outside realistic data scales, so a follow-up issue at most.
- `AGENTS.md`'s "Required Module Structure" list doesn't mention `_ci.py`, though that section explicitly permits free organisation of internal helpers.

**Things I checked and am happy with:** the `assert ... is not None` narrowing at `identify.py:591` matches existing convention (`_model.py` and `compile.py` do the same), and the characterisation-pins-then-refactor shape in `tests/test_ci_engine.py` is the right approach — including the `abs=0.0` detail on the ~1e-34 p-value pins.

drbenvincent (approved):

PR #354, opened 2026-06-26
Add instrumental-variable LATE example + regression test (#346)

daimon-pymclabs:
## Summary

Closes #346 (part of the IV epic #343). Adds a worked **instrumental-variable** example and a regression test, demonstrating that pathmc already estimates a continuous-outcome IV model with **zero API change** through the `~~` correlated-residual operator.

The example reproduces the **non-compliance / LATE** example from Matheus Facure's [*Causal Inference for the Brave and True*](https://matheusfacure.github.io/python-causality-handbook/09-Non-Compliance-and-LATE.html) and its [PyMC counterpart by Juan Orduz](https://juanitorduz.github.io/iv_pymc/): a randomized push-notification campaign where assignment is random but *delivery* is not, so engaged customers both receive more pushes and buy more.

## The IV spec

```python
spec = """
push_delivered  ~ push_assigned      # first stage (relevance)
in_app_purchase ~ push_delivered     # structural eq — instrument omitted (exclusion)
push_delivered ~~ in_app_purchase    # unobserved-confounder endogeneity
"""
model = pathmc.model(spec, data=df); model.fit()
```

The `~~` block compiles the two equations into a joint `MvNormal` with an `LKJCholeskyCov` residual covariance — the Bayesian limited-information IV structure. The residual correlation soaks up the unobserved engagement confounder; the structural coefficient on `push_delivered` recovers the LATE.

## What the example covers

- **What an IV is** — relevance, exclusion, exogeneity — and why those last two are untestable assumptions.
- **Why it's needed here** — non-compliance makes treatment-received endogenous; naive OLS is biased ~2× upward.
- **How to implement it in pathmc** — the spec mapping, `model.graph()` / `equations()`, the PyMC plate, recovery vs. the Wald estimator, the residual-correlation diagnostic, and a `do()` counterfactual.
- An honest note that `is_identifiable` is backdoor-only today (false negative for IV), pointing at #344.

On simulated data with a **known** true effect, the IV posterior brackets the truth and the frequentist Wald estimate while naive OLS sits far outside the interval.

## Regression test

`tests/test_instrumental_variable.py` — fixed-seed confounded system with a valid instrument; asserts naive OLS is biased, the IV posterior recovers the true effect, beats naive OLS, and yields a positive endogeneity correlation. Robust at the suite's fast-sampling cap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #344, opened 2026-06-23
IV-aware identification: instrument detection & validation


daimon-pymclabs:
Part of #343.

## Problem

`identify.py` reasons only about the backdoor criterion (`is_identifiable` delegates to `adjustment_sets`, which enumerates backdoor sets). When a treatment→outcome effect is confounded by an unobserved common cause (declared as a `~~` bidirected edge or a latent confounder) **but** a valid instrument exists, pathmc reports "not identifiable" — a false negative. It never inspects instruments.

## Proposal

Add graphical instrumental-variable identification over the declared DAG (including `~~`/latent edges). Standard conditions for `Z` to be a valid instrument for treatment `T` on outcome `Y`:
1. `Z` is associated with `T` (there is a directed path `Z → ... → T`);
2. every path from `Z` to `Y` is blocked once `T` is removed / intercepted (no direct `Z → Y`, no `Z`–`Y` backdoor) — i.e. `Z` affects `Y` only through `T`.

API options (pick during design):
- a dedicated `iv_identifiable(treatment, outcome)` returning a verdict + the instrument set; or
- extend `is_identifiable` to fall back to IV detection when no backdoor set exists, and surface the instrument(s).

Identification is **query-dependent** — it must be computed for a given `(treatment, outcome)` pair, which matches the existing identification API signature. DoWhy implements the analogous graph search and is a good reference.

## Acceptance
- Returns the candidate instrument set for a confounded `T→Y` with a valid `Z`.
- Correctly returns "not IV-identified" when the exclusion restriction is violated on the graph (e.g. a direct `Z → Y` edge or a `Z`–`Y` confounder).
- Unit tests over small DAGs covering valid / invalid-exclusion / no-instrument cases.

Blocks the advisory-warning issue and the binary-outcome issue.

drbenvincent:
#377 (shipped in 0.2.1) fixed a related but narrower bug: the backdoor criterion now sees `~~` residual-covariance blocks, so `is_identifiable()` no longer returns a false positive when unobserved confounding is declared via `~~`.

That is separate from this issue's scope (instrument detection and IV-aware identification). Keeping this open for the IV work tracked under #343.

PR #377, opened 2026-08-06
Make backdoor identification see `~~` residual blocks (#344)

daimon-pymclabs:
## The bug

`identify.py` reasons over the directed edges of the DAG and never reads `graph_info.residual_blocks`, so a `~~` declaration is invisible to the backdoor criterion. On the IV model

```
T ~ Z
Y ~ T
T ~~ Y
```

`adjustment_sets("T", "Y")` finds the empty set valid (the DAG it reasons over has no confounder at all), so `is_identifiable("T", "Y")` returns `True`. pathmc tells the user the effect is backdoor-identifiable in a model where they explicitly declared unobserved endogeneity, which reads as "naive OLS is fine here". That is the failure direction that matters: a false positive, not a false negative.

Surfaced while reviewing #354, whose identification section currently ships prose that contradicts its own frozen output because of this.

## The fix

Expand each residual block into a synthetic latent common cause before applying the criterion, which is the standard latent projection of an ADMG:

```python
def _with_residual_confounders(graph_info):
    ...  # one latent node per block, edge to every member
```

The node is latent, so it can never enter an adjustment set, and the backdoor path `T <- _u_resid_0 -> Y` it opens is unblockable. `adjustment_sets` returns `[]` and `is_identifiable` returns `False`, which is the correct answer.

Applied to `adjustment_sets` (hence `is_identifiable`) and `frontdoor_identifiable`. `_require_nodes` still validates against the user's own DAG so error messages do not leak synthetic node names.

## Deliberately out of scope

`implied_independences` / `test_implications` are left alone. Routing them through the same helper is also correct (the current code asserts independences that the declared confounding rules out) but it changes expected outputs and deserves its own PR. `collider_warnings` likewise. Happy to do both as follow-ups.

## Tests

Six cases in `TestResidualCovarianceIdentification`: the IV spec is not identifiable and has no adjustment set, the same DAG without `~~` still is, the synthetic node is never offered for adjustment, an unrelated `~~` block elsewhere in the graph does not break an unrelated effect, and the `PathModel` method agrees.

Related: #344, #354


drbenvincent (commented):
Reviewed with Bugbot and an independent defect-focused pass.

No actionable correctness or test-coverage findings. The residual-confounder projection and its scope are consistent with the PR description.

drbenvincent (changes requested):
Independent review: request changes. The residual-confounder projection correctly fixes backdoor/frontdoor checks, but the residual interpretation is inconsistent across the identification API: collider_warnings() and implied_independences()/test_implications() still inspect the unaugmented DAG and can miss opened collider paths or report independences contradicted by explicit ~~ covariance. Also generate collision-free synthetic latent names (a user variable named _u_resid_0 currently collides) and validate every residual endpoint during graph construction. CI is green.

daimon-pymclabs:
Addressed all three review points in 9180bce.

**1. `collider_warnings()` / `implied_independences()` (and `test_implications()`) now use the residual-augmented DAG**

Both functions read `graph_info.contemporaneous_dag` directly and never called the `_with_residual_confounders()` latent-projection helper that `adjustment_sets()`/`frontdoor_identifiable()` already used. Confirmed two concrete failure modes before fixing:

- `collider_warnings` (`pathmc/identify.py`, was line 343-344): for `T ~ Z; V ~ T; O ~ Z2; Y ~ O; V ~~ O`, `V` has a single declared parent (`T`), so the old code never even considered it as a collider candidate (`len(parents) >= 2` gate on the unaugmented graph fails). But the `~~` block makes `V` a genuine collider on the path `T -> V <- u -> O -> Y` once the synthetic confounder `u` is added. `collider_warnings(g, {"V"}, "T", "Y")` returned `[]` when it should warn.
- `implied_independences` (`pathmc/identify.py`, was line 528-550): for `A ~ Za; B ~ Zb; A ~~ B`, the old code asserted `A ⊥⊥ B | {Za, Zb}` even though `A ~~ B` explicitly declares an unobserved common cause — a textbook example of "reporting independences contradicted by explicit `~~` covariance."

Fix: both now call `_with_residual_confounders(graph_info)` to get the augmented DAG/latent set (`pathmc/identify.py:307-365` for `collider_warnings`, `:508-559` for `implied_independences`). For `implied_independences`, the observed-parent conditioning set is still computed from the *unaugmented* DAG (you can't condition on an unobservable synthetic node), but the d-separation check now runs on the augmented DAG, so a pair joined only via a `~~` path is correctly never reported as independent. `test_implications()` needed no change — it only consumes `implied_independences()`'s output, so it inherits the fix. Docstrings in `pathmc/identify.py` and `pathmc/_model.py` (`collider_warnings`, `implied_independences`) updated to describe this. Regression tests: `TestResidualCovarianceColliderConsistency` and `TestResidualCovarianceIndependenceConsistency` in `tests/test_identification.py`.

**2. Collision-free synthetic latent names**

Reproduced the exact scenario: a real confounder named `_u_resid_0` (`T ~ Z + _u_resid_0; Y ~ T + _u_resid_0`) plus an unrelated block `A ~~ B` elsewhere in the same graph. Because the unrelated block is index 0 in `residual_blocks`, the old code (`dag.add_node(f"_u_resid_{i}")`) merged straight into the real `_u_resid_0` node — giving it spurious edges to `A`/`B` and marking it latent, so `adjustment_sets(g, "T", "Y")` incorrectly returned `[]` (not identifiable) even though conditioning on the real `_u_resid_0` correctly identifies the effect.

Fix: `_fresh_latent_name()` (`pathmc/identify.py:97-107`) checks the generated name against the real DAG's node set and appends underscores until it's free, used from `_with_residual_confounders` (`pathmc/identify.py:82-93`). Regression test: `TestResidualNameCollision::test_user_variable_named_like_synthetic_latent_does_not_collide` in `tests/test_identification.py`.

**3. Validate every `~~` endpoint at graph construction**

Confirmed `build_graph()` previously accepted `T ~ Z\nY ~ T\nT ~~ Nonexistent` silently — `Nonexistent` was never checked against the DAG's node set, so it ended up as a phantom node that `_with_residual_confounders()` would later add edges to/from without any DAG context, and identification would reason about incorrectly rather than erroring.

Fix: `pathmc/graph.py:155-166` now diffs the union of all residual-block variables against `set(dag.nodes)` right after `_build_residual_blocks(spec)` and raises `ValueError` listing the unknown variable(s) and the available ones. This runs before the pre-existing latent/`~~` conflict check, and before `compile.py`'s own residual-block handling, so it fails fast and clearly. Regression test: `TestResidualBlocks::test_unknown_residual_endpoint_raises` (parametrized for both `~~` operand positions) in `tests/test_graph.py`.

**Testing**

`uv run pytest tests/test_identification.py tests/test_dag.py tests/test_graph.py tests/test_parse.py -q` → 143 passed. `uv run prek run --all-files` → all hooks pass clean (ruff, ruff-format, mypy, license header).

drbenvincent:
## Re-review (tip `a839780`)

Re-reviewed after the author's response commits and the HSGP test fix.

### Prior CHANGES_REQUESTED items — all addressed

| Finding | Status |
|---------|--------|
| `collider_warnings()` / `implied_independences()` use augmented DAG | ✅ Both call `_with_residual_confounders()` |
| Collision-free synthetic latent names | ✅ `_fresh_latent_name()` with underscore suffix loop |
| Validate `~~` endpoints at graph construction | ✅ `build_graph()` raises on unknown endpoints |

**Additional fix:** `a839780` updates `test_hsgp_compile.py` after endpoint validation correctly rejects invalid `~~` members in HSGP residual blocks.

**Local verification:** `tests/test_identification.py` + `tests/test_graph.py` + `tests/test_hsgp_compile.py` → **72 passed**. `prek` CI green.

**Recommendation:** **Approve** — review requests fully satisfied.

drbenvincent (approved):
All three review threads addressed: consistent residual-augmented DAG across identification API, collision-free latent names, and early ~~ endpoint validation. Tests pass. Approved.

drbenvincent:
## Triage summary (a839780)

CI is green after two follow-up fixes on top of 9180bce.

### CI failures fixed

1. **`test_dangling_residual_endpoint_rejected`** — the new `build_graph()` validation raised `ValueError: Residual covariance (~~)…` (capital R), which did not match the case-sensitive `match="residual covariance"` regex in `test_falsify.py`. Fixed in `aa0ece4` by rewording the message to `Unknown variable(s) in residual covariance (~~): …`.

2. **`test_hsgp_in_residual_block_raises`** — the spec `y ~ hsgp(x)\ny ~~ z` used `z` only in `~~` without a regression equation, so graph construction now correctly rejects it before compile. The test intent is compile-time HSGP+`~~` rejection; fixed in `a839780` by adding `z ~ x` so `z` is a declared DAG node.

### Review re-check (9180bce + fixes)

All three requested changes from @drbenvincent are addressed:

| Request | Status |
|---------|--------|
| Route `collider_warnings()` / `implied_independences()` through `_with_residual_confounders()` | Done — regression tests in `TestResidualCovarianceColliderConsistency` and `TestResidualCovarianceIndependenceConsistency` |
| Collision-free synthetic latent names via `_fresh_latent_name()` | Done — `TestResidualNameCollision` |
| Validate `~~` endpoints at graph construction | Done — `test_unknown_residual_endpoint_raises` in `test_graph.py` |

### CI

- `prek` ✅
- `test` ✅ (1176 tests, run 31172845199)
- Docs / wheel / zizmor ✅

Approving — still draft per author preference; not marking ready or merging.

drbenvincent (approved):
All three review points are addressed in 9180bce, CI failures from the new `~~` endpoint validation are fixed in aa0ece4 and a839780, and the full test suite is green. Residual-augmented DAG is now used consistently across `adjustment_sets`, `frontdoor_identifiable`, `collider_warnings`, and `implied_independences`.

PR #378, opened 2026-08-06
Add worked example for refute_placebo() (#340)

### LINKED ISSUE #340: Docs: Add worked example for refute_placebo() (#310)
## Summary

PR #310 adds `PathModel.refute_placebo()` — a Bayesian upgrade of DoWhy's placebo treatment refuter — but includes **zero demonstrations** in the example notebooks.

## What's needed

Add one or more worked examples demonstrating the placebo refutation workflow. Options:

### Option 1: Focused standalone example (RECOMMENDED)
Create `docs/examples/03-identification/placebo_refutation.qmd` showing:
- Complete workflow from model fitting through refutation
- Interpretation of both verdicts (`passes_placebo`, `effect_survives`)
- Both plot kinds (`kind="comparison"` vs `kind="null"`)
- When to use 4 vs 8+ permutations (cost/precision tradeoff)
- A case where the pipeline fails the placebo test (misspecified model or data issue)

### Option 2: Integrate into existing examples
Add refutation checks to one or more of:
- `docs/examples/03-identification/causal_identification.qmd` (already covers backdoor, front-door, collider warnings)
- `docs/examples/03-identification/sensitivity.qmd` (parallel robustness check)
- `docs/examples/02-effect-estimation/ate_estimation.qmd` or `mediation.qmd`

### Option 3: Both
Standalone example for depth + brief refutation checks in 1-2 existing examples showing it as part of a standard workflow.

## Context

- The user guide (`docs/user_guide/13-causal-inference.qmd`) has clear documentation with code snippets (lines 284-310)
- `refute.py` has comprehensive docstrings
- The feature is production-ready and well-tested (`tests/test_refute.py`)
- This is critical for user adoption — most users learn from examples, not API docs

## Related

- PR #310 (merged once API reference is fixed)
- Issue #253 (parent issue for refutation suite; placebo refuter is first of four)

cetagostini:
## Summary

Closes #340.

`refute_placebo()` (#310) shipped with no demonstrations in the example notebooks. This adds the standalone worked example from Option 1 of the issue: `docs/examples/03-identification/placebo_refutation.qmd`, covering the full workflow, both verdicts, both plot kinds, the permutation-count tradeoff, and a pipeline that fails the test.

## Changes

- **`docs/examples/03-identification/placebo_refutation.qmd` (new)** — executable walkthrough on a confounded `X → Y` DAG with an observed confounder `Z` (true ATE 0.5, recovered 0.464):
  - **Workflow** — `model()` → `fit()` → `ate()` → `refute_placebo()`, with `sample_kwargs` forwarded to the per-permutation re-fits and `random_seed` pinned throughout.
  - **Both verdicts, kept apart** — `passes_placebo` judges the *machinery* (does `mu_null`'s HDI contain zero?); `effect_survives` judges the *effect* (is `p_tail` below `significance_level` once the real ATE is calibrated against the null predictive distribution?). The example includes a case where they disagree, which is the reason they are two attributes and not one.
  - **Both plot kinds** — `kind="comparison"` framed as the one to report, `kind="null"` as the one to inspect when a verdict surprises you.
  - **4 vs 8+ folds** — a live table rather than an assertion: HDI width `0.1021 → 0.0529`, `tau_het 0.0352 → 0.0202`. The asymmetry is spelled out — a FAIL is trustworthy at four folds, a PASS is not, since `tau_het` is a variance component estimated from a handful of points and its prior-inflated interval straddles zero almost automatically.
  - **`z_cal` denominator** — a callout on why `sigma_pred` convolves the null predictive spread with the mean *within*-fold variance instead of treating each placebo re-fit as error-free.
  - **Cross-reference table** — where the placebo test sits relative to `falsify()` and `sensitivity()`, and why all three are needed.
- **`docs/examples/thumbnails/placebo_refutation.png` (new)** — the `kind="comparison"` figure at 1400×700, matching the existing thumbnails.
- **`great-docs.yml`** — registers the `Refutation` tag icon (`shield-check`), alongside the existing `Sensitivity Analysis` entry.
- **`_freeze/examples/identification/placebo_refutation/`** — freeze cache (pages are `freeze: true`, so CI renders from this and never spawns a kernel).

## On the failure case

The issue asks for "a case where the pipeline fails the placebo test". Finding one is not a matter of picking bad data, and the reason is worth recording.

Permuting the treatment destroys its association with *everything*, so any coefficient estimated from that association collapses toward zero, and so does the `do()` contrast built on it. A correctly specified linear/GLM path model therefore passes this test essentially by construction — the placebo ATE is centered on zero with random sign. What survives permutation is any mechanism that holds the estimate off zero regardless of the data, and the practical one is a **coefficient that cannot reach zero**.

The example uses `priors={"beta_Y": Prior("HalfNormal", sigma=10)}` — the everyday MMM "spend cannot hurt sales" sign constraint. It is defensible as domain knowledge but it is not free: under permutation the posterior piles up just above zero and the pipeline reports an effect for a treatment it should be blind to. Shown at two severities, because the verdict alone does not tell you whether to care:

| Scenario | folds | `passes_placebo` | `effect_survives` | `mu_null` |
|---|---|---|---|---|
| Sound pipeline | 4 | PASS | survives | 0.0098 `[-0.0434, 0.0587]` |
| Non-negative prior, strong effect | 8 | **FAIL** | survives | 0.0242 `[0.0106, 0.0376]` |
| Non-negative prior, weak effect | 8 | **FAIL** | **not distinguishable** | 0.0908 `[0.0400, 0.1366]` |

All eight fold means are positive in both failure runs, against scattered signs when the pipeline is sound — the sign pattern is the tell, and the example says so.

The middle row exercises the ROPE caveat in the `passes_placebo` docstring: the bias is real and precisely estimated, so the interval rule fails it, but it is 5.2% of the observed ATE and the substantive conclusion stands. The bottom row is the cautionary one: the placebo bias (0.0908) *exceeds* the observed ATE (0.0674), `z_cal` comes out negative because the real estimate sits below the null's center, and `p_tail` is 0.384 — the apparent finding is the prior talking. A closing callout ranks the four causes of a failed placebo test (over-constrained priors, bounded outcomes, endogenous treatments, non-converged folds).

## Cross-link path deviation

The links to `dag_falsification.qmd` and `sensitivity.qmd` use `/examples/identification/...`, **not** the `/examples/03-identification/...` form used in the other 22 cross-links in the repo. The build copies sections to `great-docs/examples/identification/` with the numeric prefix stripped, so the numbered form does not resolve — it emits a dead `href=".../examples/03-identification/sensitivity.qmd"` into the rendered page and a `WARN: Unable to resolve link target` at build time.

Verified: with the numbered form the build reports 24 such warnings, with the stripped form 22, and the table links become `../../examples/identification/sensitivity.html`. The 22 remaining are pre-existing across `docs/user_guide/*.qmd` and four example pages; fixing those is mechanical but out of scope here and wants its own issue.

## Verification

- `great-docs freeze docs/examples/03-identification/placebo_refutation.qmd` — every cell executes clean; both figures produced.
- Every numeric and directional claim in the prose was checked against the frozen output rather than written from expectation, including the three verdict pairs above, the HDI-width halving, `all folds positive`, the 5.2% bias ratio, and the negative `z_cal`.
- `great-docs build` — Build complete, 22 warnings (all pre-existing, see above). Page renders with its thumbnail on the examples index; the tag manifest registers `["Causal Inference", "Refutation"]` and the `shield-check` icon resolves.
- `make lint` — all hooks pass. Note that `ruff-format` formats Python cells inside `.qmd`; the page was reformatted and re-frozen, and the frozen results are byte-identical since all seeds are pinned.
- `uv run pytest tests/test_refute.py` — 34 passed, 10 deselected.

## Out of scope

- Option 2 from the issue (brief refutation checks folded into `causal_identification.qmd`, `sensitivity.qmd`, `ate_estimation.qmd`, `mediation.qmd`). This PR is Option 1 only; happy to follow up if the standalone page reads as sufficient.
- The 22 pre-existing broken example cross-links described above.
- The remaining three refuters in the #253 suite.


drbenvincent (commented):
## Overall

Strong addition — this is one of the better example notebooks in the identification section. The page does real pedagogical work: it explains *why* the test exists before the API, separates the two verdicts with plain-language questions, and turns the sign-constrained prior failure into a genuine teaching moment rather than a contrived demo. The 4-vs-8-fold asymmetry (FAIL is trustworthy early, PASS is not) is exactly the kind of honest limitation note that builds trust.

I read this against the educational-narrative and figure-excellence guidance, and checked the statistical story against `pathmc/refute.py` and doWhy's placebo refuter.

---

## Pedagogy (educational-narrative)

**What works well**

- **Examples-first hook** (opening paragraphs) — starts with the puzzle ("does the pipeline manufacture effects?") before notation.
- **Progressive complexity** — setup → run → interpret verdicts → visualize → permutation tradeoff → manufactured failure → suite context → summary. Each section earns the next.
- **Callouts are purposeful** — `z_cal` denominator, ROPE caveat, failure-mode checklist, and reflection prompts all map to specific pedagogical jobs (not decoration).
- **"Right for the wrong reasons" handled well** — the strong-effect + HalfNormal case (`passes_placebo` fails but `effect_survives` holds) is a nice calibration of *when to care*.
- **Reality check** — ground truth (0.5) stated upfront and recovered before refutation; weak-effect case shows the prior eating the signal.
- **Reflection prompts** — domain-oriented and concrete; matches the pattern in `sensitivity.qmd`.

**Suggestions**

1. **Prerequisites paragraph** — `dag_falsification.qmd` opens with an explicit "assumes you have read …" link. This page jumps straight into the hook. A short prerequisite note (e.g. [ATE estimation](/examples/effect-estimation/ate_estimation.qmd) and ideally the backdoor/confounding setup from [causal identification](/examples/identification/causal_identification.qmd)) would help non-experts who land here from the tag index.
2. **One plain-language sentence on hierarchical pooling** — paragraph 2 introduces $\mu_\text{null}$ and $\tau_\text{het}$ quickly. A single clause like "think of it as meta-analysing the placebo re-fits" would lower the bar before the symbols appear.
3. **Explicit "reality check" beat** — after the first `ate` output, one sentence calling out that 0.464 vs true 0.5 validates the DGP/workflow before we stress-test the pipeline would mirror the educational-narrative pattern more explicitly (minor — the numbers are there).

---

## Figures (figure-excellence)

- `FIG_WIDTH` / `FIG_HEIGHT` constants, no width multipliers, `code-fold: true` on plotting cells, technical `fig-cap`s with labels — all good.
- Captions describe *what* is shown, not narrative — good.
- Colors are library defaults (`tab:red` / `tab:gray`) rather than named semantic constants in the notebook setup; fine for a two-row comparison plot, but if this page grows more figures, consider aligning with the entity-color pattern used elsewhere.

No blocking figure issues.

---

## Is the Bayesian doWhy upgrade legit?

**Yes, with the caveats the page already documents.**

The core placebo logic matches doWhy: permute the treatment, sever the causal link, re-estimate, expect ~zero. That is standard refutation practice (see doWhy's placebo treatment refuter and the Sharma et al. DoWhy paper).

The pathmc upgrade is statistically coherent:

1. **Per-permutation posteriors $(m_j, s_j)$** instead of point estimates — appropriate for a Bayesian pipeline; collapsing to a single number would throw away within-fit uncertainty.
2. **Hierarchical normal–normal pooling** — standard random-effects / meta-analysis structure; $\mu_\text{null}$ is the systematic placebo bias, $\tau_\text{het}$ is between-permutation heterogeneity. The Roever-style weakly informative scale on $\tau_\text{het}$ at small $J$ is defensible (and explains why 4 folds is screening-only).
3. **Two verdicts** — cleaner than doWhy's single significance check:
   - `passes_placebo`: machinery honest? ($\mu_\text{null}$ HDI contains 0)
   - `effect_survives`: real effect larger than placebo noise? (calibrated `p_tail`)
4. **`sigma_pred` denominator** — including mean within-fold variance is the right choice; treating each re-fit as error-free would be anti-conservative.

The manufactured failure (HalfNormal on `beta_Y`) is not a trick — it is the most realistic way to fail this test in practice, and the page explains why correctly specified models pass *by construction*.

**Residual caveats worth keeping visible** (mostly already in prose or callouts):

- Gaussian likelihood on fold *summaries* is an approximation for bounded/non-Gaussian estimands.
- `p_tail` is directionally post hoc (sign chosen from the observed effect), not a preregistered one-sided test — worth one explicit sentence for readers migrating from frequentist refutation.
- Calibration asymmetry (full predictive null vs point-mean observed ATE) is documented in the API; the example could nod to it when introducing `effect_survives`.

None of these undermine legitimacy; they bound interpretation.

---

## References the page could cite

The repo already has `docs/references.bib` wired through `great-docs.yml`. `dag_falsification.qmd` cites `[@eulig2023falsifying]` inline — this page would benefit from the same treatment:

| Citation | Why |
|---|---|
| **Sharma et al. (2020)** — *DoWhy: An End-to-End Library for Causal Inference* ([arXiv:2011.04216](https://arxiv.org/abs/2011.04216)) | Canonical source for the refutation step and placebo treatment refuter that this ports. |
| **Bugaev & Trujillo** — *Bayesian Design Analysis for Time-Series Quasi-Experiments: A Placebo-Calibrated Framework* | The hierarchical calibration model is attributed to this in `refute.py`; cite it where $\mu_\text{null}$ / $\tau_\text{het}$ are introduced. |
| **Röver et al. (2021)** — *On weakly informative prior distributions for the heterogeneity parameter…* ([doi:10.1002/jrsm.1475](https://doi.org/10.1002/jrsm.1475)) | Justifies the $\tau_\text{het}$ prior behaviour at small $J$; supports the "8+ folds to certify" guidance. |
| **doWhy placebo docs** — [Placebo Treatment Refuter](https://www.pywhy.org/dowhy/main/user_guide/refuting_causal_estimates/refuting_effect_estimates/placebo_treatment.html) | Useful as an inline link in the doWhy vocabulary paragraph, even without a bib entry. |

`[@pearl2016primer]` is optional background on refutation as robustness checking.

---

## Accessibility for non-experts

Broadly **yes** — a motivated analyst without deep causal-inference training can follow the workflow and the two verdicts. The prose is clear and the doWhy bridge helps.

The main friction points for a non-expert are:

- Hierarchical pooling and `z_cal` (the callout helps; a one-line intuition upfront would help more).
- Distinguishing `passes_placebo` vs `effect_survives` — well handled, but readers may skim past it; the summary bullets reinforce it nicely.

---

## Verdict

Approve with minor suggestions. The page closes a real documentation gap (#340), the failure cases are honest and instructive, and the statistical story holds up. Adding prerequisites + 2–3 citations would bring it in line with the best sibling pages (`dag_falsification.qmd`, `sensitivity.qmd`) without changing the workflow.

drbenvincent (approved):
Approve — strong pedagogical example with a statistically sound Bayesian upgrade of doWhy's placebo refuter. See the detailed review comment for citations and minor suggestions.

cetagostini:
Thanks — this was a genuinely useful review. All the actionable suggestions are applied in 8fbce77; details and the one thing I skipped below.

## Pedagogy

**1. Prerequisites paragraph** — added as paragraph 3, mirroring `dag_falsification.qmd`'s opener, pointing at [ATE estimation](/examples/effect-estimation/ate_estimation.qmd) for `ate()`/g-computation and [Causal Identification](/examples/identification/causal_identification.qmd) for the backdoor setup, with the reason stated: you need to be comfortable with what an ATE *is* before asking whether yours is real.

**2. Plain-language pooling gloss** — the hierarchical model is now introduced as "think of it as **meta-analysing the placebo re-fits**, treating each shuffle as one study", before $\mu_\text{null}$ and $\tau_\text{het}$ appear. That framing also does double duty for the Röver citation below.

**3. Explicit reality-check beat** — added after the first `ate()` output: we simulated the data, so 0.464 against a known 0.5 with a CI excluding zero establishes that the workflow recovers what it should on data we control, which is the precondition for trusting it anywhere else.

## Residual caveats

Both of the ones you wanted surfaced are now a `callout-important` titled **"Two things `p_tail` is not"**, placed immediately where `effect_survives` is introduced:

- **Not a preregistered one-sided test** — the tail side is chosen post hoc from the sign of the observed effect relative to the null center, framed explicitly for readers arriving from frequentist refutation.
- **A distribution compared against a point** — the null enters as its full predictive distribution while the effect enters as its posterior mean, so `observed_ate_hdi` should be read next to the verdict rather than treating `effect_survives` as a claim about the whole posterior.

The Gaussian-likelihood-on-bounded-estimands caveat was already item 2 of the "What to do with a failed placebo test" callout, so I left it where it was rather than duplicating it.

## Citations

Three new entries in `docs/references.bib`, all cited inline where the relevant claim is made:

- **`@sharma2020dowhy`** — on the sentence identifying this as the Bayesian upgrade of the ported refuter, plus an inline link to doWhy's own [placebo treatment refuter](https://www.pywhy.org/dowhy/main/user_guide/refuting_causal_estimates/refuting_effect_estimates/placebo_treatment.html) page in the vocabulary paragraph, so the point-estimate version of that table is one click away.
- **`@bugaev_placebo_calibration`** — where the calibration model is introduced, matching the attribution in `pathmc/refute.py`. Filed as `@unpublished`, because I could not find a public URL or DOI for it — **if there is a preprint or a year, tell me and I will complete the entry** (it currently renders as "n.d.").
- **`@rover2021weakly`** — on the "four folds screen, eight certify" guidance, tying it to the same small-study problem in Bayesian meta-analysis: with few studies the heterogeneity parameter is prior-driven rather than data-driven. Verified against the published record (Research Synthesis Methods 12(4), 448–474).

I left `@pearl2016primer` out since you flagged it as optional and the page already has three citations doing specific work; happy to add it if you'd rather the refutation-as-robustness framing were sourced.

## Not applied

The figure note — semantic color constants instead of `tab:red`/`tab:gray`. You called it non-blocking and conditional on the page growing more figures, and it has two; I'd rather not introduce a one-off color convention here that diverges from the sibling pages. Worth doing repo-wide if we want it.

## Verification

Re-frozen and rebuilt: all three citations resolve into the rendered reference list with no unresolved markers, both prerequisite links resolve to real pages, `make lint` passes, and the frozen numeric results are unchanged (this commit touches prose and the bibliography only — no code cells, and every seed was already pinned).

One incidental datapoint: you wrote the prerequisite links in the stripped `/examples/effect-estimation/...` form, which is the form that actually resolves — the numbered `/examples/03-identification/...` form used in the other 22 cross-links emits a dead `.qmd` href. Independent confirmation of the deviation flagged in the PR description, and I think it justifies the follow-up issue.

PR #379, opened 2026-08-06
Fix TeX double-subscript errors in generated model equations

juanitorduz:
## Problem

The published [hsgp_1d](https://pathmc.pymc-labs.com/examples/foundations/hsgp_1d.html) example page shows the `model.equations()` / `model.priors()` outputs as raw red LaTeX source instead of rendered math. The same breakage is baked into the frozen outputs of the [mmm](https://pathmc.pymc-labs.com/examples/applied-models/mmm.html) and [dynamic_pricing](https://pathmc.pymc-labs.com/examples/applied-models/dynamic_pricing.html) pages.

**Root cause:** user-supplied names reach TeX unescaped. `_latexify_name()` inserted multi-part name suffixes verbatim into the subscript, e.g. `beta_hsgp_y_x` → `\beta_{hsgp_y_x}`. That is a TeX **double-subscript error**, so KaTeX (`throwOnError: false`) aborts the whole `\begin{aligned}` block and renders the raw source in red (`ParseError: KaTeX parse error: Double subscript`, reproduced in a headless-browser control).

`_latexify_name` was not the only exposed path. Response and predictor names are interpolated into TeX at a dozen other sites that never call it: the `\beta_0` intercept, the four likelihood subscripts, `mu_lhs`, and the `\mathrm{}` / `\operatorname{}` groups in `_format_term_latex` and `_format_transform_latex`. `_latexify_prior` has the same exposure in its arguments, which sit *outside* their `\text{}` group. So a spec like `birth_weight ~ tv_spend + sales_q_1` still emitted `\mathrm{sales_q_1}` — the same hard KaTeX error — and `Normal(mu_alpha, sigma_alpha)` shipped bare underscores at top level.

## Fix

Escaping is applied at every name-to-TeX boundary, via two helpers that mirror the two contexts a name can appear in:

- `_latex_index()` for subscript groups: underscore-separated tokens become comma-separated indices, standard multi-index notation — `\beta_{hsgp,y,x}`, `\eta_{y,x}`, `\mu_{slope,sales,tv}`. Single-token names are unchanged (`\theta_{tv}`).
- `_latex_escape()` / `_latex_symbol()` for names shown as symbols: `_` becomes `\_`, so `\mathrm{tv\_spend}`.

Every raw interpolation site is routed through one of them, including `_latexify_prior`.

`ell` is now recognized as `\ell`, so the HSGP lengthscale renders as `\ell_{y,x}`.

## Tests

`TestLatexifyName` and `TestLatexRendering` assert a `_BARE_UNDERSCORE` invariant: in generated LaTeX, every `_` either is escaped as `\_` or opens a subscript group `_{`. This is stricter than the `_DOUBLE_SUBSCRIPT` check and, unlike it, also inspects `\mathrm{}` and `\text{}` groups, where the same class of breakage lands. Parametrized over all five likelihood families with an underscored response name, plus a transform case and three prior strings.

Note the distinction the invariant draws. `\mu_{birth_weight}` has one underscore in the group, so KaTeX *parses* it rather than erroring — but it typesets as a nested subscript, µ with subscript "birt h_w eight". Only a name with a token between two underscores (`y_t_1`, `sales_q_1`, plausible for lagged or indexed data) produces the red error block. Both are wrong; the tests now catch both.

## Freeze cache

The fix changes what `equations()` and `priors()` emit, so 18 committed freeze artifacts (9 pages, each duplicated across the numbered and prefix-stripped layouts) disagreed with the code that produced them: counterfactual, mediation, causal_identification, did, panel_data, dynamic_pricing, mmm, mmm_awareness, saas_funnel.

These were **not** regenerated by re-running the notebooks. Refreezing three pages in the first commit of this PR moved 1343 insertions and 1266 deletions across 36 files; the TeX correction inside was about four lines each, and the rest was MCMC nondeterminism (pymc graph SVGs relaid out by ~300 lines, figure PNGs rewritten byte-for-byte). Scaling that to 18 files would bury a few dozen escaped underscores in an unreviewable diff, and would let sampler or dependency drift silently rewrite figures under cover of the noise.

`scripts/refreeze_tex.py` instead rewrites the cached TeX in place. The freeze `hash` field is derived from the source document, not the result payload, so the cache stays valid. The artifact is spliced rather than re-serialised, since a `json` round-trip does not reproduce Quarto's output byte for byte. Result: **28 blocks across 18 files, 18 insertions and 18 deletions.**

A textual rewrite is a second implementation of the escaper and free to drift from it. `scripts/verify_refreeze_tex.py` closes that empirically: emit LaTeX for 50 spec/family combinations at the pre-fix revision, emit again with the fix, and require `transform(old) == patched` exactly.

## Freeze-cache path fix

`hsgp_1d`'s freeze cache existed only under the old numbered layout (`_freeze/examples/01-foundations/`), while the CI build (great-docs 0.16.0) consumes the prefix-stripped layout (`_freeze/examples/foundations/`) — so CI re-executed that page on every docs build. The refreshed cache is now committed under both layouts, matching every other example page.

## Verification

- Full test suite: **1190 passed**, 191 deselected (`-m "not slow"`). New tests fail on the pre-fix code and pass on the fix.
- `ruff format --check`, `ruff check`, and `mypy --ignore-missing-imports pathmc/` all clean.
- Equivalence proof for the freeze rewrite: **50 cases checked, 0 skipped, all pass**, no output violating the invariant.
- Freeze rewrite is idempotent and does not double-escape the `\mathrm{treat\_post}` that `did` already had correct.
- Invariant sweep over the whole cache: **all 89 `aligned` blocks in `_freeze/` are clean.** (The first commit's claim of "no double-subscript TeX remains anywhere" held only for hard KaTeX errors; nine sites still rendered as nested subscripts. This is the stricter statement.)
- `great-docs build` restores all three originally-affected pages from the committed cache (frozen sampler values appear verbatim in the built HTML — no re-execution).
- Headless-Chrome render of the built site: all math blocks typeset by KaTeX, **0** `katex-error` nodes; a control page with the old TeX reproduces the exact `Double subscript` red-error block.


scripts/verify_refreeze_tex.py
```diff
@@ -0,0 +1,140 @@
...
+#   limitations under the License.
+"""Check that the freeze rewrite agrees with the real escaper.
+
+``scripts/refreeze_tex.py`` transforms cached TeX textually, which makes it a
+second implementation of the escaping in ``pathmc.introspect`` and therefore
+free to drift from it. This closes that gap empirically: emit the LaTeX for a
+corpus of specs at the unpatched revision, emit it again with the patch
+applied, and require that transforming the former reproduces the latter
+exactly. If that holds, rewriting the freeze artifacts is equivalent to
+re-executing the notebooks as far as the equations are concerned.
```
drbenvincent, line 22: Nice equivalence check. If we expect future cache TeX patches, consider a `make verify-freeze-tex` target (or a lightweight CI job) so this doesn't become dead tooling after merge. Not required for this PR.

scripts/refreeze_tex.py
```diff
@@ -0,0 +1,185 @@
...
+
+* a subscript group ``_{...}`` holds indices, so a bare ``_`` becomes ``,``
+  (:func:`pathmc.introspect._latex_index`);
+* an upright-text group ``\\mathrm{}`` / ``\\operatorname{}`` / ``\\text{}``
+  shows a name as written, so a bare ``_`` becomes ``\\_``
+  (:func:`pathmc.introspect._latex_escape`).
+
+Groups nest (``f_{\\mathrm{hsgp}}``), so the walk is recursive and the
+innermost group wins. ``scripts/verify_refreeze_tex.py`` checks the result
+against the real escaper rather than trusting this description.
```
drbenvincent, line 41: The recursive group walk and idempotency guard for existing `\_` are the subtle bits — good that `verify_refreeze_tex.py` proves equivalence rather than relying on the docstring alone.

pathmc/introspect.py
```diff
@@ -60,22 +61,58 @@
 )
 
 
+def _latex_index(name: str) -> str:
```
drbenvincent, line 64: Clean factoring. One pre-existing edge (out of scope here): `_latexify_expression` still routes tokens through `_latexify_name` (subscript-index style) rather than `_latex_symbol`, so a defined param referencing `tv_spend` renders as `tv_{spend}` while the predictor in the same equation shows `\mathrm{tv\_spend}`. Both parse in KaTeX; just a cosmetic inconsistency if we ever want to unify.

tests/test_introspection.py
```diff
@@ -17,12 +17,43 @@
...
 from conftest import MEDIATION_SPEC, PARALLEL_MEDIATORS_SPEC
 
+# Two bare underscores inside one subscript group (e.g. \beta_{hsgp_y_x}) are
+# a TeX double-subscript error: MathJax aborts the whole aligned block and
+# dumps raw source. Generated names should carry no bare underscore at all —
+# suffix tokens become comma-separated indices — so assert the stricter form.
+_DOUBLE_SUBSCRIPT = re.compile(r"_\{[^{}]*_[^{}]*\}")
+
+# The invariant that actually covers every emission site: every underscore in
+# generated LaTeX is either escaped (``\_``) or opens a subscript group
```
drbenvincent, line 45: The `_BARE_UNDERSCORE` regex + comment explaining why it's stricter than `_DOUBLE_SUBSCRIPT` is exactly the kind of invariant test I want for this class of bug. Good coverage.

drbenvincent (approved):
## Review summary

Solid fix for a real docs bug. The root-cause analysis is correct: bare underscores in subscript groups and upright-text groups break KaTeX, and the PR routes every name-to-TeX boundary through the right helper.

**What looks good**
- `_latex_index` / `_latex_escape` / `_latex_symbol` split the two TeX contexts cleanly and are applied consistently (intercept, likelihood lines, transforms, priors).
- The `_BARE_UNDERSCORE` invariant is stricter and more useful than `_DOUBLE_SUBSCRIPT` alone; parametrizing over all five likelihood families is the right coverage.
- The freeze-cache strategy is pragmatic: `refreeze_tex.py` keeps the diff reviewable, and `verify_refreeze_tex.py` closes the "second implementation" drift risk empirically. I re-ran the introspection tests locally (40 passed) and swept `_freeze/` aligned blocks — 0 invariant violations.
- Adding `hsgp_1d` under the prefix-stripped `_freeze/examples/foundations/` layout fixes the CI re-execution issue.

**Nits (non-blocking)**
- Commit `560b8ae` still leaves ~28 figure SVG/PNG files with byte-level churn from MCMC nondeterminism (mmm, dynamic_pricing). The later textual-rewrite commit avoids that for the other pages; consider reverting the figure-only diffs from the first commit while keeping the TeX fixes and the new `foundations/hsgp_1d` cache. Would make this much easier to review and bisect.
- The verify/refreeze scripts are well-documented but not wired into `Makefile` or CI. Fine as one-time migration tooling; a short note in CONTRIBUTING (when to use `refreeze_tex.py` vs `make refreeze-docs`) would help the next person who hits a similar cache mismatch.

CI is green. Approving.

PR #381, opened 2026-08-06
Ensure pathmc Jupyter kernel is registered before docs builds

drbenvincent:
## Summary
- Add a `jupyter-kernel` Make target that registers the `pathmc` Jupyter kernel from the current `.venv`
- Wire it into `setup`, `make docs`, and `make refreeze-docs` so kernelspec paths stay in sync with the active checkout

This prevents Quarto from failing with a stale Python path (e.g. after working in a Cursor worktree) when a page needs live kernel execution instead of the `_freeze/` cache.

## Test plan
- [x] `make jupyter-kernel` registers kernel pointing at `.venv/bin/python3`
- [ ] `make docs` succeeds without manual kernel setup after switching worktrees

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,17 +28,11 @@ jobs:
         # `make setup`), so CI resolves the exact versions used in local dev.
         run: uv sync --all-extras
 
-      - name: Register `pathmc` Jupyter kernel
-        # `.qmd` files (and great-docs.yml) declare `jupyter: pathmc`, so the
-        # CI runner needs a kernelspec under that name pointing at the same
-        # Python interpreter the package was installed into above.
-        run: uv run python -m ipykernel install --user --name pathmc
-
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
 
       - name: Build docs
-        run: uv run great-docs build
+        run: make docs
 
       - name: Verify agent-context files were generated
         # great-docs builds llms.txt / llms-full.txt / skill.md at build step 3,
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,10 +55,11 @@ This package targets **PyMC ≥ 6.0, ArviZ ≥ 1.1, PyTensor ≥ 3.0, and NumPy 
 
 The site is built with [Great Docs](https://posit-dev.github.io/great-docs/) via `great-docs.yml`: `uv run great-docs build`. The output dir `great-docs/_site/` is ephemeral and gitignored — never edit it. Edit sources instead: `docs/user_guide/*.qmd`, `docs/examples/**/*.qmd`, `great-docs.yml`, and `pathmc/skills/pathmc/SKILL.md`.
 
-Notebooks are frozen (`freeze: true`): the committed `_freeze/` cache supplies cell outputs and CI never spawns a kernel, so **local previews show the last-frozen output, not in-progress edits.** After editing an executable page, refresh and commit the cache (local freezing needs the kernel once: `uv run python -m ipykernel install --user --name pathmc`):
+Notebooks are frozen (`freeze: true`): the committed `_freeze/` cache supplies cell outputs and CI never spawns a kernel, so **local previews show the last-frozen output, not in-progress edits.** After editing an executable page, refresh and commit the cache (local freezing needs the kernel once via `make jupyter-kernel`):
 
 ```bash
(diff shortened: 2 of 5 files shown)


daimon-pymclabs (commented):
Reviewed at `chore/ensure-jupyter-kernel`. The change is correct and solves a real annoyance — the stale `argv[0]` path in `~/.local/share/jupyter/kernels/pathmc/kernel.json` after switching worktrees is exactly the failure mode, and re-running the install is the fix. `ipykernel` is in the `dev` dependency-group, which `uv run` syncs by default, so `make docs` works from a clean checkout without `--all-extras`. Non-blocking notes below.

**1. The install command now exists in four places.** `Makefile` (new), `.github/workflows/docs.yml:36`, `CONTRIBUTING.md:116`, and the comment at `great-docs.yml:176`. This PR adds the fourth copy rather than consolidating. Since the Makefile target is now the canonical entry point, worth folding the others into it in the same PR:

```yaml
# .github/workflows/docs.yml
- name: Register `pathmc` Jupyter kernel
  run: make jupyter-kernel
```

and pointing `CONTRIBUTING.md:113-116` at `make jupyter-kernel` (also dropping "once", which is now inaccurate — `make setup` and `make docs` handle it).

**2. Consider `--sys-prefix` instead of `--user`.** `--user` writes a machine-global kernelspec, so with two pathmc checkouts on one machine, `make docs` in checkout A silently repoints checkout B's kernel — the same class of bug this PR fixes, just moved from "stale after worktree switch" to "clobbered by the other checkout". `--sys-prefix` installs into `.venv/share/jupyter/kernels/pathmc`, scoped per-venv, no global state:

```make
jupyter-kernel: ## Register the pathmc Jupyter kernel for Quarto execution
	uv run python -m ipykernel install --sys-prefix --name pathmc --display-name pathmc
```

This depends on Quarto resolving the venv interpreter (so `jupyter_client` searches that `sys.prefix`), which it should under `uv run` since `.venv/bin` leads `PATH` — but worth confirming locally with a `make refreeze-docs` on a page that actually executes before switching. If it doesn't resolve, `--user` is the right call and this note can be ignored.

**3. `--display-name`.** Without it the spec's display name defaults to `Python 3 (ipykernel)`, which is what shows up in Jupyter/VS Code kernel pickers. Harmless for Quarto (it matches on the `--name` slug), but confusing for anyone opening a notebook against this env. One extra flag.

**4. Build targets now mutate state outside the repo.** Making `docs` and `refreeze-docs` depend on `jupyter-kernel` means a plain doc build writes to `~/.local/share/jupyter` (or `.venv/` with `--sys-prefix`). Worth being deliberate about, especially given `freeze: true` means ordinary `make docs` runs spawn no kernel at all — the dependency is pure insurance for the cache-miss path. Fine as-is; just noting it's a side effect on a build target.

**5. Test plan.** The second box (`make docs` after switching worktrees) is unchecked, and that's the case the PR exists for. Worth ticking before merge.

Nothing blocking. (1) is the one I'd most want in before merge, since a fourth copy of the command is how these drift.

drbenvincent:
Thanks for the review — addressed in abd6fa6:

**1. Consolidated the install command**
`ipykernel install` now lives only in the Makefile `jupyter-kernel` target. CI runs `make docs` (which depends on it), and `CONTRIBUTING.md`, `AGENTS.md`, and the `great-docs.yml` comment all point there instead of duplicating the command.

**2. `--sys-prefix` + `--display-name`**
Kernel installs into `.venv/share/jupyter/kernels/pathmc`, scoped per checkout rather than clobbering a machine-global user kernelspec.

**3. `JUPYTER_PATH` on docs targets**
Confirmed locally that Jupyter still prefers a stale user-level `pathmc` kernel over `--sys-prefix` by default, so `make docs` and `make refreeze-docs` now set `JUPYTER_PATH=$(CURDIR)/.venv/share/jupyter` to force the venv-scoped spec. Verified with `great-docs freeze` on `dag_testing.qmd`.

**4. Docs updates**
Dropped "once" from CONTRIBUTING (since `make setup`/`make docs` handle registration). Single-page freeze instructions in CONTRIBUTING and AGENTS.md include the `JUPYTER_PATH` prefix for ad-hoc `great-docs freeze` outside the Make targets.

Re note 4 (build targets mutating state): accepted as-is — with `--sys-prefix` the side effect stays inside `.venv/`, and the dependency is insurance for cache-miss paths only.

juanitorduz (approved):

PR #383, opened 2026-08-06
Harden GitHub Pages deploy against re-runs and transient failures

drbenvincent:
## Summary

Docs **builds** have been passing on `main` today, but **publishing** to GitHub Pages has been failing intermittently. This PR makes the publish job more resilient in two ways.

### Problem 1: "Re-run failed jobs" makes things worse

When a Pages deploy times out, clicking **Re-run failed jobs** uploads a second artifact also named `github-pages` into the *same* workflow run. `actions/deploy-pages` then errors immediately:

> Multiple artifacts named "github-pages" were unexpectedly found for this workflow run.

This is a known limitation of `deploy-pages` — artifacts persist across attempts within one run.

**Fix:** name the Pages artifact `github-pages-${{ github.run_attempt }}` (on both upload and deploy) so each attempt uses its own artifact.

### Problem 2: GitHub Pages backend is flaky

Several repos (including ours) are hitting deploy timeouts where the status never advances past `deployment_queued` or `deployment_in_progress`. This matches [actions/deploy-pages#406](https://github.com/actions/deploy-pages/issues/406) and the [GitHub Pages deployment-lag incident on 2026-08-06](https://www.githubstatus.com/). It is a GitHub-side issue, not a docs build regression.

**Fix:** retry `deploy-pages` up to 3 times with a 30-second pause between attempts. Tries 1 and 2 use a 3-minute timeout (healthy deploys take ~1 min); try 3 keeps the default 10-minute timeout as the patient last attempt.

## What this does *not* fix

If GitHub Pages is completely stuck for hours, retries may still fail. In that case the site stays on the last successful deploy until GitHub recovers. This PR avoids re-run artifact collisions and gives transient errors a fair chance to succeed without tripling the worst-case wait on every attempt.

## Test plan

`publish-docs` is gated on `if: github.ref == 'refs/heads/main'`, so **this PR only exercises Build Docs**. The retry/artifact logic is untested until the first post-merge push.

- [x] **Build Docs** passes on this PR (confirms SHA pinning in `build-docs` resolves)
- [ ] After merge: confirm the next `main` push publishes docs once GitHub Pages is healthy
- [ ] If publish fails once, verify **Re-run failed jobs** no longer hits the duplicate-artifact error
- [ ] If all three deploy attempts fail at `Require Pages deploy`, that means Pages is down — not that this workflow is wrong

**Note on red CI checks during the Pages incident:** transient `Failed to resolve action download info` failures in `Set up job` are GitHub Actions infra, not findings about this workflow. Re-run those jobs once the service recovers.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Deploy to GitHub Pages
         id: deployment_try1
         continue-on-error: true
-        uses: actions/deploy-pages@v5
+        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5
         with:
           artifact_name: ${{ env.PAGES_ARTIFACT }}
 
@@ -103,7 +103,7 @@ jobs:
         id: deployment_try2
         if: steps.deployment_try1.outcome == 'failure'
         continue-on-error: true
-        uses: actions/deploy-pages@v5
+        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5
         with:
           artifact_name: ${{ env.PAGES_ARTIFACT }}
 
@@ -118,7 +118,7 @@ jobs:
         if: >-
           steps.deployment_try1.outcome == 'failure' &&
           steps.deployment_try2.outcome == 'failure'
-        uses: actions/deploy-pages@v5
+        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5
         with:
           artifact_name: ${{ env.PAGES_ARTIFACT }}
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,12 +16,12 @@ jobs:
     # Authors refresh the cache locally via `great-docs freeze <page>` (see the
     # "Building the docs" section of AGENTS.md, "Notebooks are frozen").
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
         with:
           fetch-depth: 0 # Full history for accurate page timestamps
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.2.0
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
 
       - name: Install package and dependencies
         # uv sync installs the locked dependency set from uv.lock (same as
@@ -29,7 +29,7 @@ jobs:
         run: uv sync --all-extras
 
       - name: Set up Quarto
-        uses: quarto-dev/quarto-actions/setup@v2
+        uses: quarto-dev/quarto-actions/setup@8a96df13519ee81fd526f2dfca5962811136661b # v2
 
       - name: Build docs
         run: make docs
@@ -51,7 +51,7 @@ jobs:
           done
 
       - name: Save docs artifact
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: docs-html
           path: great-docs/_site
@@ -71,17 +71,17 @@ jobs:
     environment:
       name: github-pages
       url: >-
-        ${{ steps.deployment.outputs.page_url ||
+        ${{ steps.deployment_try3.outputs.page_url ||
         steps.deployment_try2.outputs.page_url ||
         steps.deployment_try1.outputs.page_url }}
     steps:
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
         with:
           name: docs-html
           path: great-docs/_site
 
       - name: Upload Pages artifact
-        uses: actions/upload-pages-artifact@v5
+        uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9 # v5
         with:
           path: great-docs/_site
           name: ${{ env.PAGES_ARTIFACT }}
@@ -94,8 +94,9 @@ jobs:
         uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5
         with:
           artifact_name: ${{ env.PAGES_ARTIFACT }}
+          timeout: 180000 # fail fast; healthy deploys take ~1 min
 
-      - name: Wait before Pages deploy retry
+      - name: Wait before Pages deploy retry 2
         if: steps.deployment_try1.outcome == 'failure'
         run: sleep 30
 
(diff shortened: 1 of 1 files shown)


drbenvincent:
Addressed the zizmor review: all three `deploy-pages` steps are now pinned to `cd2ce8fcbc39b97be8ca5fce6e763baed58fa128` (# v5), matching the repo's SHA pinning convention.

daimon-pymclabs (commented):
## Review

The diagnosis is correct and both failure modes are confirmed in this repo's own logs. The mechanism of each fix is sound. One sizing issue is worth addressing before merge, plus some naming nits.

### What I verified

**Problem 1 is real and observed here.** Attempt 2 of run [31099742237](https://github.com/pymc-labs/pathmc/actions/runs/31099742237) failed with exactly the error in the description:

```
Error: Multiple artifacts named "github-pages" were unexpectedly found
for this workflow run. Artifact count is 2.
```

And the fix targets the right code path. `deploy-pages`' `getArtifactMetadata` does a bare exact-name filter over every artifact in the run, with no attempt scoping:

```js
const filteredArtifacts = response.data.artifacts.filter(a => a.name === artifactName)
if (artifactCount > 1) { throw new Error(`Multiple artifacts named ...`) }
```

So a per-attempt name genuinely disambiguates. `upload-pages-artifact@v5` does accept `name` (default `github-pages`) and `deploy-pages@v5` does accept `artifact_name`, so both inputs are valid. The upload step runs once per attempt, so the count stays at 1 across all three deploy tries within an attempt.

**Problem 2 is real too**, and it is the stall described in actions/deploy-pages#406 rather than a build regression. Runs 31112336170 and 31108994378, plus attempts 1 and 3 of 31099742237, all show `Build Docs: success` and then:

```
Current status: deployment_in_progress   (x120, every 5s)
##[error]Timeout reached, aborting!
Canceled deployment with ID 7523e53d...
```

**The retry design is safe.** I checked that on timeout the action calls `this.cancel()` and explicitly cancels the pending deployment before exiting, so retry N+1 is not racing a deployment that is still queued. That was my main worry and it is unfounded.

**`environment.url` ordering is fine.** `page_url` is set immediately after `create()`, before `check()` polls, so a failed attempt does populate its output. But the fallback prefers the later attempt and skipped steps yield empty strings, so the resolved URL is always the successful attempt's. The YAML parses, `env` is a valid context in both job-level `env` and step `with`, and `zizmor` passes.

### 1. Each attempt burns the full 10 minutes, so worst case goes from ~10 min to ~31 min

This is the one I would change. `deploy-pages` hard-caps and defaults its timeout to 10 minutes:

```js
const MAX_TIMEOUT = 600000
this.timeout = MAX_TIMEOUT                      // constructor
this.timeout = !timeoutInput || timeoutInput <= 0
  ? MAX_TIMEOUT : Math.min(timeoutInput, MAX_TIMEOUT)
```

The workflow passes no `timeout`, so every attempt waits the full 600s before conceding. Three of those plus two 30s sleeps is **1860s, about 31 minutes**, up from ~10 today.

For calibration, the last healthy deploy ([run 31080503779](https://github.com/pymc-labs/pathmc/actions/runs/31080503779)) finished in **57 seconds**: `07:22:43 -> 07:23:40`. Waiting 10 minutes to conclude something is wrong is an order of magnitude more than needed.

Setting a short timeout on the early attempts makes the retry actually pay for itself:

```yaml
      - name: Deploy to GitHub Pages
        id: deployment_try1
        continue-on-error: true
        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5
        with:
          artifact_name: ${{ env.PAGES_ARTIFACT }}
          timeout: 180000  # fail fast; healthy deploys take ~1 min
```

Same on retry 2, leaving retry 3 at the default 600000 as the patient last attempt. Worst case becomes ~17 min, and more to the point a *transient* stall recovers in ~3.5 min instead of ~10.5. As written, the PR mostly triples the time it takes to learn that Pages is down.

### 2. Two steps share the name "Wait before Pages deploy retry"

Legal in Actions, but both render identically in the log UI so you cannot tell which wait you are looking at. Suggest `Wait before Pages deploy retry 2` / `... retry 3`.

### 3. `id: deployment` on the third attempt reads as the primary deploy

Keeping the old id on the *last-ditch* attempt is the sort of thing that misleads the next person to edit this job. Renaming to `deployment_try3` and updating the three-way `environment.url` fallback costs nothing.

Relatedly: `Require Pages deploy` only adds value if try3 can fail soft. Since try3 has no `continue-on-error`, a total failure already reds the job and the gate then fires too, giving two error annotations for one problem. Adding `continue-on-error: true` to try3 would make the gate the single source of truth and produce one clean `::error::` instead.

### 4. Pinning, minor

SHA-pinning `deploy-pages` matches what most other workflows in this repo already do (release, codeql, create-pull-request are all pinned), so the direction is right. But `upload-pages-artifact@v5`, `download-artifact@v7` and `checkout@v6` in this same file stay on tags, which makes the pin look incidental. Worth pinning the rest of `docs.yml` in the same pass, or leaving all four on tags. No strong opinion on which.

### On the test plan

Worth noting that `publish-docs` is gated on `if: github.ref == 'refs/heads/main'`, so it is skipped on this PR and none of this can be exercised before merge. Only `Build Docs` ran here. That is unavoidable, just means the first post-merge push to `main` is the real test. The duplicate-artifact half of the fix is the part you can actually confirm deliberately, by letting one publish fail and then clicking **Re-run failed jobs**.

Net: correct fixes for correctly-identified problems. I would add the `timeout` inputs before merging, since without them the retry loop makes the bad case meaningfully worse; the rest is cosmetic.

drbenvincent:
Addressed all review points in 5ce443d:

1. **Fast-fail timeouts** — try 1 and try 2 now pass `timeout: 180000` (3 min); try 3 keeps the default 10 min as the patient last attempt.
2. **Wait step names** — renamed to \`Wait before Pages deploy retry 2\` / \`... retry 3\`.
3. **Step ids** — renamed \`deployment\` → \`deployment_try3\` and updated the \`environment.url\` fallback.
4. **Single failure annotation** — added \`continue-on-error: true\` to try 3 so \`Require Pages deploy\` is the sole source of truth on total failure.
5. **Action pinning** — pinned every action in \`docs.yml\` to a commit SHA (checkout, setup-uv, quarto, upload/download-artifact, upload-pages-artifact, deploy-pages).

daimon-pymclabs:
Follow-up on CI state, since the red check here is misleading.

**The zizmor failure is not a finding.** The job never reached zizmor. It failed in `Set up job`:

```
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
Retrying in 25.207 seconds
Failed to resolve action download info. Error: Internal Server Error
Retrying in 15.853 seconds
##[error]Service Unavailable
##[error]Failed to resolve action download info.
```

That is GitHub's action-resolution service failing before any workflow content is parsed, so it says nothing about the changes in this PR. `test`, `prek`, and the dist build are also sitting at 0s elapsed, consistent with the same stall. A re-run once things settle should clear them.

**Build Docs passed (2m05s).** That was the one load-bearing check I flagged, since the SHA pinning in this PR extended past the `publish-docs` job into `build-docs` (checkout, setup-uv, quarto, upload-artifact). Those pins resolve and the docs build is green, so that added scope is clear.

**External corroboration for the original diagnosis.** githubstatus currently reports Pages as `degraded_performance` with an open incident, "Incident with Pages - Deployment Lag" (opened 2026-08-06T15:03Z). That is precisely the failure mode this PR hardens against, happening live. It's independent confirmation that the `deployment_in_progress` timeouts in runs 31112336170 / 31108994378 / 31099742237 were the Pages backend rather than a docs regression.

One consequence worth planning around: `publish-docs` is gated on `main`, so the retry logic is still unexercised by this PR. The first post-merge push is the real test, and if the Pages incident is still open at that point, expect the retry path to actually fire. A deploy that takes all three attempts and then fails the `Require Pages deploy` gate would mean Pages is down, not that this workflow is wrong.

Issue #327, opened 2026-06-21
Suspected / under-tested bug classes to audit (follow-up to #316, #326)


daimon-pymclabs:
## Purpose

Spun out of #316 (the `lag()` logp-graph bug) and #326 (the adversarial-testing strategy). This issue tracks **suspected / under-tested bug classes** elsewhere in pathmc. Each is likely a **separate PR**, so it's a checklist.

**Why these are plausible:** the test suite's "replace posterior with predictive sampling" speed trick only exercises the **forward/generative graph**. Everything below lives somewhere that trick can't see — the **logp/gradient graph**, the **graph algorithms**, the **causal queries** (`do`/`effect`/counterfactual), the **effect math**, or **data-shape assumptions**. Each item notes a cheap oracle and a confidence tag: **[confirmed]** by code inspection, **[concrete]** an assumption visible in code, or **[hypothesis]** to probe.

Confirmations below are against `pathmc @ 8ac501e`.

## Checklist

- [x] **1. `_use_observed_carry` leaks observed outcomes into prediction / counterfactual — [confirmed].** Fixed in #387.
  `fit()` sets the `pm.Data` flag `_use_observed_carry` to `1` (`model.py:231`), making the scan carry use *observed* outcomes. `predict()` (`model.py:557`) and `do()` (`model.py:912`) both call `pm.sample_posterior_predictive` **without resetting the flag to `0`** — so any model with a lag of an endogenous/observed outcome (`y ~ lag(y)`) generates predictions/counterfactuals using the real future instead of recursive generation. Predictions look artificially good. Scope: endogenous-lag models (exogenous lag like #316 is unaffected). *Fix:* reset the flag to `0` for generative queries (or context-manage it); *test:* posterior-predictive of `y ~ lag(y)` must not reproduce held-out `y` better than the recursive generative model.

- [x] **2. Interventions through the scan (`pm.do`) — [hypothesis, high].** Fixed in #404.
  The scan exists "so `pm.do()` handles interventions natively" (`compile.py` docstring). If intervening doesn't correctly sever incoming edges *inside the temporal recursion*, or the carry leaks the pre-intervention value, you get a wrong-but-finite effect that samples cleanly. *Oracle:* linear-Gaussian SCM has closed-form interventional means (products of path coefficients) — compare `do()` / `effect()` output to the analytic value. Rides the same code path as #316.

- [x] **3. Balanced-panel assumption — [concrete, high].** Fixed in #388.
[...]

PR #387, opened 2026-08-07
Force generative carry off inside do() (towards #327)

daimon-pymclabs:
Towards #327, sub-components **1 and 12** (`do()` carry leak / observed-carry guard parity).

Adds a `_forced_generative_carry(model)` context manager in `pathmc/simulate.py` and wraps all four `pm.do(gen_model, replacements)` call sites (two in `run_do_pymc`, two in `run_do_panel_unified`) so `do()` forces the generative model's `_use_observed_carry` flag off and restores it afterwards. This matches the guard `predict()` already had. The pattern is duplicated locally rather than imported, to avoid a circular import.

**Scope, stated honestly:** this is defensive invariant hardening, not a fix for a currently reachable user-facing miscomputation. With the flag artificially set, the without-fix run mismatches on 96% of elements (0.189 max relative difference), so the guard matters *if* the flag leaks. On today's code paths it never does, because normal `fit()`/`predict()` always operate on a separate `pm.observe()` clone of the generative model, not the generative model itself — so `do()`'s `gen_model` never sees a leaked flag in practice. This PR closes the latent hazard (the guard would only bite for a model where the observation and generative models were literally the same object, which isn't a state `fit()`/`predict()` produce today) and makes the invariant explicit rather than implicit, so future code changes can't silently reintroduce the leak.

**Tests:** `tests/test_panel_predict_carry.py` covers, for both `kind="mean"` and `kind="predictive"` on the panel (`run_do_panel_unified`) path: that `do()` forces the flag off even when the ambient flag is leaked to 1, that the previous value is restored immediately after `do()` returns (not just in a `finally` that would mask a broken restore), that the flag is restored when the wrapped call raises, and a full `fit -> do -> predict -> do` reuse cycle that checks the generative and observation models' flags never cross-contaminate. The cross-sectional `run_do_pymc` call sites are not covered by a flagged-model test: constructing a fitted panel model with the observed-carry flag and exercising it through the cross-sectional path is not currently possible, because that path errors on scan-compiled panel models independent of this change (filed as #394). The `_forced_generative_carry` guard on those two call sites is therefore defensive-only until #394 is resolved.

**Verification:** CI test job on the current commit: 1404 passed, 0 skipped, in 1790s (0:29:50).

### DIFF
--- a/pathmc/simulate.py
+++ b/pathmc/simulate.py
@@ -26,7 +26,8 @@ handles temporal propagation natively.
 from __future__ import annotations
 
 import warnings
-from collections.abc import Mapping
+from collections.abc import Iterator, Mapping
+from contextlib import contextmanager
 from typing import Any
 
 import narwhals.stable.v1 as nw
@@ -49,6 +50,47 @@ from pathmc.panel import PanelInfo
 from pathmc.reprs import ReprSpec, ResultReprMixin
 
 __all__ = ["DoResult", "EstimandResult"]
+
+
+#: pm.Data flag on scan-compiled panel models. Mirrors
+#: ``pathmc._model._OBSERVED_CARRY_FLAG``; kept as a separate literal here
+#: (rather than importing from ``pathmc._model``) to avoid a circular import,
+#: since ``_model.py`` imports this module.
+_OBSERVED_CARRY_FLAG = "_use_observed_carry"
+
+
+@contextmanager
+def _forced_generative_carry(model: pm.Model) -> Iterator[None]:
+    """Force the observed-carry flag off for the duration of the block.
+
+    ``do()`` must always forward-simulate using the model's own propagated
+    values -- never conditioning on observed data, which is what
+    ``predict(one_step_ahead=True)`` is for. Interventional queries that
+    silently fell back to observed carries would mix "what actually
+    happened" into "what would happen under this intervention".
+
+    The flag defaults to 0 on a freshly compiled generative model and stays
+    there in the common case (``fit()``/``predict()`` operate on a separate
+    ``pm.observe()`` clone), but nothing previously *forced* it for ``do()``.
+    That made correctness depend on an implicit invariant rather than an
+    explicit one -- e.g. it silently breaks for models with no non-latent
+    observed vars, where the generative and observation models are the same
+    object. Forcing (and restoring) the flag here, exactly like
+    ``pathmc._model._observed_carry`` does for ``predict()``, removes that
+    dependency. A no-op on models without the flag (anything but a
+    scan-compiled panel model with a lagged endogenous term).
+    """
+    if _OBSERVED_CARRY_FLAG not in model.named_vars:
+        yield
+        return
+
+    flag = model[_OBSERVED_CARRY_FLAG]
+    previous = flag.get_value()
+    flag.set_value(np.array(0, dtype="int8"))
+    try:
+        yield
+    finally:
+        flag.set_value(previous)
 
 
 # ---------------------------------------------------------------------------
@@ -988,7 +1030,8 @@ def run_do_pymc(
         replacements[key] = arr.astype(target_dtype)
 
     if kind == "mean":
-        do_model = pm.do(gen_model, replacements)
+        with _forced_generative_carry(gen_model):
+            do_model = pm.do(gen_model, replacements)
 
         mean_det_names: dict[str, str] = {}
         expr_replacements: dict[Any, Any] = {}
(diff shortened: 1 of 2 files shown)


drbenvincent (commented):
## Findings

- **Medium — `tests/test_panel_predict_carry.py:124-130`**: The test resets the carry flag to `0` in `finally` before asserting anything. A broken context manager that fails to restore the prior value `1` would still pass. Capture and assert the flag immediately after `do()` returns, then restore it in cleanup; also add an exception-path assertion.
- **Medium — `pathmc/simulate.py:1095-1102, 1240-1247`**: The regression test only exercises `kind="mean"`, while the new helper wraps separate predictive branches. Parameterize the leaked-flag test over both `mean` and `predictive` so either branch cannot regress silently.
- **Low — `pathmc/simulate.py:1033-1034, 1095-1096`**: The cross-sectional `run_do_pymc` call sites are not covered by the new tests; the reuse test uses `simulate_over="time"` and therefore exercises `run_do_panel_unified`. Add a focused flagged-model test for the cross-sectional mean/predictive paths.

drbenvincent (commented):
Independent review: the guard is semantically correct and exception-safe, but this is defensive invariant hardening rather than a currently reachable user-facing bug because normal fit()/predict() use a separate observed clone. Please describe that scope explicitly. The latest tests cover the relevant panel mean/predictive branches and restoration; update the stale PR description test count once the current test job completes.

daimon-pymclabs:
Addressed the review feedback. Summary of what was already in place vs. what I verified/changed:

**Review 1**

- `tests/test_panel_predict_carry.py:124-130` (flag reset before assertion) — already fixed in commit e5bbbdd. `test_do_forces_generative_carry_even_if_leaked` now asserts `flag.get_value() == 1` (restored) immediately after `do()` returns, *before* the test's own cleanup resets it to 0 — see `tests/test_panel_predict_carry.py:169-171`. A separate `test_flag_restored_when_do_raises` (`tests/test_panel_predict_carry.py:184-201`) monkeypatches `pm.do` to raise and asserts the flag is still restored to 1 on the exception path, wrapped in `try/finally` for cleanup.
- `pathmc/simulate.py:1095-1102, 1240-1247` (parameterize mean/predictive) — already done. `test_do_forces_generative_carry_even_if_leaked` is `@pytest.mark.parametrize("kind", ["mean", "predictive"])` (`tests/test_panel_predict_carry.py:110`), spies on `pm.do` to record the flag value seen at each call site, and asserts `seen == [0, 0]` for both branches of `run_do_panel_unified`.
- `pathmc/simulate.py:1033-1034, 1095-1096` (cross-sectional `run_do_pymc` coverage) — I verified this is **not currently reachable**, not just asserted in a docstring. I reproduced it directly: fitting `y ~ x + lag(y)` with `panel={...}` and then calling `.do(set={"x": 0.0}, kind="mean")` (no `simulate_over="time"`) raises `ValueError: Shape of data ((75,)) does not match shape of type ((None, None))` before any flag-relevant code runs. `_model.PathModel.do()` only dispatches scan-compiled panel models to `run_do_panel_unified` when `simulate_over="time"`; otherwise it falls through to `run_do_pymc`, which flattens interventions to `(n_rows,)` against a `(n_times, n_units)`-shaped scan model — a pre-existing shape bug unrelated to this change. Since the observed-carry flag only exists on scan-compiled panel models, and those can't complete the cross-sectional path, there's no way to build a "flagged model exercising `run_do_pymc`" test today. This was filed as pymc-labs/pathmc#394 with a standalone reproducer, and the two `_forced_generative_carry` guards on `run_do_pymc`'s call sites (`pathmc/simulate.py:1033-1034`, `1095-1096`) are documented as defensive-only pending that fix. I did not add a synthetic/mocked test for this since it would test a code path that cannot be reached by any real model construction, and would give false confidence.

**Review 2**

- Updated the PR description to state explicitly, in a dedicated "Scope, stated honestly" section, that this is defensive invariant hardening rather than a fix for a currently-reachable user-facing miscomputation — normal `fit()`/`predict()` operate on a separate `pm.observe()` clone of the generative model, so `do()`'s `gen_model` never actually sees a leaked flag today.
- Replaced the stale "1374 passed / 2 skipped in 10m01s" figure with the actual CI test job result on the current commit (e5bbbdd): **1404 passed, 0 skipped, in 1790s (0:29:50)**, pulled from the `test` job log on this PR's latest CI run.

**Verification run locally (foreground, targeted):**
```
uv run pytest tests/test_panel_predict_carry.py tests/test_do.py -q
# 16 passed, 1 warning in 50.77s
```
`uv run prek run --all-files` passes clean (ruff, ruff-format, mypy, license headers).

No source or test changes were needed beyond what commits f5d969a/e5bbbdd already contained — only the PR description was updated in this pass.

drbenvincent (approved):

PR #389, opened 2026-08-07
Make Adstock l_max/normalize configurable for pymc-marketing parity (towards #327)

daimon-pymclabs:
Towards #327, sub-component **7** (adstock parity with `pymc_marketing`).

`pathmc/transforms.py` hardcoded `l_max = 12` as a class attribute and had no `normalize` option at all, despite a docstring claiming parity with `pymc_marketing.geometric_adstock`. This adds `normalize` to `_geometric_adstock()` and a real `__init__(l_max=12, normalize=False)` on `Adstock`, threaded through `apply_pymc` and `_apply_pymc_panel`. Defaults are unchanged, so existing models produce identical numbers.

**Scan/vectorized divergence — resolved by explicit rejection, not by fixing scan:** `Adstock.step()` inside `pytensor.scan` (used for panel models with temporal dependencies, `pm.do()`, and predictive sampling on them) implements the *unbounded, unnormalized* recursion `y_t = x_t + decay * y_{t-1}`, which disagrees with the truncated/normalized vectorized kernel for non-default configs. Carrying finite-window state through scan to fix this was rejected: the scan path today applies the unbounded recursion even for the *default* `l_max=12`, so bounding it to a true 12-lag window would change existing default-configured panel `do()`/`predict()` numerics for current users — worse than the original bug. Instead, `Adstock.step()` now raises `NotImplementedError` for any non-default `l_max`/`normalize` on scan-compiled models, and `Adstock.__init__` validates `l_max` (int >= 1) and `normalize` (bool).

**Tests:** 55 parameterised tests in `tests/test_adstock_parity.py` (kernel parity, `__init__` validation, scan-rejection of non-default configs, default-config scan compilation still works), plus the existing transforms/panel/do suites. Reference kernel values are hardcoded behind an `importorskip` because `pymc-marketing` 0.18.0 currently fails to import against pymc 6.2.0 (`extract_dims` moved).

**Verification:** targeted runs after the latest changes: `test_adstock_parity.py` + `test_transforms_compile.py` + `test_transforms_do.py` = 78 passed; panel suites (`test_panel.py`, `test_panel_do.py`, `test_panel_engines.py`, `test_panel_predict_carry.py`, `test_panel_smoke.py`, `test_panel_by_time.py`) = 72 passed; `test_do.py` + `test_do_predictive.py` + `test_do_regressions.py` + `test_do_slopes.py` + `test_compile.py` + `test_public_api.py` = 37 passed. The broad suite has not been run against this branch specifically.

### DIFF
--- a/pathmc/transforms.py
+++ b/pathmc/transforms.py
@@ -186,12 +186,21 @@ class Adstock(Transform):
     match a reference model's configuration, e.g.
     ``register_transform(Adstock(l_max=8, normalize=True))``.
 
-    Note that :meth:`step`, used inside ``pytensor.scan`` for panel
-    interventions with temporal dependencies, implements the *unbounded*
-    recursion ``y_t = x_t + decay * y_{t-1}`` — the ``l_max -> inf`` limit
-    of the truncated kernel above. For ``decay`` close to 1 with a small
-    ``l_max`` the two paths can disagree; increase ``l_max`` to tighten
-    the approximation.
+    Panel models with temporal dependencies (``lag()`` or another
+    ``adstock()`` in the same equation, or any ``adstock()`` combined
+    with panel data — see ``_has_temporal_deps`` in ``pathmc.compile``)
+    are compiled with ``pytensor.scan`` and go through :meth:`step`
+    instead of the vectorized kernel. :meth:`step` only implements the
+    *unbounded, unnormalized* recursion ``y_t = x_t + decay * y_{t-1}``
+    (the ``l_max -> inf``, ``normalize=False`` limit of the truncated
+    kernel above), so it cannot honour a non-default ``l_max`` or
+    ``normalize=True`` without silently disagreeing with the vectorized
+    path. Rather than silently ignoring those options on scan-compiled
+    models, :meth:`step` raises ``NotImplementedError`` for any
+    non-default configuration. Use the default ``Adstock()`` for panel
+    models with temporal dependencies; non-default configurations are
+    only supported for cross-sectional models and panel models without
+    temporal dependencies, which use the vectorized kernel.
     """
 
     name = "adstock"
@@ -199,19 +208,40 @@ class Adstock(Transform):
         "decay": ParamSpec(constraint="unit_interval", default_prior="Beta(2, 2)"),
     }
 
-    def __init__(self, l_max: int = 12, normalize: bool = False) -> None:
+    #: Configuration the scan recursion in :meth:`step` actually implements.
+    #: The guard there compares against these, so changing a default cannot
+    #: silently widen what scan-compiled models accept.
+    DEFAULT_L_MAX = 12
+    DEFAULT_NORMALIZE = False
+
+    def __init__(
+        self, l_max: int = DEFAULT_L_MAX, normalize: bool = DEFAULT_NORMALIZE
+    ) -> None:
         """Create a geometric-adstock transform.
 
         Parameters
         ----------
         l_max : int
             Number of lags (including lag 0) included in the truncated
-            carryover sum. Matches ``pymc_marketing``'s ``l_max``.
+            carryover sum. Matches ``pymc_marketing``'s ``l_max``. Must be
+            an integer ``>= 1``.
         normalize : bool
             If ``True``, divide by the sum of the lag weights so they sum
             to 1. Matches ``pymc_marketing``'s ``normalize``.
+
+        Raises
+        ------
+        ValueError
+            If ``l_max`` is not an integer ``>= 1``, or ``normalize`` is
+            not a ``bool``.
         """
-        self.l_max = l_max
+        if isinstance(l_max, bool) or not isinstance(l_max, (int, np.integer)):
+            raise ValueError(f"l_max must be an int >= 1, got {l_max!r}.")
+        if l_max < 1:
+            raise ValueError(f"l_max must be >= 1, got {l_max!r}.")
+        if not isinstance(normalize, bool):
+            raise ValueError(f"normalize must be a bool, got {normalize!r}.")
+        self.l_max = int(l_max)
         self.normalize = normalize
 
     def apply_pymc(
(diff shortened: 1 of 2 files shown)


drbenvincent (commented):
## Findings

- **High — `pathmc/transforms.py:189-194, 263-267`**: `l_max` and `normalize` affect the vectorized adstock kernel, but `Adstock.step()`—used by the panel scan compilation and time-based `do()` path—still implements unbounded, unnormalized recursion. The same configured transform therefore produces different numerics depending on the execution path, and the new options do not provide the advertised parity for panel models. Please either implement bounded/normalized scan state and add a panel regression comparing the paths, or explicitly scope/reject these options for scan models rather than silently ignoring them.

drbenvincent (changes requested):
Independent review: request changes. The vectorized kernel honors l_max and normalize, but panel scan, do(), and predictive paths call Adstock.step(), which still implements unbounded and unnormalized recursion. Adstock(l_max=3, normalize=True) therefore has different semantics depending on execution path. Please either carry finite-window state through scan and add panel fit/do/predict parity tests, or reject non-default scan configurations explicitly instead of silently ignoring them. Also validate l_max >= 1 and normalize as a boolean in __init__; invalid values currently bypass the vectorized guard on scan paths.

daimon-pymclabs:
Addressed both reviews' main finding on the scan/vectorized divergence, plus the requested `__init__` validation.

**Chosen resolution: (b) — explicitly reject non-default configs on scan paths.**

I evaluated (a) first (carry finite-window scan state so `step()` truncates at `l_max` and honours `normalize`). It doesn't work as a same-PR fix: `Adstock.step()` (`pathmc/transforms.py`, used by the scan compilation in `pathmc/compile.py:1361` and reused by `pm.do()`/predictive sampling on scan-compiled models — see `pathmc/simulate.py` around line 1105) currently implements the *unbounded* recursion `y_t = x_t + decay * y_{t-1}` for every configuration, including the existing default `l_max=12`. Bounding that recursion to a true 12-lag window to match the vectorized kernel would change the numerics of every *currently shipping default-configured* panel model with adstock under `do()`/`predict()` — silently changing behaviour for existing users, which the review flagged as worse than the original bug. So I went with (b) instead, and say so explicitly rather than leaving the divergence undocumented.

**What changed** (`pathmc/transforms.py`):
- `Adstock.step()` (around line 285) now raises `NotImplementedError` when `self.l_max != 12 or self.normalize` is truthy, i.e. for any non-default configuration, with a message pointing at the vectorized path and explaining why the scan recursion can't honour those options.
- `Adstock.__init__` (around line 202) now validates: `l_max` must be a non-bool `int` (or `np.integer`) `>= 1`; `normalize` must be a `bool`. Both raise `ValueError` with the offending value.
- Updated the `Adstock` class docstring to describe the scan-path limitation and point to `_has_temporal_deps` in `pathmc.compile` for what triggers scan compilation.

**Evidence defaults are numerically unchanged:** the `step()` method body for the default path (`decay * state + x_t`) is byte-for-byte unchanged — the only addition is a guard that raises *before* reaching that code, and it only fires for non-default configs. I did not touch scan numerics at all, so there's nothing to regress-test against a "before" baseline beyond confirming defaults still compile and run, which the existing and new tests do. Foreground test runs on this branch:
- `tests/test_adstock_parity.py tests/test_transforms_compile.py tests/test_transforms_do.py` → 78 passed
- `tests/test_panel.py tests/test_panel_do.py tests/test_panel_engines.py tests/test_panel_predict_carry.py tests/test_panel_smoke.py tests/test_panel_by_time.py` → 72 passed
- `tests/test_do.py tests/test_do_predictive.py tests/test_do_regressions.py tests/test_do_slopes.py tests/test_compile.py tests/test_public_api.py` → 37 passed

**New tests** (`tests/test_adstock_parity.py`):
- `TestAdstockInitValidation` — rejects non-positive/non-integer `l_max`, non-bool `normalize`; accepts `np.integer` `l_max`; confirms valid construction still works.
- `TestAdstockScanRejectsNonDefaultConfig` — builds a real panel model (`panel={"unit": "region", "time": "week"}`, `pooling="partial"`) with `adstock()`, registers non-default `Adstock` instances (`l_max=3`, `normalize=True`, both), and asserts `pathmc.model(...)` raises `NotImplementedError` matching `"scan-compiled"`; a companion test confirms the default `Adstock()` still compiles fine on the same scan path. Registry mutation is restored via an autouse fixture.

`uv run prek run --all-files` is clean (ruff-format auto-fixed one import line on the first pass; re-run is all green). Pushed to `327-adstock-parity` at `59c505e`.

daimon-pymclabs:
Small follow-up to the above, included in `59c505e`: the scan guard originally compared against a literal `12`, which would have silently widened what scan-compiled models accept if the constructor default ever changed. It now compares against class-level `Adstock.DEFAULT_L_MAX` / `Adstock.DEFAULT_NORMALIZE` (`pathmc/transforms.py:211-218, 308`), which are also the `__init__` defaults, so the guard and the default cannot drift apart.

Re-verified on the exact pushed commit: `uv run prek run --all-files` clean, `tests/test_adstock_parity.py tests/test_transforms_compile.py tests/test_transforms_do.py` → 78 passed, and the six panel suites (`test_panel.py`, `test_panel_do.py`, `test_panel_engines.py`, `test_panel_predict_carry.py`, `test_panel_smoke.py`, `test_panel_by_time.py`) → 72 passed, confirming default-configured panel models are unaffected.

drbenvincent (approved):
## Re-review (59c505e)

Re-reviewed after the scan-path resolution commit. The earlier **CHANGES_REQUESTED** items are addressed.

### Resolution (b) — explicit rejection on scan paths

`Adstock.step()` now raises `NotImplementedError` before the scan recursion when `l_max != DEFAULT_L_MAX` or `normalize != DEFAULT_NORMALIZE`. The guard compares against class-level `DEFAULT_L_MAX` / `DEFAULT_NORMALIZE` (not a literal `12`), so it cannot drift from constructor defaults.

Verified behaviour:
- Non-default configs (`l_max=3`, `normalize=True`, `l_max=5, normalize=True`) fail at **compile time** on a real scan-compiled panel model (`panel` + `adstock()` + `pooling="partial"`), with a clear message referencing scan compilation.
- Default `Adstock()` still compiles on the same scan path — no regression in the panel/transform/do suites exercised locally.

This is the right trade-off given that bounding scan state to a true 12-lag window would silently change existing default-configured panel `do()` / `predict()` numerics (scan today uses unbounded recursion even at the default `l_max=12`). Documenting that pre-existing default-path approximation in the class docstring is appropriate; this PR correctly prevents the *new* `l_max` / `normalize` options from widening silent divergence.

### `__init__` validation

`l_max` must be a non-bool `int` / `np.integer` `>= 1`; `normalize` must be a `bool`. Invalid values raise `ValueError` at construction time rather than slipping through to scan paths. Covered by `TestAdstockInitValidation`.

### Tests

`tests/test_adstock_parity.py` is solid: numpy-oracle kernel parity (unnormalized + normalized, multiple `alpha`/`l_max`), pinned regression values, configurable constructor wiring through `apply_pymc`, init validation, and scan-path rejection/acceptance. Ran locally on this commit: `test_adstock_parity.py` + `test_transforms_compile.py` + `test_transforms_do.py` → 78 passed; `test_panel.py` + `test_panel_do.py` + `test_panel_engines.py` → 37 passed.

### Default-config path divergence (acknowledged, not a blocker)

For default `Adstock()`, the vectorized kernel remains truncated at 12 lags while scan `step()` remains unbounded — a pre-existing approximation, unchanged by this PR. Non-default configs no longer silently disagree across paths; they are rejected on scan compilation instead.

**Verdict: approve.** Review requests satisfied. Leaving draft status / merge to the author.

PR #391, opened 2026-08-07
Warn when path effects are computed through interaction terms (towards #327)

daimon-pymclabs:
Towards #327, sub-component **5** (effect math oracle).

`compute_path_effect` in `pathmc/effects.py` implements Wright's tracing rule, which is exact for linear-Gaussian models but silently incomplete when an edge's regression carries an interaction term. For `Y ~ b*M + g*M:X + c*X` with `M ~ a*X`, `effect("X -> M -> Y")` returns `a*b` while the true marginal effect is `c + a*b + 2*g*a*X`: correct at X=0, but at X=2 the truth is 1.5 against a constant path-sum of 0.7.

This adds a `UserWarning` rather than changing any returned numbers. Auto-detecting interaction paths and returning an X-dependent effect is an API redesign and is deliberately out of scope.

The warning now also covers the `effects_summary()` / `build_effects_summary()` / `evaluate_defined_params()` path: a caller who only ever calls `effects_summary()` (never `effect()` directly) is warned when a defined parameter's expression (`total := ...`) uses a labeled coefficient whose edge also carries an interaction term, since the reported total can silently omit that interaction's contribution. This addresses review feedback that the original warning only fired from `compute_path_effect()`.

**Tests:** 16 tests in `tests/test_effects_oracle.py`, all passing (12 original + 4 addressing review feedback: interaction warning reaching the summary-only path, and no false-positive warnings on non-interaction models). The parallel-mediators test now also asserts each individual path effect (via `effect()`) against its analytic value, not just the manually-defined total expression.

**Verification:** 82 passed / 0 failed across `tests/test_effects_oracle.py tests/test_effects.py tests/test_causal_queries.py tests/test_interactions.py tests/test_internal_helpers.py`. The broad suite has not been run against this branch specifically.

Trap worth knowing for anyone writing posterior-injection tests: `model._idata.posterior = new_ds` updates the DataTree attribute but not the `idata["posterior"]` node, and `pathmc/idata.py` reads via the bracket path, so `do()` and `effects.py` can end up reading different draws. The existing pattern in `tests/test_causal_queries.py` has the same shape.

### DIFF
--- a/pathmc/effects.py
+++ b/pathmc/effects.py
@@ -19,6 +19,7 @@ Provides the logic behind ``PathModel.effects_summary()`` and
 
 from __future__ import annotations
 
+import re
 import warnings
 from dataclasses import dataclass
 
@@ -37,6 +38,48 @@ __all__ = ["EffectResult"]
 def _has_labeled_terms(spec: Spec) -> bool:
     """Check whether any regression term has a user-supplied label."""
     return any(term.label is not None for reg in spec.regressions for term in reg.terms)
+
+
+def _interacting_vars(reg) -> set[str]:
+    """Variables that participate in any interaction term of a regression."""
+    interacting: set[str] = set()
+    for t in reg.terms:
+        if t.interaction_of is not None:
+            interacting.update(t.interaction_of)
+    return interacting
+
+
+def _interaction_affected_labels(spec: Spec) -> dict[str, str]:
+    """Map label -> ``"source -> target"`` for labeled main-effect terms whose
+    edge also carries an interaction term involving the same source variable.
+
+    A defined parameter (or path effect) built from such a label reports only
+    the linear coefficient and silently omits the interaction's contribution
+    to the true, state-dependent marginal effect.
+    """
+    affected: dict[str, str] = {}
+    for reg in spec.regressions:
+        interacting = _interacting_vars(reg)
+        if not interacting:
+            continue
+        for t in reg.terms:
+            if (
+                t.label is not None
+                and t.interaction_of is None
+                and t.variable in interacting
+            ):
+                affected[t.label] = f"{t.variable} -> {reg.lhs}"
+    return affected
+
+
+_INTERACTION_DEFINED_PARAM_WARNING = (
+    "Defined parameter '{name} := {expression}' uses coefficient(s) ({labels}) "
+    "from edge(s) ({edges}) whose regression also has an interaction term "
+    "involving the same source variable. compute_path_effect()-style path "
+    "multiplication only captures the linear (main-effect) coefficient, so "
+    "'{name}' silently omits the interaction's contribution to the true "
+    "(state-dependent) marginal effect."
+)
 
 
 @dataclass(repr=False)
@@ -141,11 +184,36 @@ def evaluate_defined_params(
         Mapping from defined parameter name to computed draws.
     """
     defined_draws: dict[str, np.ndarray] = {}
+    affected_labels = _interaction_affected_labels(spec)
+    # Names (labels or previously-computed defined params) known to omit an
+    # interaction contribution; grows as flagged defined params are chained
+    # into later ones (e.g. total := indirect + c).
+    affected_names = set(affected_labels)
 
     for dp in spec.defined_params:
         namespace: dict = {k: v for k, v in labeled_draws.items()}
         namespace.update(defined_draws)
         namespace["__builtins__"] = {}
+
+        tokens = set(re.findall(r"[A-Za-z_]\w*", dp.expression))
+        used_affected = sorted(tokens & affected_names)
+        if used_affected:
+            edges = sorted({
+                affected_labels[name]
+                for name in used_affected
+                if name in affected_labels
+            })
+            warnings.warn(
+                _INTERACTION_DEFINED_PARAM_WARNING.format(
+                    name=dp.name,
+                    expression=dp.expression,
+                    labels=", ".join(used_affected),
+                    edges=", ".join(edges) if edges else "unknown",
+                ),
+                UserWarning,
+                stacklevel=2,
+            )
+            affected_names.add(dp.name)
 
         draws = eval(dp.expression, namespace)  # noqa: S307
         defined_draws[dp.name] = np.asarray(draws)
--- a/pathmc/effects.py
+++ b/pathmc/effects.py
@@ -19,6 +19,7 @@ Provides the logic behind ``PathModel.effects_summary()`` and
 
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass
 
 import narwhals.stable.v1 as nw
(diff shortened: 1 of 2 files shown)


drbenvincent (commented):
Reviewed with Bugbot and an independent defect-focused pass.

No actionable correctness or test-coverage findings. The interaction warning and effect-oracle tests are consistent with the PR's stated scope.

pathmc/effects.py
```diff
@@ -284,6 +285,16 @@ def compute_path_effect(
...
     """Compute the effect along a specified causal path.
 
+    Each edge's contribution is the linear (main-effect) regression
+    coefficient of the source variable on the target; the path effect is
+    the product of these coefficients (Wright's tracing rule). This is
+    exact for linear-Gaussian structural models. If the target's
+    regression also contains an interaction term involving the source
+    variable (e.g. ``Y ~ b*M + g*M:X``), the interaction's contribution is
+    *not* included and the true marginal effect becomes state-dependent
+    (a function of the interacting variable's value); a ``UserWarning``
```
drbenvincent, line 363: The warning text explicitly calls out a defined parameter such as `total := direct + indirect`, but this warning is only emitted from `compute_path_effect()`. A caller that uses `effects_summary()` alone goes through `build_effects_summary()` / `evaluate_defined_params()` and receives no warning for an interaction model, even though the reported total can omit the interaction contribution. Please either cover this summary path (with a regression test) or narrow the warning text/contract.

tests/test_effects_oracle.py
```diff
@@ -0,0 +1,312 @@
...
+    posterior["beta_Y"].loc[{"Y_predictors": "M2"}] = 0.3
+    posterior["beta_Y"].loc[{"Y_predictors": "T"}] = 0.2
+    for v in ["sigma_M1", "sigma_M2", "sigma_Y"]:
+        posterior[v] = posterior[v] * 0 + 1e-6
+    model._idata["posterior"] = posterior
+    return model
+
+
+class TestParallelMediatorsExactIdentity:
+    def test_total_sums_all_paths_exactly(self, pinned_parallel_model):
```
drbenvincent, line 218: This test only evaluates the manually defined expression `c + a1*b1 + a2*b2`; it never calls `effect("T -> M1 -> Y")` or `effect("T -> M2 -> Y")`. Therefore it would still pass if path-effect multiplication dropped or miscomputed one of the mediator paths. Please assert each path effect and compare their sum (including the direct path) with the defined total; the existing `do()` assertion does not exercise `compute_path_effect()`.

daimon-pymclabs:
Addressed both review comments in 64921a3.

**1. `pathmc/effects.py:295` — warning didn't reach the `effects_summary()`-only path.**

Went with the preferred fix (covering the summary path) rather than narrowing the warning text. Added `_interacting_vars()` and `_interaction_affected_labels()` in `pathmc/effects.py` (around line 43): for each regression, this finds labeled main-effect terms whose variable also appears in that regression's interaction term(s) (e.g. `c` on `X -> Y` and `b` on `M -> Y` in `Y ~ b*M + g*M:X + c*X`). `evaluate_defined_params()` (line ~168) now scans each `:=` expression's identifier tokens against this affected-label set and emits the same `UserWarning` if a defined parameter uses one of them — so `total := c + a*b` warns even if the caller never calls `effect()` directly, only `effects_summary()`. The check also propagates transitively: if `total := indirect + c` is flagged, and a later param uses `total`, that one is flagged too. Emits at most one warning per defined parameter (not per draw), and only when an affected label is actually referenced, so non-interaction models stay silent.

New tests in `tests/test_effects_oracle.py`:
- `TestInteractionWarningReachesSummaryPath::test_effects_summary_alone_warns_about_omitted_interaction` — calls only `effects_summary()` on an interaction model with `total := c + a*b` and asserts the `UserWarning` fires (regression test for exactly the gap you flagged).
- `test_evaluate_defined_params_warns_directly` — same check calling `evaluate_defined_params()` directly.
- `test_no_warning_for_non_interaction_model` / `test_no_warning_for_non_interaction_parallel_model` — wrap `effects_summary()` and `effect()` calls on the existing non-interaction fixtures in `warnings.simplefilter("error", UserWarning)` to confirm no spurious firing.

**2. `tests/test_effects_oracle.py:218` — test didn't exercise `compute_path_effect()` for each mediator path.**

Rewrote `TestParallelMediatorsExactIdentity::test_total_sums_all_paths_exactly` to call `model.effect("T -> Y")`, `model.effect("T -> M1 -> Y")`, and `model.effect("T -> M2 -> Y")` individually, assert each against its known analytic value (0.2, 0.30, 0.12), sum them, and compare that sum against both the hardcoded 0.62 and `effects_summary()`'s `total`. This would now fail if path-effect multiplication dropped or miscomputed either mediator path, which the original test (only checking the manually-defined expression) would not have caught.

**Verification:**
- `uv run pytest tests/test_effects_oracle.py -q` → 16 passed (was 12; +4 new tests, 0 removed).
- `uv run pytest tests/test_effects_oracle.py tests/test_effects.py tests/test_causal_queries.py tests/test_interactions.py tests/test_internal_helpers.py -q` → 82 passed, 0 failed. The only `UserWarning` about interactions came from the test that intentionally triggers it; the pre-existing "intervention outside observed range" warnings in `test_causal_queries.py` are unrelated and pre-existing.
- `uv run prek run --all-files` → clean (ruff-format auto-reformatted the two touched files on first run; re-ran and it passed clean).
- Broad/full test suite was not run (out of scope per instructions); only the targeted files above.

Pushed to `327-effects-oracle` at `64921a3`. PR description updated to reflect the new test count and the summary-path coverage; "towards #327" line preserved. Not marked ready for review.

drbenvincent:
## Re-review (64921a3)

Re-reviewed the full diff vs `main` after the follow-up commit. Both prior inline comments are addressed on tip.

### 1. Interaction warning on the `effects_summary()` path — **fixed**

`evaluate_defined_params()` now mirrors the `compute_path_effect()` contract via `_interaction_affected_labels()` / `_interacting_vars()`. Defined parameters that reference labeled main-effect coefficients on edges with interactions (e.g. `total := c + a*b` when `c` sits on `X -> Y` alongside `M:X`) emit the same class of `UserWarning`, including transitive propagation when a flagged defined param is reused (`total := indirect + c`).

`TestInteractionWarningReachesSummaryPath` covers the exact gap: `effects_summary()` alone warns, `evaluate_defined_params()` warns directly, and non-interaction fixtures stay silent under `warnings.simplefilter("error", UserWarning)`.

### 2. Parallel-mediator oracle test — **fixed**

`TestParallelMediatorsExactIdentity.test_total_sums_all_paths_exactly` now calls `effect()` for all three `T -> Y` paths (`T -> Y`, `T -> M1 -> Y`, `T -> M2 -> Y`), asserts each analytic value, sums them, and cross-checks against `effects_summary()`'s `total`. This would fail if `compute_path_effect()` dropped or miscomputed either mediator path.

### New issues

None actionable. The token-scan approach in `evaluate_defined_params()` is a pragmatic match for pathmc's label namespace; warning text is consistent across both entry points; docstring on `compute_path_effect()` documents the limitation clearly.

### CI

All checks green on `64921a3` (test, prek, docs build).

Approving — ready for the author to mark out of draft when they want merge consideration.

drbenvincent (approved):
Both review comments from the initial pass are fully addressed in 64921a3:

1. The interaction warning now reaches `effects_summary()` / `evaluate_defined_params()` (not only `compute_path_effect()`), with regression tests for the summary-only path and negative controls on non-interaction models.
2. The parallel-mediator oracle test now exercises `effect()` for every path before summing.

No new correctness or coverage issues found. CI is green. Approving; leaving draft status to the author.

PR #392, opened 2026-08-07
chore: adopt git-ai for AI code attribution

### LINKED ISSUE #386: Adopt git-ai for AI code attribution + document it in CONTRIBUTING.md
## Summary

Adopt [git-ai](https://usegitai.com/docs/get-started) in `pathmc` to record line-level attribution for AI-generated code, following the pilot already running in [`pymc-marketing`](https://github.com/pymc-labs/pymc-marketing/pull/2746).

## What git-ai does

git-ai is an open-source git extension. Coding agents (Claude Code, Cursor, Copilot, Windsurf, and ~12 others) call `git ai checkpoint` as they write, and on commit the attribution is stored in **git notes** under `refs/notes/ai`, linking lines to the agent, model, and session, along with token usage and human overrides. Attribution is self-reported by the agent rather than inferred heuristically, and it survives rebase, squash, stash, and cherry-pick.

Relevant properties for us:

- Local-first. No account, no hosted platform, no cost. The data lives in this repo's git notes.
- **Not a dependency and not required.** Contributors who don't install it simply produce no notes; nothing breaks, and no PR is blocked. There is zero friction for external contributors.
- The hosted dashboard is the vendor's only upsell. Because the underlying data is in git, we can query and visualize it ourselves.

The pilot in `pymc-marketing` is live and producing data (`refs/notes/ai` exists on that remote), so this is a second repo on a proven setup rather than a new experiment.

### 1. CI workflow

Add `.github/workflows/git-ai.yaml`, adapted from `pymc-marketing`. It triggers on pull request `closed` and `synchronize`, installs the git-ai CLI, and runs `git-ai ci github run` to consolidate attribution when a PR is squash- or rebase-merged (without this, squash merges lose the per-commit notes). It needs `contents: write` at the job level and uses no third-party actions.

Two hardening items to apply while porting, so the new file matches this repo's existing conventions and doesn't regress the `zizmor` check:

- Add a top-level `permissions: {}` block and keep the write scope job-local (`zizmor`'s excessive-permissions audit will otherwise flag it).
- Follow the SHA-pinning convention used in the other workflows for any action that gets added.

### 2. `CONTRIBUTING.md` updates

The point of this issue is that the instructions are **persistent and discoverable**, so both core maintainers and first-time external contributors know what git-ai is without having to ask. Add a subsection under the existing **Use of agents** heading (it is the natural home, since it already covers agent-generated PRs) covering:

- **What it is and why we use it**, in two or three sentences: we want an honest, measurable picture of how much of pathmc is agent-written and how that changes over time, and self-reported attribution is more reliable than guessing after the fact.
- **It is optional.** State plainly that installing git-ai is not required to contribute, is not a dependency, will not block a PR, and that not installing it simply means your commits carry no attribution notes. External contributors should not feel they are being asked to install tooling as a condition of contributing.
- **How to install**, for those who opt in: a single `curl -fsSL https://usegitai.com/install.sh | bash`, once per machine, not per repo. Agent hooks are managed automatically.
- **Allow-list note for core devs.** Anyone who scoped `allow_repositories` to a single repo during the `pymc-marketing` pilot must add pathmc, or widen the glob to `https://github.com/pymc-labs/*`, or their pathmc commits are silently ignored. This is the most likely reason for a maintainer's attribution to go missing, so it is worth calling out explicitly.
- **Privacy note.** Attribution notes go into this public repository's git notes. Prompt session contents are redacted and stored outside git, but contributors should know what is recorded before opting in.
- **A link to the git-ai docs** for anything beyond the above.

If a `CONTRIBUTING.md` change alone leaves it too easy to miss, a one-line pointer from `AGENTS.md` to that section is a reasonable addition.

## Caveats

- Attribution completeness is a direct function of adoption. If only some maintainers install it, the notes will undercount AI authorship and overcount human authorship. Any reporting built on this data should be read with that bias in mind, especially early on.
- Contributors who want to inspect the notes locally need to fetch them explicitly (`git fetch origin 'refs/notes/*:refs/notes/*'`); they do not arrive with a default clone.

## Acceptance criteria

- [ ] `.github/workflows/git-ai.yaml` added, with a top-level `permissions: {}` block, and passing `zizmor`
- [ ] `refs/notes/ai` observed on the remote after a merged PR
- [ ] `CONTRIBUTING.md` documents what git-ai is, that it is optional, how to install it, and the allow-list gotcha

thomaspinder:
Add the git-ai CI workflow and document the tool in CONTRIBUTING.md so that line-level AI authorship is recorded in refs/notes/ai and survives squash- and rebase-merges.

Without CI consolidation, GitHub's squash merge orphans the per-commit notes: all 15 attribution notes currently on the remote are attached to commits unreachable from main.

The workflow is ported from the pymc-marketing pilot (#2746) with four changes to match this repo's conventions and keep zizmor quiet: a top-level `permissions: {}` block, a job-local `contents: write` with an explanatory comment, a concurrency group keyed on workflow, PR and action so concurrent runs cannot race each other's note pushes, and `continue-on-error` on the job.

The last of these makes the job advisory rather than a gate. A pull_request event from a fork gets a read-only GITHUB_TOKEN regardless of the permissions block, so the note push can 403 on an external contributor's PR; usegitai.com could equally be down. Neither is the contributor's problem, and CONTRIBUTING.md promises git-ai will never block a pull request, so the workflow now provides that guarantee structurally instead of relying on the failure being rare.

Installing git-ai stays optional: it is not a dependency, not part of the test or lint gates, and will never block a pull request.

Closes #386


daimon-pymclabs:
Pre-mortem review. I checked the premise and the upstream pilot's real run history before writing this, so most of what follows is grounded in observed behaviour rather than speculation.

**The premise checks out.** `refs/notes/ai` on this repo currently holds 20 note objects (the description says 15), and *zero* of the annotated commits are ancestors of `main`. Squash merges have orphaned all of it, exactly as claimed.

**The `continue-on-error` justification also checks out empirically.** Upstream (`pymc-marketing`, same workflow without `continue-on-error`) has 95 successes / 3 failures / 2 skipped over ~2.5 weeks. Both diagnosable failures are precisely the two modes named in the description:
- [run 31020850084](https://github.com/pymc-labs/pymc-marketing/actions/runs/31020850084) — fork PR: `remote: Permission to pymc-labs/pymc-marketing.git denied to github-actions[bot]` / 403 on `push origin refs/notes/ai:refs/notes/ai`.
- [run 30800892739](https://github.com/pymc-labs/pymc-marketing/actions/runs/30800892739) — `RPC failed; HTTP 503` fetching notes from GitHub.

So the reasoning is sound. What follows is what I think goes wrong *later*.

---

### 1. The concurrency group doesn't protect the resource it needs to protect

`refs/notes/ai` is a single repo-global ref. The concurrency group is keyed on `${{ github.workflow }}-${{ pr.number }}-${{ github.event.action }}`, which serialises runs *per PR per action* and therefore permits:

- PR #A's merge run and PR #B's merge run to push concurrently;
- a `synchronize` run and a `closed` run for the **same** PR to run concurrently (different `action` → different group).

From the upstream logs, the final step is `git push --quiet ... origin refs/notes/ai:refs/notes/ai` — no `+`, so a loser in that race gets a non-fast-forward rejection rather than a silent clobber. Good, but combined with `continue-on-error: true` the rejection is swallowed, and for the `closed` event there is no second chance: that event never fires again, so the merged PR's attribution is lost permanently and silently.

This is not a rare-tail scenario. `synchronize` fires on *every push to every open PR*, and each run takes 4–7 minutes end to end (upstream: install at 15:33:07, first git-ai output at 15:37:33; another at 09:19:30 → 09:26:39). Long jobs on a frequent trigger contending for one ref means overlap is close to inevitable over a release cycle.

The narrow fix is a repo-global group that queues rather than cancels:

```yaml
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: false
```

Honest caveat: GitHub keeps at most one pending run per group, so under a burst some *pending* runs get evicted. That is still strictly better than concurrent pushes racing (an evicted `synchronize` is recovered by the next push), but it does not fully protect an evicted `closed` run — which points at item 3.

Also worth separating: `cancel-in-progress: true` on a job whose whole purpose is a one-shot irreversible consolidation is the wrong default regardless of the key. Cancelling a `closed` run destroys work that will never be retried.

### 2. No `timeout-minutes`

The job inherits the 360-minute default. It does a full clone plus network fetches against a third-party installer endpoint, and empirically takes 4–7 minutes. A hang — usegitai.com half-up, a stalled clone — burns six hours of runner time, and because of `continue-on-error` nobody sees a red check to prompt a look. `timeout-minutes: 15` costs one line.

### 3. `continue-on-error` trades a visible harmless failure for an invisible harmful one

I found the two upstream failure modes *because* upstream lets them go red. This PR removes that signal. The description says failures "stay visible in the run logs," but nobody opens logs for a green run — that's the entire failure mode of advisory instrumentation.

The empirical failure rate is ~3%, and the causes (fork PRs, GitHub 5xx) are recurring, not one-offs. Left as-is, the plausible six-month outcome is that attribution quietly stops consolidating and the first anyone notices is when someone runs `git-ai stats` for a blog post and the numbers are wrong in a direction nobody can reconstruct.

Suggestion: keep `continue-on-error` (it's the right call for contributor experience) but add a self-healing backstop — the same job on `schedule:` weekly plus `workflow_dispatch:`. That single addition covers *four* separate holes at once: silent CI failures, the fork-PR case that can never push, races lost under item 1, and the existing orphaned notes in item 5. Worth checking whether `git-ai ci github run` works outside a `pull_request` event context, or whether a different subcommand is needed for a scheduled sweep.

### 4. Unpinned `curl | bash` in a job that holds `contents: write`

Every action in this repo is SHA-pinned (`actions/checkout@34e1148...`, `github/codeql-action/upload-sarif@e46ed2c...`). This workflow then pipes an unpinned remote script from a third-party domain straight into `bash`, in a job carrying a write-capable token.

The step-scoping is thoughtful — `GITHUB_TOKEN` is only on the second step, and skipping `actions/checkout` means there's no persisted credential for the installer to read. But the installer writes the `git-ai` binary that the *next* step executes with the token in its environment, so a compromised or hijacked `usegitai.com` yields arbitrary pushes to a public repo that publishes to PyPI. zizmor won't flag this.

Beyond security, an unpinned installer means CI behaviour can change overnight with no PR — and under `continue-on-error`, a breaking upstream change is invisible. Pinning a release version (and a checksum if git-ai publishes one) fixes both.

### 5. The 20 existing orphaned notes aren't backfilled, and the docs don't say so

The new CONTRIBUTING section tells a reader to `git fetch origin 'refs/notes/*:refs/notes/*'` then `git log --show-notes=ai` / `git-ai stats`. Today that produces nothing, because every existing note is attached to an orphaned pre-squash commit and this PR only fixes PRs merged *from now on*. First person to follow the instructions concludes the tooling is broken. Either backfill once, or add a sentence saying history before this PR is not consolidated.

### 6. Privacy timing is broader than the doc implies

Two things the section doesn't quite say:

- The `synchronize` trigger pushes notes to this public repo on **every push to an open PR** — so agent name, model, session identifiers and token counts are published *before* review, and stay published even if the PR is closed unmerged. The doc frames publication as a consequence of contributing; it's actually a consequence of pushing.
- The doc actively recommends `git-ai config --add allow_repositories 'https://github.com/pymc-labs/*'`. That's convenient, but it silently enrols every current and future pymc-labs repo the person clones. Worth flagging in the same breath, since the surrounding text is otherwise careful about informed consent.

### Nits

- File is `.yaml`; all five existing workflows here are `.yml`. Upstream used `.yaml`, so this is inherited, but consistency within this repo seems worth more.
- Description says 15 orphaned notes; I count 20 note objects on `refs/notes/ai`.
- CONTRIBUTING says the job "re-consolidates attribution when a pull request is squash- or rebase-merged." It also runs on every `synchronize` and on merge-commit merges. Minor, but it's the sentence a future reader will use to reason about when notes appear.

---

To be clear about weight: items 1–3 are the ones I'd want addressed before merge, and 1 and 2 are two lines of YAML between them. 4–6 are judgement calls. The prose in CONTRIBUTING is unusually good — the `allow_repositories` gotcha and the explicit "this will never block your PR" framing are exactly the things that make optional tooling actually get adopted.

thomaspinder:
Thanks. This caught two bugs. All of 1, 2, 4, 5, 6 and the nits are fixed in ab480f8.

1. concurrency. Now group: ${{ github.workflow }} with cancel-in-progress: false. The old per-PR-per action key only serialised a PR against itself, and split closed from synchronize into separate groups.

2. timeout-minutes: 15.

4, installer. Replaced curl | bash with a version-pinned binary plus sha256sum --check --strict against the release SHA256SUMS. CI needs only the binary, not the agent hooks, shell-rc edits, or daemon. Verified including a negative control.

5, 6, nits. Count corrected to 20 in the commit message. Renamed to .yml. Documented that pre-workflow history is not consolidated, the org-glob enrolment trade-off, and the "when notes appear" wording.

3. The schedule: backstop does not work as proposed. ci github run calls get_github_ci_context(), which returns Ok(None) without a PR payload, prints "No GitHub CI context found; nothing to do", and exits 0. On a schedule it is a permanently green no-op. A working sweep needs gh api to enumerate merged PRs and drive git-ai ci local merge, so I filed that separately rather than bolting it on: #393 

Instead, continue-on-error is now scoped to forks only:

```yaml
continue-on-error: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
```

Forks get a read-only token unconditionally, so that failure carries no information. Internal PRs go red. Used full_name != github.repository rather than head.repo.fork so a deleted fork (null head.repo) fails toward tolerant.

6, mechanism. Pre-review publication comes from git-ai's push hook syncing notes on git push, not from the synchronize trigger; the CI sync path exits at SkippedFastForward on ordinary pushes without pushing. Documented with that mechanism.

Not done here: backfilling the 20 orphans, which needs the same tooling as the sweep. Covered in the linked issue, with a caveat in CONTRIBUTING.md in the meantime.

daimon-pymclabs:
Re-reviewed `ab480f8`. I verified the fixes rather than taking them on trust, and your rebuttal on item 3 is correct — I was wrong to propose the `schedule:` backstop as a one-line addition.

**What I verified independently:**

- **Checksum.** Downloaded `git-ai-linux-x64` from the `v1.6.21` release: `8465f820...4b17d3d7`, matching both the pinned `GIT_AI_SHA256` and the release's own `SHA256SUMS` entry. The asset name is correct for that tag.
- **The standalone binary is sufficient.** Ran it with no installer, no daemon, no hooks: `git-ai --version` → `1.6.21`, and `git-ai ci github run` outside a PR context → `No GitHub CI context found; nothing to do`, **exit 0**. So your `get_github_ci_context()` analysis is exactly right — a `schedule:` trigger would have been a permanently green no-op, which is the worst possible outcome for a job whose purpose is to catch silent failures. Filing #393 instead of bolting it on was the right call.
- **`ci local` supports the sweep.** `git-ai ci local merge --merge-commit-sha --base-ref --head-ref --head-sha --base-sha [--fork-clone-url]` and `ci local sync` both exist with the flags #393 needs.

Three things remain. The first is a consequence of the fix I recommended, so it's mine to raise.

### 1. The global concurrency group moves the loss mode rather than removing it — and it's now load-bearing on #393

With `group: ${{ github.workflow }}` and `cancel-in-progress: false`, GitHub holds **at most one pending run per group**: when a new run queues behind an in-progress one, any already-pending run in that group is cancelled. So:

- run A in progress (4–7 min),
- `closed` run B for a just-merged PR queues → pending,
- any `synchronize` run C arrives from any other open PR → **B is cancelled**, C takes the pending slot.

B is the irreplaceable one. `closed` never re-fires, so that merge's consolidation is gone — the same permanent-loss outcome as the non-fast-forward race, reached by a different route. And because `synchronize` fires on every push to every open PR while `closed` fires once per merge, the frequent runs are structurally favoured to evict the rare irreplaceable ones.

I don't think this makes the change wrong: a cancelled pending run is strictly better than two concurrent unsynchronised pushes, and the window is only 4–7 minutes of overlap. But it does mean neither concurrency key is durable on its own, and the sweep in #393 stops being a nice-to-have — it becomes the only mechanism that actually guarantees eventual consistency. Worth putting it on a milestone rather than leaving it open-ended, and worth a line in the workflow comment saying so, since the current comment reads as though repo-wide serialisation closes the hole.

### 2. `git-ai` dispatches on `argv[0]`, and getting it wrong fails silently with exit 0

Found this by accident while probing. I copied the binary to `/tmp/gitai-bin` and ran it:

```
$ ./gitai-bin --version
git version 2.47.3
$ ./gitai-bin ci --help
git: 'ci' is not a git command. See 'git --help'.
```

Under any name other than `git-ai` the binary transparently proxies to `git`. Copied to `bin/git-ai` it behaves correctly (`1.6.21`). **Your workflow gets this right** — the `curl -o` target is `$RUNNER_TEMP/git-ai/git-ai`. But it's an invisible tripwire: the filename is load-bearing, nothing in the file says so, and the failure mode is `git ci` → nonzero-but-plausible output, or worse a future dispatch change that exits 0 having done nothing. A one-line assertion after the checksum makes it self-defending:

```yaml
git-ai --version | grep -qE '^[0-9]+\.' || { echo "git-ai not dispatching as git-ai (argv[0]?)"; exit 1; }
```

Cheap insurance for a job that only runs on merge, where nobody is watching the output.

### 3. The pin has no bump path and nothing will tell you it has rotted

Solving the supply-chain problem introduced a staleness one. Upstream ships fast — v1.6.18 (07-30), .19 (07-31), .20 (07-31), .21 (08-04) — while contributors' local installs come from `install.sh`, which always fetches latest. So CI and local will diverge continuously, and this repo has no `dependabot.yml` at all, and Dependabot doesn't track raw release-URL downloads even where it is configured. Nothing will ever open a bump PR.

That's tolerable — the checksum turns a stale pin into a loud failure rather than a quiet one if the URL ever moves — but it's worth deciding explicitly whether anyone owns the bump, and noting in the `env:` comment that local installs are unpinned so a note-format change between versions would show up as CI/local divergence first.

### Smaller notes

- The `continue-on-error` expression is correct for the current trigger set, but it evaluates to `'' != 'pymc-labs/pathmc'` → `true` under any non-`pull_request` event. If #393 lands as a trigger on *this* workflow rather than a separate one, the sweep would inherit fork-style silence by default. Worth a comment so the next person doesn't have to rediscover it.
- The upstream 503 ([run 30800892739](https://github.com/pymc-labs/pymc-marketing/actions/runs/30800892739)) was not a fork PR, so under the new scoping that class of transient GitHub failure now reddens an internal PR. That's the trade you're deliberately making and I agree with it, but expect the first one to look like a real bug, and the standing risk with a check that is occasionally noise is that people learn to ignore it.
- Using `full_name != github.repository` rather than `head.repo.fork` so a deleted fork fails toward tolerant is a nice touch — that's the kind of null-handling that usually gets found in production.

Nothing here blocks merge from my side. Item 1 is worth a milestone on #393 and a comment correction; 2 and 3 are one line each.

drbenvincent:
One last thing before we approve: could we preserve all queued runs, not just the active one?

`cancel-in-progress: false` prevents cancelling the running job, but GitHub still keeps only one pending run by default and replaces it when another run queues. A merged PR's `closed` run could therefore be evicted by a later `synchronize` or merge event, with no future event to retry it. Adding `queue: max` would reduce that risk, with the periodic backfill in [#393](https://github.com/pymc-labs/pathmc/issues/393) as the durable backstop.

Everything else from the previous review looks addressed.

thomaspinder:
@drbenvincent Added queue: max in 0c7ae64. Verified against the docs and live on this PR: GitHub accepts it and the run executes. Note actionlint 1.7.12 flags queue as an unexpected key; its schema predates the May 2026 addition. zizmor is clean and CI doesn't run actionlint.

PR #398, opened 2026-08-07
Fix temporal panel do dispatch

### LINKED ISSUE #394: do() without simulate_over='time' crashes on scan-compiled panel models
## Summary

Calling `do()` on a scan-compiled panel model without `simulate_over="time"` raises a shape error instead of running the cross-sectional path.

```
ValueError: Shape of data ((75,)) does not match shape of type ((None, None))
```

## Reproducer (fails on `main` @ 3f403ed)

```python
import numpy as np, pandas as pd, pathmc

rng = np.random.default_rng(0)
rows = []
for r in "ABC":
    y_prev = 4.0
    for w in range(1, 26):
        x = rng.normal()
        y = 2 + 0.5 * y_prev + 0.4 * x + rng.normal(scale=0.3)
        rows.append({"region": r, "week": w, "x": x, "y": y})
        y_prev = y

m = pathmc.model("y ~ x + lag(y)", data=pd.DataFrame(rows), panel={"unit": "region", "time": "week"})
m.fit(random_seed=42, compute_log_likelihood=False)
m.do(set={"x": 0.0}, kind="mean")   # ValueError
```

`_model.PathModel.do()` dispatches to `run_do_panel_unified` only when `simulate_over == "time"`; otherwise it falls through to `_run_do` -> `run_do_pymc`, which flattens the intervention to `(n_rows,)` while the scan model's inputs are `(n_times, n_units)`.

## Impact

Two options, both fine, but the current behaviour is neither:

1. If a cross-sectional `do()` is meaningful for a scan model, reshape the intervention to the panel grid on that path.
2. If it is not meaningful, raise a clear error at dispatch ("this model was compiled with a temporal scan; pass simulate_over='time'") rather than surfacing a pytensor shape mismatch.

## Context

Found while extending test coverage for #387. Ben asked there for a flagged-model regression test on the cross-sectional `run_do_pymc` call sites; that coverage is not currently reachable, because the observed-carry flag only exists on scan-compiled panel models and those cannot complete the cross-sectional path. The `_forced_generative_carry` guard in `run_do_pymc` is therefore defensive-only until this is resolved.

drbenvincent:
## Summary
- reject cross-sectional `do()` dispatch for scan-compiled temporal panel models
- direct callers to `simulate_over="time"` for valid forward propagation
- add regression coverage for the clear error

## Test plan
- [x] `uv run pytest tests/test_panel_do.py -x -v`
- [x] `make lint`

Closes #394

### DIFF
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -1203,7 +1203,8 @@ class PathModel:
             If ``simulate_over="time"`` without panel, or if an intervention
             on an ``hsgp()`` input falls outside the basis boundary
             ``[mid - L, mid + L]`` frozen from the fitted data (beyond it
-            the basis aliases instead of extrapolating).
+            the basis aliases instead of extrapolating), or if a temporal
+            scan model is simulated without ``simulate_over="time"``.
         """
         idata = self._require_fitted("do")
         assert self._data is not None
@@ -1248,6 +1249,12 @@ class PathModel:
                     families=self._families,
                 )
             # Non-scan panel models fall through to the cross-sectional path.
+
+        if getattr(self._gen_model, "_pathmc_panel_scan", None) is not None:
+            raise ValueError(
+                "This model has temporal dependencies (lag() or adstock()). "
+                "Pass simulate_over='time' to do()."
+            )
 
         return self._run_do(set, kind)
 
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -1186,8 +1186,9 @@ class PathModel:
             ``"mean"`` for deterministic propagation via mu Deterministics,
             ``"predictive"`` to include residual noise.
         simulate_over : str | None
-            ``"time"`` to activate time-forward panel simulation.
-            Requires the model to have been fitted with ``panel=``.
+            ``"time"`` to activate time-forward panel simulation. Required
+            for models with ``lag()`` or ``adstock()``, which also require
+            the model to have been fitted with ``panel=``.
 
         Returns
         -------
@@ -1210,6 +1211,15 @@ class PathModel:
         assert self._data is not None
         assert self._gen_model is not None
 
+        scan_info = getattr(self._gen_model, "_pathmc_panel_scan", None)
+        if scan_info is not None and simulate_over != "time":
+            raise ValueError(
+                "This model has temporal dependencies (lag() or adstock()), so "
+                "interventions must be simulated forward in time. Pass "
+                "simulate_over='time' directly to do(), or as a keyword to "
+                "ate(), cate(), prob(), or sensitivity()."
+            )
+
         if set:
             _reject_hsgp_out_of_bounds(self._spec, self._data, set)
             _warn_extrapolation(self._data, set)
@@ -1221,7 +1231,6 @@ class PathModel:
                     "Pass panel={...} to model()."
                 )
 
-            scan_info = getattr(self._gen_model, "_pathmc_panel_scan", None)
             n_times = (
                 scan_info.n_times
                 if scan_info is not None
@@ -1249,12 +1258,6 @@ class PathModel:
                     families=self._families,
                 )
             # Non-scan panel models fall through to the cross-sectional path.
-
-        if getattr(self._gen_model, "_pathmc_panel_scan", None) is not None:
-            raise ValueError(
-                "This model has temporal dependencies (lag() or adstock()). "
-                "Pass simulate_over='time' to do()."
-            )
 
         return self._run_do(set, kind)
 
--- a/pathmc/simulate.py
+++ b/pathmc/simulate.py
@@ -79,6 +79,9 @@ def _forced_generative_carry(model: pm.Model) -> Iterator[None]:
     ``pathmc._model._observed_carry`` does for ``predict()``, removes that
     dependency. A no-op on models without the flag (anything but a
     scan-compiled panel model with a lagged endogenous term).
+
+    The cross-sectional call sites retain this guard defensively; current
+    scan-compiled models are dispatched to ``run_do_panel_unified`` instead.
     """
     if _OBSERVED_CARRY_FLAG not in model.named_vars:
         yield
--- a/tests/test_panel_do.py
+++ b/tests/test_panel_do.py
@@ -87,6 +87,13 @@ class TestPanelDoAPI:
         with pytest.raises(ValueError, match="panel"):
             model.do(set={"X": 1.0}, simulate_over="time")
 
+    def test_scan_panel_requires_time_simulation(self, panel_lag_model):
+        """Temporal scan models require time-forward simulation."""
+        with pytest.raises(
+            ValueError, match=r"temporal dependencies.*simulate_over='time'"
+        ):
+            panel_lag_model.do(set={"spend": 10.0}, kind="mean")
+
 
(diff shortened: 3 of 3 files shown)


daimon-pymclabs:
## Review (pre-mortem focus)

Approach is right: #394 offered two options and option 2 (fail fast at dispatch) is the correct one here, since a cross-sectional `do()` on a `(n_times, n_units)` scan graph has no well-defined meaning. Change is small and the guard is correctly placed after the `simulate_over == "time"` early-return, so scan models can still take the time path and non-scan panel models still fall through to cross-sectional.

I checked the rest of the surface for the same crash and for collateral damage. Summary of what I verified, then the nits.

### Verified safe

- **No docs breakage.** Every `do()` in the docs that sits on a panel + `lag()`/`adstock()` model already passes `simulate_over="time"` (`mmm.qmd` 266/401/455, `mmm_awareness.qmd` 732ff, `panel_data.qmd` 599, `17-panel-interventions.qmd` 304). The later bare `model.ate(...)` / `model.prob(...)` calls in `mmm.qmd` (1219, 1228, 1237) are on the `sales ~ tv + digital + trend` panel model defined at line 1035, which has no temporal terms, so `_has_temporal_deps` is False and it is not scan-compiled. Those keep working.
- **No other entry point leaks the raw pytensor shape error.** `att()` and `refute_placebo()` already raise `NotImplementedError` for panel models (`_model.py` att docstring, `refute.py:739`), and `run_counterfactual` already raises `NotImplementedError` for `lag`/`transform` slots (`simulate.py:851`). So `do()` plus the estimand wrappers that forward to it were the only exposed path. The guard is complete as written.
- `run_do_panel_unified` supports both `kind="mean"` and `kind="predictive"` (`simulate.py:1151`), so redirecting users to `simulate_over="time"` is not a dead end for `prob()`, which defaults to predictive.

### Worth addressing

**1. The error message names the wrong entry point for most callers.** `ate()`, `cate()`, `prob()` and `sensitivity()` all forward `**do_kwargs` into `do()`, so a user who calls `model.ate("y", "x")` on a lag panel model now gets *"Pass simulate_over='time' to do()"* while never having called `do()`. The fix on their side is `model.ate("y", "x", simulate_over="time")`, which the message does not suggest. Something like:

```
"This model has temporal dependencies (lag() or adstock()), so "
"interventions must be simulated forward in time. Pass "
"simulate_over='time' — do() takes it directly; ate(), cate(), "
"prob() and sensitivity() forward it as a keyword."
```

**2. The guard fires after `_warn_extrapolation`.** Lines 1214-1216 run `_reject_hsgp_out_of_bounds` and `_warn_extrapolation` on `set` before dispatch, so a scan-model user with an out-of-range intervention gets a spurious extrapolation `UserWarning` immediately before the hard error telling them the call was invalid anyway. (The hsgp reject is moot here — `compile.py:394` routes panel + hsgp down a different branch before scan, so scan models never carry hsgp.) Moving the new check above the `if set:` block would keep the failure clean.

**3. Duplicate `getattr`.** `scan_info = getattr(self._gen_model, "_pathmc_panel_scan", None)` is already computed at line 1223 inside the time branch. Hoisting one local above `if simulate_over == "time":` and reusing it in both places would avoid the two expressions drifting apart later.

**4. `simulate_over` param docstring still reads as optional.** Only the `Raises` section was updated. The parameter description still says `"time"` *activates* time-forward simulation, which now understates it: for models with `lag()`/`adstock()` it is required, not an opt-in. Worth one clause there, since that is the docstring a user reads before hitting the error.

**5. Knock-on for #387 that should be recorded somewhere.** #394 flagged that the `_forced_generative_carry` guard on the cross-sectional `run_do_pymc` call sites (`simulate.py:1033`, `simulate.py:1095`) is defensive-only, because `_use_observed_carry` only exists on scan-compiled models and those could not complete the cross-sectional path. Choosing option 2 makes that permanent: those two call sites are now unreachable for flagged models by construction, so the regression test asked for in #387 can never be written. That is a fine outcome, but it should be explicit rather than left as an open thread — either a one-line comment at those call sites pointing here, or a note on #327/#387 closing that sub-item.

### Tests

The added test is good and matches the message tightly. Two cheap additions I would want as regression armour:

- A negative control: a panel model *without* `lag()`/`adstock()` still completes the cross-sectional `do()`. That is what stops a future refactor from broadening this check to all panel models, which would silently break `mmm.qmd`'s line-1035 model and the `ate()` calls that follow it.
- One parametrised case over `ate` / `prob` asserting the same error surfaces through the wrappers, since that is how most users will actually hit it (and it pins down whatever wording is chosen in point 1).

None of this is blocking; 1 and 5 are the ones I would not skip.

drbenvincent:
Addressed the review feedback in 27a1c75 (rebased onto current `main` first):

- Hoisted scan detection ahead of intervention validation, so invalid default calls fail cleanly without extrapolation warnings.
- Expanded the error and `simulate_over` docs to explain that `ate()`, `cate()`, `prob()`, and `sensitivity()` accept `simulate_over="time"` as a forwarded keyword.
- Added coverage for `ate()` and `prob()` error propagation, a no-warning out-of-range scan intervention, and preserved cross-sectional `do()` for non-temporal panels.
- Documented that the cross-sectional carry guard remains defensive because current scan models dispatch to `run_do_panel_unified`.

Verified with `uv run pytest tests/test_panel_do.py -x -v` and `make lint`.

daimon-pymclabs:
## Follow-up: happy to approve

All five points are addressed, and two of them better than I suggested.

- **Message** now names the wrapper entry points explicitly, so `model.ate("y", "x")` on a lag panel model tells the user exactly which keyword to add and where.
- **Guard moved above `if set:`**, so the extrapolation warning can no longer fire ahead of the hard error. Bumping the fixture intervention from `10.0` to `100.0` in `test_scan_panel_requires_time_simulation` and wrapping it in `simplefilter("error")` is the right way to pin this — the value is now genuinely out of range, so the test fails if the guard ever drifts back below `_warn_extrapolation`. That is a stronger check than the ordering comment I asked for.
- **`scan_info` hoisted** once and reused in the time branch.
- **Docstring** now states `simulate_over="time"` is required for `lag()`/`adstock()` rather than merely activating.
- **`_forced_generative_carry`** carries the note about why the cross-sectional call sites keep the guard. Putting it on the function docstring rather than at each call site is cleaner than what I proposed. Nothing left dangling on the #387 thread.

I checked the new tests for the usual traps and they are sound: `DoResult.mean()` returns `float` (`simulate.py:290`), so the `np.isfinite(...)` assert is a scalar and not an ambiguous-truth array; `pooling="partial"` is a supported string form (`compile.py:734`); and the wrapper test's regex matches the message as concatenated, with no newlines to trip `.`. The negative control is the piece I most wanted — it is what stops a later refactor from broadening the check to all panel models and silently breaking the non-temporal panel model in `mmm.qmd`.

One small note, not a change request: the guard keys on `simulate_over != "time"`, so a scan model passed a typo like `simulate_over="Time"` gets the temporal error rather than an unknown-value error. That reads fine here. The pre-existing gap is on the other side — a *non-scan* model with a typo'd `simulate_over` still falls through to the cross-sectional path silently. Unchanged by this PR and worth its own issue if it bothers anyone.

LGTM. Approving once the `test` job goes green — it is still pending as of this comment, and it is the only check that exercises any of this.

PR #400, opened 2026-08-07
Fix PyPI README images with absolute GitHub raw URLs

### LINKED ISSUE #302: Fix PyPI project description images (use absolute README URLs)
## Summary

PyPI renders `README.md` as the project description but cannot resolve **relative** image paths. Logo and contributors images are missing on https://pypi.org/project/pathmc/.

## Problem

README currently uses repo-relative paths:

- `docs/assets/logo_light.png` (and dark variant via `<picture>`)
- `docs/assets/contributors.svg`

PyPI requires **absolute URLs** (e.g. `raw.githubusercontent.com`).

## Work

- [ ] Replace README image `src` values with absolute URLs pointing at `main` on GitHub (or another stable CDN).
- [ ] Confirm `<picture>` / dark-mode logo still renders acceptably on PyPI (PyPI's markdown renderer may strip or ignore some HTML — verify after change).
- [ ] Ship in the next PyPI release (README is baked into the sdist/wheel at publish time).

## Notes

Enabling the docs site (#239) does **not** fix this — PyPI does not proxy images from `pymc-labs.github.io`.

## Acceptance criteria

- Logo and contributors graphic visible on the PyPI project page after the next release.

drbenvincent:
## Summary
- Point README logo and contributors images at absolute `raw.githubusercontent.com` URLs on `main`.
- Fixes broken images on the PyPI project page, which cannot resolve repo-relative paths like `docs/assets/...`.

Closes #302

## Notes
- PyPI's readme renderer keeps the `` fallback inside `<picture>` but strips `<source>`, so the light logo shows on PyPI while GitHub still gets dark-mode switching.
- Images will appear on PyPI after the next release (README is baked into the sdist at publish time).

## Test plan
- [x] Verified `raw.githubusercontent.com/pymc-labs/pathmc/main/docs/assets/{logo_light.png,logo_dark.png,contributors.svg}` return HTTP 200.
- [x] Rendered README with `readme_renderer`; `` tags with absolute URLs survive sanitization.
- [ ] After merge and release, confirm logo and contributors graphic render on https://pypi.org/project/pathmc/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,3 +148,7 @@ git add _freeze/
 This wipes `_freeze/`, re-runs all example and user-guide notebooks except the homepage (which needs a separate build step because of an upstream path-mapping quirk), copies the refreshed homepage cache, and prints a reminder to commit `_freeze/`. Expect this to take a long time: many example notebooks run MCMC sampling.
 
 See the "Building the docs" section of [AGENTS.md](https://github.com/pymc-labs/pathmc/blob/main/AGENTS.md) for freeze-cache details and caveats.
+
+## README assets on PyPI
+
+The PyPI project page renders `README.md` and loads its images from absolute `raw.githubusercontent.com/.../main/docs/assets/...` URLs at **view** time, not from the sdist. Do not rename or relocate `docs/assets/logo_light.png`, `logo_dark.png`, or `contributors.svg` without updating `README.md`; if those paths disappear from `main`, images break on PyPI for every previously published release. PyPI currently has no dark-mode toggle, so the README uses a `<picture>` block for GitHub (light/dark logos) with a light-background fallback for PyPI.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # pathmc
 
+<!-- PyPI resolves README images from absolute raw.githubusercontent.com URLs below at view time; do not move or rename docs/assets/logo_*.png or contributors.svg without updating README. -->
 <p align="center">
   <picture>
     <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/pymc-labs/pathmc/main/docs/assets/logo_dark.png">
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 <p align="center">
   <picture>
-    <source media="(prefers-color-scheme: dark)" srcset="docs/assets/logo_dark.png">
-    <img src="docs/assets/logo_light.png" alt="pathmc logo" width="75%">
+    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/pymc-labs/pathmc/main/docs/assets/logo_dark.png">
+    <img src="https://raw.githubusercontent.com/pymc-labs/pathmc/main/docs/assets/logo_light.png" alt="pathmc logo" width="75%">
   </picture>
 </p>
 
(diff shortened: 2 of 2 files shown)


daimon-pymclabs:
## Review: looks correct, ship it

I reproduced the PyPI render path locally and the claims in the description hold up.

**Verified**

- `readme_renderer.markdown.render()` on this branch's README: `<source>` is stripped, the `` with the absolute URL survives (`width="75%"` and the wrapping `<a href=".../graphs/contributors">` survive too). So PyPI gets the light logo plus a clickable contributors graphic.
- All three raw URLs return 200 with correct content types (`image/png`, `image/png`, `image/svg+xml`) — the SVG content type matters, since PyPI proxies description images through camo and a `text/plain` response would not render.
- `docs/assets/contributors.svg` is fully self-contained: every avatar is a base64 `data:` URI, no external `<image href>`. That is what makes it safe to render inside an `` on PyPI (external refs inside an SVG loaded via `` are blocked by the browser regardless of origin). The generator workflow already enforces this, so it stays true on regeneration.
- No docs impact: `great-docs.yml` sets `homepage: user_guide`, so the README is not the docs landing page. No link checker in `.pre-commit-config.yaml` or any workflow that these absolute URLs would newly hit.

**Two things worth noting, neither blocking**

1. *`main` is mutable, and every past release dereferences it at view time.* The README is baked into each sdist, but the image URLs resolve when someone loads the page. If `docs/assets/logo_light.png` or `contributors.svg` is ever renamed or moved, the PyPI pages for **all** previously published releases silently lose their images, retroactively and unfixably. Given there is active docs restructuring going on, that path is now load-bearing in a non-obvious way. Cheapest mitigation is a one-line comment in the README next to the `<picture>` block (or a note in `CONTRIBUTING.md`) saying these two paths are referenced by absolute URL from published PyPI pages and must not be moved. Pinning to a release tag instead would fix immutability but means the GitHub `main` README shows a stale logo, which is probably the worse trade.

2. *The `<picture>` fallback is the dark-ink logo.* `logo_light.png` is 89% transparent with mean opaque RGB ≈ (38, 46, 54), i.e. near-black artwork on transparency. Since PyPI drops the `<source>`, that is what every PyPI visitor gets. I checked the live pathmc project page and its stylesheet: no `prefers-color-scheme` rules, no theme toggle, so PyPI is light-only today and this renders fine. It only becomes a problem if PyPI ships dark mode later, at which point the logo goes nearly invisible there. Not worth acting on now, just flagging that the fallback choice has that latent property.

**One expectation to set:** since 0.2.0 shipped this morning, the images will not appear on PyPI until 0.2.1. The final unchecked test-plan box is genuinely blocked until then.

drbenvincent:
Addressed the two non-blocking review notes:

1. **Mutable `main` paths** — Added an HTML comment above the `<picture>` block in `README.md` and a short "README assets on PyPI" section in `CONTRIBUTING.md` documenting that `docs/assets/logo_*.png` and `contributors.svg` are dereferenced from `main` at PyPI view time and must not be moved/renamed without updating README.

2. **PyPI dark mode / fallback logo** — Left as-is: `<picture>` keeps GitHub README dark-mode switching via `logo_dark.png`, and PyPI gets the light-background fallback (`logo_light.png`, dark ink on transparency). CONTRIBUTING now notes PyPI has no theme toggle today; we can revisit if that changes.

No change to pinning URLs to release tags — agreed that stale logos on the GitHub `main` README would be the worse trade.

PR #401, opened 2026-08-07
Confirm counterfactual() performs real abduction, add multi-hop oracle (towards #327)

daimon-pymclabs:
## Sub-component

Item 6 of #327: counterfactual semantics — does `counterfactual()` perform
real Pearl abduction-action-prediction (infer the unit's exogenous noise
from its observed data, then intervene), or does it just re-run the
intervention with the noise reset?

## Verdict: real abduction is already implemented — confirmed, no fix needed

Code inspection of `run_counterfactual()` in `pathmc/simulate.py` (added in
#368) shows it correctly implements Pearl's three-step procedure:

1. **Abduction**: for each endogenous variable, `u = evidence[var] -
   linear_predictor(evidence)` — the noise is inferred from the unit's
   *observed* values of its parents (not the population mean).
2. **Action**: `do` variables are pinned to their intervened value.
3. **Prediction**: downstream variables are recomputed using the
   *propagated/predicted* parent values (not the observed ones) plus each
   variable's abducted noise, walking `graph_info.topological_order`.

This is evidenced by the existing test fixture in `tests/test_counterfactual.py`
(`test_joe_counterfactual_matches_analytical`, `test_matches_manual_implementation`,
`test_differs_from_population_do`), which already compares `counterfactual()`
against a hand-worked linear-Gaussian oracle and shows it differs from a plain
`do()`.

### What this PR adds

The existing Joe/encouragement fixture only exercises **one** downstream hop
(`do(H)` feeds directly into `Y`), which can't distinguish two implementations:

- (a) abduct once, then plug the intervened value straight into every
  downstream equation using **observed** values for anything further away, vs.
- (b) abduct once, then **propagate** the counterfactual (not observed) value
  through every downstream hop.

Only (b) is correct abduction-action-prediction; (a) would silently fail to
carry an intervention's effect past the first hop — a "finite, plausible,
wrong answer" of exactly the kind #327 worries about.

I added an independent, hand-derived 3-hop chain oracle
(`X -> M -> Y -> Z`, `TestCounterfactualChainAbduction` in
`tests/test_counterfactual.py`) where `do(M)` must reach `Z` through a
*propagated* `Y_cf` that differs from the observed `Y`. I verified this test
has real discriminating power by temporarily patching `run_counterfactual` to
feed observed (evidence) values instead of predicted ones into downstream
equations — the new tests failed as expected (`Z` came back unchanged at its
observed value, ignoring the intervention on `M` entirely), then reverted the
patch and confirmed the tests pass against the real implementation.

Also added a guard test (`test_pinning_took_effect`) that the
`idata["posterior"]` bracket-path pinning used by the fixture actually took
effect, per the known `model._idata.posterior = ds` vs `idata["posterior"]`
footgun.

No production code changes — this PR is test-only, since the implementation
was already correct.

## Test plan

- [x] `tests/test_counterfactual.py` — 23 passed (was 15; added 8), all
      deterministic (pinned posterior draws, no MCMC via `mock_pymc_sample_module`).
- [x] Ran the broader suite: `pytest -q -m "not slow"` — 1203 passed, 1 skipped,
      6 failed. The 6 failures are pre-existing and unrelated
      (`ModuleNotFoundError: No module named 'pyarrow'` in polars-backend
      tests, `test_falsify.py`/`test_narwhals_backends.py`); confirmed
      pre-existing by reproducing on a clean stash of this branch before my
      change. `-m "slow"` (MCMC) tests were deselected — not run.
- [x] `prek run --all-files` — clean (ruff, ruff-format, mypy, license header).

### DIFF
--- a/tests/test_counterfactual.py
+++ b/tests/test_counterfactual.py
@@ -256,6 +256,128 @@ class TestCounterfactual:
         )
         assert abs(result.mean("Y") - 3.6) < 1e-10
 
+
+class TestCounterfactualChainAbduction:
+    """Independent hand-derived oracle for a 3-hop chain X -> M -> Y -> Z.
+
+    Item 6 of #327 asks whether ``counterfactual()`` performs real
+    abduction-action-prediction, or just re-runs the intervention with the
+    noise terms reset. The Joe/encouragement fixture above only exercises a
+    single downstream hop (do(H) -> Y directly), which can't distinguish
+    "abduct once, then plug the intervened value straight into every
+    downstream equation" from "abduct once, then *propagate* the
+    counterfactual (not the observed) value through every downstream hop".
+
+    This chain does distinguish them: Z's equation depends on Y, and Y's
+    *observed* value (0.5) differs from Y's *counterfactual* value under
+    do(M=3.0) (2.12). If ``run_counterfactual`` fed the observed Y into Z's
+    prediction step instead of the counterfactual Y, Z would come out at the
+    observed value (2.0) regardless of the intervention -- the intervention
+    would never reach Z at all. The analytic oracle below is hand-derived
+    from Pearl's three steps:
+
+        abduction:  u_M = M_obs - a*X_obs
+                    u_Y = Y_obs - b*M_obs
+                    u_Z = Z_obs - c*Y_obs
+        action:     M := 3.0 (do)
+        prediction: M_cf = 3.0
+                    Y_cf = b*M_cf + u_Y
+                    Z_cf = c*Y_cf + u_Z
+    """
+
+    TRUE_A = 0.6  # X -> M
+    TRUE_B = 0.9  # M -> Y
+    TRUE_C = 1.1  # Y -> Z
+
+    EVIDENCE = {"X": 1.0, "M": 1.2, "Y": 0.5, "Z": 2.0}
+    DO_M = 3.0
+
+    # Hand-derived oracle values (see docstring above).
+    U_M = EVIDENCE["M"] - TRUE_A * EVIDENCE["X"]
+    U_Y = EVIDENCE["Y"] - TRUE_B * EVIDENCE["M"]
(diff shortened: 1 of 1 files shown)


drbenvincent:
## Defect-focused review

**Verdict: no blocking defects.** Test-only PR; implementation already correct.

### What was reviewed

- Diff vs `main`: single file `tests/test_counterfactual.py` (+122 lines)
- `run_counterfactual()` in `pathmc/simulate.py` — abduction uses factual parent values (`u = evidence - lp(factual)`), prediction walks topological order with **propagated** parent values (`predicted[parent]`), not observed evidence. Matches Pearl abduction → action → prediction.
- New `TestCounterfactualChainAbduction` (X → M → Y → Z): hand-derived oracle checked manually (e.g. `Y_cf = 2.12`, `Z_cf = 3.782` under `do(M=3.0)`). The negative control (`Z_IF_ABDUCTION_SKIPPED = 2.0`) would catch an implementation that feeds observed Y into Z instead of propagated `Y_cf`.
- `test_pinning_took_effect` — good guard for the `idata["posterior"]` bracket-path vs `_idata.posterior` attribute footgun.
- `test_latent_model_raises` reorganized into `TestCounterfactualLatent` — fine.

### Local verification

```
uv run pytest tests/test_counterfactual.py -q  → 23 passed
```

Merged `origin/main` locally (branch was 6 commits behind); merge is clean and all 23 tests still pass.

### Non-blocking notes

1. **Rebase/merge `main` before merge** — remote branch is behind `main` by 6 commits (`cd0a329`..). No conflicts expected; verified locally.
2. **Optional follow-up** — the existing `encouragement_model` fixture uses the same posterior pinning pattern but lacks an explicit pinning guard test. Not required for this PR.

### Recommendation

Approve once rebased onto `main`. Keep as draft until the #327 sub-item series is ready to land together (per current workflow).

drbenvincent (approved):
No blocking defects. Test-only PR with a discriminating 3-hop abduction oracle; implementation verified correct. Please rebase onto main before merge.

PR #390, opened 2026-08-07
Add d-separation oracle tests for identification (towards #327)

daimon-pymclabs:
Towards #327, sub-component **4** (identification oracle).

Tests only, no source changes. Adds `tests/test_identification_oracle.py`: named canonical DAGs with hand-derived verdicts (fork, chain, collider, M-bias, bow pattern both latent and observed, joint-confounder, mediator-exclusion), plus a property-based sweep that cross-checks `adjustment_sets`/`is_identifiable` against an independent `networkx.is_d_separator` oracle on the graph with the treatment's outgoing edges cut.

**Result: no bug found in `pathmc/identify.py`.** The initial 7 failures all traced to the oracle helper itself (cutting edges *into* treatment instead of out of it, a labeling bug, and a cyclic random fixture). Once fixed, pathmc's backdoor logic matched the oracle everywhere.

**Tests:** 18 passed / 2 skipped in the new file. The 2 skips are randomized DAGs that happen not to contain both X and Y, skipped by design. 105 passed across the identification, DAG, and graph suites.

Incidental finding, left alone: `pathmc/identify.py:98-106` builds a `mutilated` graph that is never passed into the blocking check, so `mutilated_dag` is dead code. `collider_warnings` and `frontdoor_identifiable` are not oracle-checked here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #388, opened 2026-08-07
Validate panel shape at construction (towards #327)

daimon-pymclabs:
Towards #327, sub-component **3** (panel shape validation).

Adds `_validate_panel_shape()` to `pathmc/panel.py`, wired into `build_panel_info()` (the single construction point for `PanelInfo`). Checks, each raising `ValueError`: null/NaN unit or time identifiers, empty unit column, duplicate `(unit, time)` rows, unbalanced per-unit counts, and non-rectangular panels where per-unit counts match but timepoint sets differ. Order-independent, consistent with the existing sort-then-reshape behavior.

The rectangularity checks (duplicate rows, balance, timepoint-set matching) are scoped to the temporal/scan compilation path via `require_rectangular` (gated on `_has_temporal_deps()`, the same predicate `compile_to_pymc()` uses). Non-temporal panel models take the row-wise compiler, which indexes rows by unit without reshaping and legitimately supports unbalanced panels; only the empty-panel and null-identifier checks apply there.

Without this, a malformed panel reaching the scan path would silently mis-reshape via `n_times = len(data) // n_units` into a wrong-but-plausible array.

**Tests:** 16 parameterised tests in `tests/test_panel_shape_validation.py`, ~1s. Covers the reshape path (via `lag()`) and the row-wise compatibility path (unbalanced/duplicate rows accepted), the `n_units > 1, n_times = 1` edge case with two distinct units, and null/NaN unit and time identifiers on both compilation paths.

**Verification:** 37 passed across `tests/test_panel_shape_validation.py`, `tests/test_panel.py`, `tests/test_compile.py`. `prek run --all-files` clean.

Out of scope and left alone: sub-component 9 (out-of-sample shape mismatch via `pm.set_data`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #402, opened 2026-08-07
Non-Gaussian mediator validation + DSL parse golden tests (towards #327)

daimon-pymclabs:
## Summary

Addresses two independent sub-components of #327's checklist: **item 8** (non-Gaussian mediators in the scan) and **item 11** (DSL parsing golden tests). Kept as two separate commits, one per concern, with dedicated test files. Other checklist items (#388, #389, #390, #391; #387 merged) are out of scope here.

### Item 8 — non-Gaussian mediators (`pathmc/compile.py`)

Audited `_validate_scan_non_gaussian_intermediaries`, the guard that rejects scan-compiled panel models where a Bernoulli/Poisson/NegativeBinomial endogenous node is used as a *predictor* elsewhere (the scan step function passes that node's probability/rate forward, never a sample).

**Coverage verified** with parametrized golden tests in `tests/test_scan_non_gaussian_mediators.py` enumerating every topology the validator must catch: plain term, self-lag, cross-equation lag, single transform, nested transform, interaction term, random slope, and transitive chains (M feeds N feeds Y, each individually flagged) — all raise. Also verified it correctly stays silent for safe cases: terminal scan outcome, and no temporal deps at all (cross-sectional compiler, mean-leak concern doesn't apply).

**Gap found and closed**: the validator explicitly excludes `latent` variables (`if reg.lhs not in latent`), so a latent node assigned a discrete family (e.g. `latent=["M"], families={"M": "bernoulli"}`) was silently accepted — but latent nodes only ever get a deterministic `mu` pass-through or a `latent_normal` stochastic innovation; the family label itself is otherwise ignored, with no sigmoid/exp link applied at all. Added `_validate_latent_families` to reject that combination outright instead of silently ignoring the family setting.

**Clip documented**: `clip(mu, -20, 20)` before `exp()` on Poisson/NegBinomial rates inside the scan is a numerical-stability guard for the recursive computation (a transient blow-up in one step would otherwise poison the whole scan's logp with inf/nan), not a modeling choice — documented inline. Added `_warn_high_rate_clip_risk`, which raises a `UserWarning` at compile time when observed counts imply a rate near the clip bound (`exp(20) ≈ 4.85e8`), so genuinely high-rate data doesn't get silently truncated without any signal.

### Item 11 — DSL parsing (`pathmc/parse.py`)

Added parse-tree golden tests in `tests/test_dsl_parse_golden.py` (no data, no PyMC, no sampling — the whole file runs in ~2s) covering:
- Operator precedence (`*` label binds one term, `:` interaction binds tighter than `+`, labels don't distribute over later terms)
- `a:b` (real interaction) vs `a*b` (coefficient label on a single term — **not** an R-style `a + b + a:b` expansion; documented as intentional, since pathmc's DSL is lavaan-inspired, not patsy-formula-inspired)
- Nested transforms in both orders (`adstock(logistic_saturation(x))` vs `logistic_saturation(adstock(x))`), triple-nested leaf resolution, `lag()` rejecting nested transforms
- `0 +` / `-1` intercept removal
- Multi-equation specs (newline- and semicolon-separated, continuation lines, mixed with `~~` and `:=`)
- Variable names colliding with transform/keyword names (`adstock`, `lag`, `hsgp` used as plain variables, transform inputs, or LHS names)

**Behavior change**: added `_reject_top_level_minus`, which now raises a `ParseError` for R/patsy-style subtraction. Previously, `y ~ x - 1`, `y ~ x + z - 1`, `y ~ -1`, and `y ~ -x` all parsed "successfully" into a term literally named e.g. `"x - 1"` (a single un-splittable string, since `_parse_regression` only splits on `+`) — that garbage variable name would then fail much later with a confusing "column not found" error at data-binding/compile time instead of a clear parse error. This is exactly the intercept-removal crash vector #327 item 11 calls out (per #316's analogous concern for `lag()`). These specs now raise immediately with a message pointing at the syntax pathmc actually supports (`0 + ...` for dropping the intercept, `value*variable` for a fixed/negative coefficient).  The same guard also rejects hyphenated plain terms and coefficient labels that previously parsed on `main` (e.g. `y ~ my-var`, `y ~ my-label*x`); use underscores instead (`my_var`, `my_label*x`). Legitimate negative fixed coefficients (`-1*x`) and negative transform kwargs (`decay=-0.5`) are explicitly preserved and tested.

## Test results

- `tests/test_scan_non_gaussian_mediators.py` + `tests/test_dsl_parse_golden.py`: **79 passed** (foreground, ~2s, no sampling).
- Broader regression check of everything `pathmc/compile.py` / `pathmc/parse.py` touch (foreground, not the full ~1500-test suite):
  - `test_parse.py`, `test_lag_syntax.py`, `test_interactions.py`, `test_hsgp_parse.py`, `test_transforms_parse.py`: **126 passed**
  - `test_compile.py`, `test_panel.py`: **21 passed**
  - `test_panel_do.py`, `test_panel_by_time.py`: **29 passed**
  - `test_panel_engines.py`, `test_panel_predict_carry.py`: **28 passed**
  - `test_do.py`, `test_do_predictive.py`, `test_do_regressions.py`, `test_do_slopes.py`, `test_causal_queries.py`, `test_exog_lag_missing_base.py`, `test_stochastic_latent.py`, `test_transforms_do.py`, `test_latent.py`: **90 passed**
  - `test_families.py`: **19 passed**
  - `test_narwhals_backends.py`: 33 passed, **9 pre-existing failures** — all `ModuleNotFoundError: No module named 'pyarrow'` on the `polars` backend parametrization, confirmed present in this sandbox regardless of branch (`pyarrow` isn't installed at all: `python -c "import pyarrow"` fails on `main` too). Unrelated to this diff.
- **The full, unscoped test suite (~1500 tests) was not run to completion** in this session — background full-suite attempts were unreliable under this sandbox's resource contention (multiple concurrent pytest processes). The targeted suites above were chosen to cover every test file that imports or exercises `compile.py`'s scan path and `parse.py`, and all pass.
- `prek run --all-files`: clean (ruff, ruff-format, mypy, license header, whitespace/EOF/yaml/toml checks all pass).

## Findings summary

| Item | Finding | Action |
|---|---|---|
| 8 | Validator already covers all predictor topologies (plain, lag, transform, nested transform, interaction, slope, transitive chains) | Golden tests added, no fix needed there |
| 8 | Latent + discrete family silently ignored the family, using raw `mu` | Fixed: `_validate_latent_families` now raises |
| 8 | `clip(mu, -20, 20)` undocumented, silent for high-rate data | Documented inline; `_warn_high_rate_clip_risk` warns at compile time |
| 11 | `-1`/`x - 1` (R-style subtraction) silently parsed to a garbage variable name | Fixed: `_reject_top_level_minus` raises a `ParseError` (behavior change, see above) |
| 11 | Hyphenated plain terms / coefficient labels (`my-var`, `my-label*x`) parsed on `main` | Rejected by `_reject_top_level_minus`; use underscores (golden tests added) |
| 11 | `a*b` vs `a:b`, nested transforms, name collisions | Confirmed correct via golden tests; no bugs found |


drbenvincent (commented):
## Review summary (defect-focused)

**Verdict:** Solid work on #327 items 8 and 11 — good golden tests, clear compile-time guards, and the latent+discrete-family silent-ignore gap is a real bug worth closing. One parse regression and a sync item before this is merge-ready.

### What looks good

- **`_validate_latent_families`**: Correctly rejects latent nodes with discrete families instead of silently compiling raw `mu` pass-through. Tests cover non-panel and panel-scan paths; `latent_normal` / default gaussian still allowed.
- **Non-Gaussian mediator scan validator**: Parametrized golden tests enumerate plain/lag/transform/nested-transform/interaction/slope/transitive-chain topologies; safe cases (terminal outcome, no temporal deps, non-panel) correctly stay silent.
- **`_SCAN_MU_CLIP_BOUND` + `_warn_high_rate_clip_risk`**: Inline documentation of the clip rationale is helpful; compile-time warning for high-rate Poisson/NegBin outcomes is a sensible guard against silent truncation.
- **`_reject_top_level_minus`**: Fixes the real crash vector where `y ~ x - 1` parsed to a bogus variable name. Golden parse tests are fast (~2s) and cover precedence, nested transforms, multi-equation specs, and name collisions well.
- **CI**: All checks green on the PR head.

### Defect — scientific notation fixed coefficients regressed

`_reject_top_level_minus` rejects any top-level `-` except a leading `-` before `*` (i.e. `-1*x`). That breaks **scientific-notation fixed coefficients with negative exponents**, which parsed fine on `main`:

```python
parse_spec("y ~ 1e-5*x")  # main: fixed_value=1e-05
parse_spec("y ~ 1e-5*x")  # this PR: ParseError (Unsupported '-')
```

Same for `1E-5*x`. `2e+3*x` still works; `-1*x` still works.

**Suggested fix:** in the minus scan, skip `-` when it is part of a float literal (e.g. preceded by `e`/`E`, or inside a numeric prefix before `*`). Add a golden test for `1e-5*x` (and optionally `1E-5*x`).

### Sync with `main`

`327-mediators-and-dsl` is **6 commits behind `main`** at review time. Please merge/rebase `main` before final merge (local merge was clean in a test worktree).

### Recommendation

- **Do not mark ready yet** until the `1e-5*x` regression is fixed (small, localized).
- After that fix + `main` sync, this looks good to mark ready for review.

No other blocking defects found in the diff.

drbenvincent:
## Follow-up: parse regression fixed (`cb25fa8`)

Addressed the review finding from the defect-focused pass: `_reject_top_level_minus` now skips `-` when it is the exponent sign in a float literal (preceded by `e`/`E`), so `1e-5*x` and `1E-5*x` parse as fixed coefficients again — matching `main`.

Added parametrized golden tests in `test_scientific_notation_fixed_coefficient_still_allowed`. R-style subtraction cases (`x - 1`, `-x`) still raise `ParseError`.

**Verification:** `tests/test_dsl_parse_golden.py` → 39 passed locally; pre-commit hooks clean on push.

**Still outstanding before merge:** rebase onto current `main` (branch was 6 commits behind at review time).

drbenvincent (approved):
Parse regression fixed in cb25fa8 — scientific-notation fixed coefficients work again while R-style subtraction still rejected. Golden tests added. Approve after rebase onto main.

drbenvincent:
## Follow-up fixes complete (`f7976a3`)

1. **Parse regression** — already fixed in `cb25fa8`: `_reject_top_level_minus` skips `-` in scientific-notation exponents (`1e-5*x`, `1E-5*x`); golden tests added.

2. **Sync with `main`** — merged current `main` into `327-mediators-and-dsl` (clean merge, no conflicts). Branch is now up to date with `main`.

**Verification on merge tip:**
- `tests/test_dsl_parse_golden.py` + `tests/test_scan_non_gaussian_mediators.py` → **81 passed**
- `prek run --all-files` → all hooks pass

CI will re-run on the merge push. PR remains approved; mark ready when you want merge consideration.

daimon-pymclabs:
## Bugbot finding addressed (`86a7141`)

**"Minus guard false negative on e-names"** — confirmed real, not a false positive.

`_reject_top_level_minus` accepted any `-` whose preceding character was `e`/`E`, without checking that the `e` terminated an actual numeric mantissa. Any variable name ending in `e` slipped past the guard and became the garbage single term Bugbot describes.

**Fix:** a new `_is_float_exponent_sign(raw, i)` requires all three of:
- the character before the `-` is `e`/`E`;
- the text ending at that `e` is an anchored numeric mantissa (`(?:^|[^0-9A-Za-z_.])(?:\d+\.?\d*|\.\d+)[eE]$`), so the trailing `2e` of an identifier like `x2e` does not qualify;
- the character after the `-` is a digit.

**Coverage** (verified failing on this branch *before* the fix, 6 failures):

| now rejected | still parsed as a fixed coefficient |
|---|---|
| `y ~ response-1` | `y ~ 1e-5*x` |
| `y ~ income-1` | `y ~ 1E-5*x` |
| `y ~ x + rate-1` | `y ~ 2.5e-3*x` |
| `y ~ RATE-1` | `y ~ .5e-3*x` |
| `y ~ x2e-1` | `y ~ -1e-5*x` |
| `y ~ 1e-x` | `y ~ -1*x` |

**Verification:** `tests/test_dsl_parse_golden.py` + `tests/test_scan_non_gaussian_mediators.py` → **90 passed**; `ruff check` and `ruff format --check` clean on both touched files.

drbenvincent (changes requested):
## Request changes

Solid work on #327 items 8 and 11 — the latent+discrete-family guard is a real bug fix, the scan-mediator golden tests are thorough, and the parse regressions from the first review (`1e-5*x`, identifiers ending in `e`) are fixed in `cb25fa8` / `86a7141`. CI is green on the current head.

Two items before merge:

### 1. Sync with current `main` (required)

The branch is **1 commit behind** `main` (`447215e` — #403, OOS panel-shape / NaN predictor guards). Both PRs touch `pathmc/compile.py`.

I verified locally: `git merge origin/main` **auto-merges cleanly** (no conflict markers), and the combined tip passes:

- `tests/test_dsl_parse_golden.py`
- `tests/test_scan_non_gaussian_mediators.py`
- `tests/test_panel_nan_handling.py`
- `tests/test_panel_oos_shape.py`

→ **111 passed**

Please merge/rebase `main` and push so CI runs on the true merge tip. The sync done at `f7976a3` predates #403.

### 2. Document the broader parse behavior change (required)

The PR description covers R/patsy-style subtraction (`y ~ x - 1`, `y ~ -x`), but `_reject_top_level_minus` rejects **any** top-level `-` in a term. That also blocks specs that parsed fine on `main`:

```python
parse_spec("y ~ my-var")        # was: variable "my-var"
parse_spec("y ~ my-label*x")    # was: label "my-label" on x
```

This is understandable collateral from catching `response-1` / `income-1` (correctly tested as errors), and hyphenated names were never valid on the LHS anyway. Still, it is a user-visible behavior change beyond what the PR summary states.

Please add a line to the PR description (and/or changelog) noting that hyphenated plain terms and coefficient labels are no longer accepted — users should use underscores instead.

### Optional polish (non-blocking)

- `_warn_high_rate_clip_risk` uses `stacklevel=2`; `_warn_partial_pooling_intercept` uses `stacklevel=4` in the same file. Bumping to 4 would surface the warning at the `model()` call site.
- `_validate_latent_families(spec, ...)` — `spec` is unused; could drop the parameter.

Happy to re-review once `main` is synced and the parse behavior change is documented.

drbenvincent:
## Review follow-up (`c8660e4`)

Addressed both required items from the latest review:

### 1. Sync with `main` — already on branch
`0d14b48` merged current `main` (including #403) before this push. CI will run on the updated tip.

### 2. Document hyphenated-name parse behavior change
- PR description updated (behavior-change paragraph + findings table row).
- `_reject_top_level_minus` docstring now states that hyphenated plain terms and coefficient labels are rejected; use underscores.
- Golden tests added for `y ~ my-var` and `y ~ my-label*x`.

**Also applied the optional polish:**
- `_warn_high_rate_clip_risk` `stacklevel` bumped to `4` (matches `_warn_partial_pooling_intercept`).
- Unused `spec` parameter removed from `_validate_latent_families`.

**Verification:** `tests/test_dsl_parse_golden.py` + `tests/test_scan_non_gaussian_mediators.py` + `tests/test_panel_nan_handling.py` + `tests/test_panel_oos_shape.py` → **113 passed**; `make lint` clean.

daimon-pymclabs:
## Re-review of `c8660e4`

Both required items from the previous review are addressed: the branch is **0 commits behind `main`** (`0d14b48` merged #403), and the hyphenated-name parse change is now documented in `_reject_top_level_minus`'s docstring with golden tests for `my-var` / `my-label*x`. CI is green on this head (`test` passed in 32m50s, `prek` and Bugbot pass).

Two things I verified empirically on `c8660e4` that are worth resolving before merge. Neither is a regression against `main` — both are gaps in the new code.

### 1. `stacklevel=4` still points at an internal frame (optional polish that didn't land)

I instrumented `warnings.warn` and walked the frames on a real `pathmc.model(...)` call that triggers the clip warning:

```
stacklevel=1: compile.py:1485  _warn_high_rate_clip_risk
stacklevel=2: compile.py:1601  _compile_scan_panel
stacklevel=3: compile.py:463   compile_to_pymc
stacklevel=4: _model.py:293    _compile      <-- where 4 lands
stacklevel=5: _model.py:267    __init__
stacklevel=6: _model.py:1844   model
stacklevel=7: <user script>    <module>      <-- actual user call site
```

The warning is attributed to `_model.py:293`. `stacklevel=2` pointed at `_compile_scan_panel`, also internal, so the bump changes which internal frame is blamed rather than surfacing the caller — Bugbot's finding is still open.

The "matches `_warn_partial_pooling_intercept`" rationale doesn't transfer: that warning is raised one frame shallower (called from `compile_to_pymc` at line 448, not from inside `_compile_scan_panel`), so its own `stacklevel=4` resolves to `_model.py:267 in __init__` — internal as well. A single fixed constant cannot be correct for both call sites; the right value *here* is 7.

Suggest either computing it (a small helper that walks out to the first non-`pathmc` frame, which is robust to future call-chain changes and would fix the partial-pooling warning too), or reverting to 2 with a comment recording that a fixed constant can't serve both sites. Bumping to 4 buys nothing.

### 2. The minus guard has a hole inside parentheses

`_reject_top_level_minus` only scans at paren depth 0, so a `-` inside a transform's argument is never checked:

```python
y ~ lag(rate-1)      # PARSES -> variable = 'lag(rate-1)'
y ~ adstock(my-var)  # PARSES -> variable = 'my-var'
```

Both produce garbage column names that fail later at data binding with exactly the confusing "column not found" error this guard was added to prevent. It also contradicts the docstring added in this commit, which now states that `-` is never a name character in the DSL — that's only enforced at top level.

Top-level cases behave correctly (`y ~ x - 1`, `y ~ my-var`, `y ~ my-label*x` all raise `ParseError`), and `1e-5*x` still parses as a fixed coefficient, so the fix is narrow: apply the same guard to a transform's leaf input argument. Keyword values (`decay=-0.5`) are parsed separately and would be unaffected.

### Verification notes

- Targeted suites on this head: `tests/test_dsl_parse_golden.py` + `tests/test_scan_non_gaussian_mediators.py` → **98 passed**.
- Full suite on my sandbox: 1653 passed, 6 failed, 2 skipped. The 6 failures are all `test_graph_consistency.py::...[panel-lag(x)-*]`. I checked out `main` (`447215e`) and reproduced **the identical 6 failures**, so they are pre-existing and unrelated to this PR. They also pass in CI, so they look environment-specific to my sandbox rather than a real defect on `main`.
- `_validate_latent_families` has exactly one caller (`compile.py:447`), so dropping the unused `spec` parameter is safe.

### Verdict

Not ready to accept as-is. Item 2 is the one I'd treat as blocking: it's a real hole in the guard this PR introduces, and it's small to close. Item 1 is cosmetic (warning attribution only) but currently claims a fix it doesn't deliver. With those two resolved, this is merge-ready — the latent+discrete-family guard and the scan mediator tests are good work and the parse regressions from earlier rounds are all fixed.

drbenvincent:
## Review items addressed (`b727d02`)

Both items from the re-review of `c8660e4` are resolved:

### 1. Minus guard hole inside parentheses (blocking)

`_reject_top_level_minus` now applies to transform **leaf inputs** via a one-line addition in `_parse_transform_expr`:

```python
else:
    _reject_top_level_minus(input_raw)
    input_expr = input_raw
```

All four reported vectors now raise `ParseError`:

| Now rejected | Was |
|---|---|
| `y ~ lag(rate-1)` | Parsed to garbage `lag_of='rate-1'` |
| `y ~ adstock(my-var)` | Parsed to leaf `'my-var'` |
| `y ~ logistic_saturation(income-1, lam=lam)` | Parsed to leaf `'income-1'` |
| `y ~ adstock(logistic_saturation(my-var), decay=theta)` | Nested leaf slipped through |

Negative keyword values (`decay=-0.5`) are unaffected — parsed in a separate branch.

### 2. Warning stacklevel (cosmetic, fixed correctly this time)

Replaced both fixed `stacklevel=4` constants with `_user_stacklevel()`, which walks frames to the first non-`pathmc` module. This correctly attributes warnings to the user's `model()` call site regardless of internal call-chain depth:
- `_warn_high_rate_clip_risk`: 7 frames deep → now resolves to user site
- `_warn_partial_pooling_intercept`: 6 frames deep → now resolves to user site

### Verification

- `test_dsl_parse_golden.py` + `test_scan_non_gaussian_mediators.py` → **96 passed** (4 new tests)
- Broader parse/compile/panel suites → **147 passed**
- `test_families.py` + `test_panel_nan_handling.py` + `test_panel_oos_shape.py` → **40 passed**
- All pre-commit hooks clean (ruff, ruff-format, mypy, license)
- Branch is up to date with `main`

PR #403, opened 2026-08-07
Guard OOS panel-shape prediction and NaN predictors (towards #327)

daimon-pymclabs:
## Summary

Addresses two sub-components of #327:

- **Item 9 — out-of-sample prediction with a different panel shape.** `predict()` on a scan-compiled panel model (any `lag()`/`adstock()` term) now raises a clear `ValueError` if `pm.set_data()` was used to swap in a differently-shaped panel first, instead of crashing with a cryptic PyTensor error or, worse, silently truncating the extra rows.
- **Item 10 — NaN / missing-data handling.** A NaN in a *predictor* now raises a clear `ValueError` at model-construction time instead of silently propagating into `mu`/logp. A NaN in an outcome that also lags itself (`y ~ lag(y)`) no longer crashes model construction outright.

### Item 9

`n_units`/`n_times` are baked into `_compile_scan_panel`'s RV shapes (`shape=(n_times, n_units)`) and scan-carry initial values (`np.zeros(n_units)`) at model-compile time. Reproduced by fitting a small `y ~ lag(y) + x` panel model, then calling `pm.set_data({"x": ...})` with a differently-shaped array before `predict()`:

- **Different unit count** (e.g. 3 → 5 units): raised `ValueError: Vectorized input 1 has an incompatible shape in axis 0` deep inside `pm.sample_posterior_predictive` — an error, but an unhelpful one that gives no hint the panel shape was the problem.
- **Different time count, same unit count** (e.g. 5 → 8 timesteps): **no error at all.** `pytensor.scan` derives its step count from the *shortest* sequence among the resized exogenous `pm.Data` and the still-fixed-length `carry_innovations_y` RV (baked at fit time), so `predict()` silently returned only 5 timesteps of output for a request that supplied 8. This is exactly the "silent mis-reshape" the issue warned about — confirmed by direct reproduction before the fix (`posterior_predictive["y"].shape` came back `(1, 5, 5, 3)`, not `(1, 5, 8, 3)`).
- Same total row count, different unit/time split (3×5 vs 5×3): also raised (init-carry arrays are sized by the old `n_units`).

**Fix:** `pathmc.compile.validate_panel_scan_shape()` checks every 2-D `pm.Data` node in the compiled model against the compile-time `(n_times, n_units)` shape and raises a `ValueError` naming the offending variable and both shapes, called from `predict()` before sampling. Predicting without touching the data, or re-setting it at the identical shape, is unaffected (verified in tests).

**Behavior change to flag:** this makes previously-"succeeding" (but silently wrong, or crash-with-cryptic-error) calls now raise a `ValueError` with an actionable message. No previously-*correct* input is affected — only inputs that were already broken (silently truncated or already raising an unrelated error) now raise our own, clearer error instead. No real out-of-sample workflow is disabled, because none currently produced correct results for a resized panel.

Out of scope / not fixed here: genuinely *supporting* OOS prediction on a new panel shape (rebuilding the scan for a new `n_units`/`n_times`) would need `predict()` to take new data and recompile, which is a larger redesign. Users should build a new model with `pathmc.model(..., data=new_data, panel=...)` on the new panel instead — noted in the `predict()` docstring's `Raises` section.

### Item 10

- **Predictor NaN:** a NaN in an exogenous column flows straight into a `pm.Data` node and from there into every downstream `mu` (row-wise: plain tensor arithmetic; panel: a `pytensor.scan` sequence), producing a non-finite `mu`/logp with no indication that a NaN predictor was the cause — typically surfacing much later as an unexplained NUTS divergence. `pathmc.compile._reject_nan_predictors()` now raises a `ValueError` naming the column and the bad-value count at model-construction time (`pathmc.model(...)`), for both the row-wise and scan-panel compilers, and independent of whether the panel is balanced (tested against both).
- **Masked-outcome off-by-one check:** verified there is no off-by-one in the `(unit, time)` → `(n_times, n_units)` reshape used to build the masked-outcome array. A NaN placed at one specific `(unit, time)` cell lands at exactly the corresponding `(time_idx, unit_idx)` cell after reshaping — confirmed with a `monkeypatch` spy on `np.ma.masked_invalid` in the new tests, both with and without a self-referencing lag.
- **Concrete bug found and fixed:** a NaN outcome in a variable that lags itself (`y ~ lag(y)`) previously **crashed model construction outright** with an unrelated `pm.Data` error — `NotImplementedError: Masked arrays or arrays with nan entries are not supported` — instead of reaching the masked-likelihood path that was already built to handle it. Root cause: `_compile_scan_panel` builds a `_obs_carry_{var}` node holding the raw (possibly-NaN) values of any self-lagged variable, and its scan step deliberately branches on `pt.isnan(obs_state)` to fall back to the model's own simulated value at missing carry positions — but wrapping that node in `pm.Data` routed every NaN-containing array through PyMC's auto-imputation guard, which unconditionally rejects masked/NaN input regardless of intent. Fixed by using a plain `pytensor.shared` instead (this node is never looked up by name or resized via `pm.set_data`, so nothing is lost).

**Behavior change to flag:** any call that previously passed a NaN predictor and got a confusing downstream failure (or, worse, sampled "successfully" with a poisoned `mu`) will now raise `ValueError` immediately at `pathmc.model(...)` construction, before any sampling. This is intentional — no previously-*correct* fit is affected, since a NaN predictor could never have produced a correct one. The self-lag-NaN-outcome case is a pure bug fix: it previously always raised (a different, wrong error); it now succeeds and matches the masked-likelihood behavior a non-self-lagged NaN outcome already had.

## Relationship to #388

No overlap. #388 (`327-panel-shape-validation`) adds `_validate_panel_shape()`/`require_rectangular` to `pathmc/panel.py`, validating the *training* panel's rectangularity (balanced, gapless, no duplicates) at `pathmc.model()` construction. This PR does not touch `pathmc/panel.py` at all — it adds `validate_panel_scan_shape()` in `pathmc/compile.py`, called from `predict()` in `pathmc/_model.py`, which guards *prediction-time* reshaping on an already-compiled model (i.e. after `pm.set_data()` swaps in new data post-construction). The two are checking different data at different times and do not conflict.

## Commits

1. `Reject out-of-sample panel-shape changes in predict()` — item 9, `tests/test_panel_oos_shape.py`.
2. `Reject NaN predictors and fix self-lag NaN outcome crash` — item 10, `tests/test_panel_nan_handling.py`.

## Test plan

Ran targeted suites in the foreground (the machine had other concurrent agents running the full suite, so the full suite was **not** run for this PR — see note below):

- `tests/test_panel_oos_shape.py` + `tests/test_panel_nan_handling.py` (new): **19 passed**.
- Broader targeted sweep — `test_panel*.py`, `test_do*.py`, `test_lag_syntax.py`, `test_exog_lag_missing_base.py`, `test_stochastic_latent.py`, `test_compile.py`, `test_data_optional.py`, `test_transforms_compile.py`, `test_transforms_do.py`, `test_hsgp_compile.py`, `test_nutpie_scan_compile.py` (everything plausibly touched by the `compile.py`/`_model.py` changes), run once with both commits' changes applied together: **232 passed, 1 skipped**.
- `prek run --all-files`: all hooks passed (ruff, ruff format, mypy, license header, etc.), no fixes needed.
- **Not run:** the full `tests/` suite, deliberately, per instructions to avoid the full suite while the shared sandbox had multiple concurrent agents already running it. The targeted sweep above covers every test file that imports or exercises `pathmc.compile`, `pathmc._model`, or panel/do/predict behavior.
- Pre-existing, unrelated failures observed elsewhere on this branch: `ModuleNotFoundError: No module named 'pyarrow'` in `tests/test_falsify.py::TestBackends::test_polars_backend` (environment gap, not touched by this PR).

## Left unfixed / out of scope

- Genuine OOS prediction support (rebuild the scan for a new panel shape) — noted above, would need a larger `predict(new_data=...)` redesign.
- A pre-existing, adjacent issue noticed but *not* fixed here: the masked-outcome likelihood (`sparse_data`/`sparse_panel_data`) always emits `pm.Normal(observed=masked_array)` regardless of the variable's declared `family` — so a `bernoulli`/`poisson`/`negbinomial` outcome with any NaN silently gets a Gaussian likelihood instead of its declared family. This is a distinct bug from items 9/10 (family-dispatch on masked likelihoods, not shape/NaN-propagation) and is left for a follow-up.


drbenvincent:
## Defect-focused review (towards #327 items 9 & 10)

**Verdict: approve — no correctness blockers found.** CI is green (`prek`, full `test` job). Scope is tight (4 files, two logical fixes) and well-tested.

### Item 9 — OOS panel-shape guard in `predict()`

- **Root cause is correct:** `n_times`/`n_units` are baked into scan RV shapes and carry inits; `pm.set_data()` on exogenous `pm.Data` can silently truncate (more timesteps) or crash cryptically (more units). The reproduction described in the PR body matches how `pytensor.scan` picks step count.
- **`validate_panel_scan_shape()` is sound for current scan models.** I spot-checked 2-D `named_vars` with `get_value` on a `y ~ lag(y) + x` panel model — only the exogenous `pm.Data` (`x`) is 2-D at that layer, so the blanket loop is safe in practice today. (Nit: it keys off any 2-D `get_value` tensor, not strictly `pm.Data`; fine now, but worth keeping in mind if future scan models add other 2-D constants.)
- **Integration point is right:** called from `PathModel.predict()` only when `_pathmc_panel_scan` is set; cross-sectional / non-temporal models are untouched.
- **Tests are strong:** parametrized shape mismatches (including the silent-truncation case), same-shape `set_data` baseline, and message-content assertions. Fake-posterior setup keeps them fast without sacrificing the actual failure mode.

**Intentional behavior change (good):** previously-wrong OOS reshapes now raise a clear `ValueError` with remediation (`pathmc.model(..., data=new_data, panel=...)`). No correct workflow is disabled.

**Follow-up (non-blocking):** the guard lives only on `predict()`; direct `pm.sample_posterior_predictive()` after `pm.set_data()` (e.g. via `simulate`/`refute`) could still hit the old failure modes. Acceptable for this PR since `predict()` is the documented entry point.

### Item 10 — NaN predictors + self-lag outcome crash

- **`_reject_nan_predictors()` placement and scope are correct:** runs in `compile_to_pymc` before both row-wise and scan paths; uses `graph_info.exogenous` so outcome NaNs (masked likelihood) are not double-flagged. `inf` is covered too.
- **`pytensor.shared` for `_obs_carry_{var}` is the right minimal fix:** `pm.Data` auto-imputation guard was rejecting NaN carry state that the scan step deliberately branches on via `pt.isnan(obs_state)`. The node is internal, never resized via `set_data`, so nothing is lost.
- **Mask positioning tests are valuable:** monkeypatch spies on `np.ma.masked_invalid` / `pytensor.shared` confirm no off-by-one in `(unit, time) → (n_times, n_units)` reshaping, with and without self-lag.
- **End-to-end self-lag NaN case:** builds and prior-predictive draws are finite.

**Pre-existing gap (out of scope, already noted in PR):** NaN in an *endogenous mediator* used as a downstream predictor is still not rejected here (mediators are not in `exogenous`). That's a separate follow-up, not a regression from this PR.

### Tests run locally

```
tests/test_panel_oos_shape.py + tests/test_panel_nan_handling.py → 19 passed
```

(at commit `5654dcf`, isolated worktree)

### Merge readiness notes

- Branch is **mergeable** but **6 commits behind `main`** — rebase/merge before final merge (no conflicts indicated).
- Still **draft**; marking ready is left to the author.
- Distinct from #388 (`panel.py` training-time rectangularity) — no overlap.

**Recommendation:** approve and merge after rebase + undraft.

drbenvincent (approved):
LGTM. Both fixes address real failure modes (silent scan truncation, NaN predictor poisoning, self-lag `pm.Data` crash) with clear errors and solid targeted tests. No correctness blockers; rebase onto current `main` before merge.

drbenvincent:
## Merge conflicts resolved (`884aa44`)

Merged current `main` into `327-oos-shape-and-nan`.

**Conflict:** `pathmc/_model.py` imports — kept both:
- `validate_panel_scan_shape` (this PR's OOS shape guard on `predict()`)
- `_has_temporal_deps` (panel rectangular validation from #388 on `main`)

`compile.py` auto-merged cleanly.

**Verification:** `test_panel_oos_shape.py` + `test_panel_nan_handling.py` + `test_panel_shape_validation.py` → **35 passed** locally; pre-commit hooks green on push.

daimon-pymclabs:
## Bugbot finding addressed (`21ec892`)

**"Self-lag NaN skips observed carry"** — confirmed real and high severity, exactly as described.

`_compile` set `_use_observed_carry` to 1 only inside the `if observations:` branch. A variable holding any NaN is skipped from `pm.observe` (its likelihood is already masked inside the compiled model), so a model whose every outcome has a NaN — the `y ~ lag(y)` with missing outcomes case this PR exists to unlock — left the flag at 0. The fit then free-ran the scan, using simulated instead of observed previous values *for the timesteps that are observed*. Wrong AR likelihood, silently.

**Fix:** the flag is set whenever the compiled model carries it, independent of whether `observations` is empty. One consequence worth naming: when `observations` is empty the observation model *is* the generative model, so `sample_prior_predictive` now forces the flag off for the duration of the call (via the existing `_observed_carry` contextmanager) and restores it afterwards. Prior predictive is free-running by definition and must not depend on which branch `_compile` took.

**Verification** — two regression tests added, both confirmed failing on this branch before the fix (`assert 0 == 1`):
- `test_observed_carry_enabled_when_every_outcome_has_a_nan`
- `test_prior_predictive_stays_free_running`

`tests/test_panel_nan_handling.py` + `test_panel_oos_shape.py` + `test_panel_shape_validation.py` → **37 passed**; `ruff check` and `ruff format --check` clean on both touched files.

PR #395, opened 2026-08-07
fix: restore exog-lag scan carry after PyTensor fix

drbenvincent:
## Summary
- Raise the minimum PyTensor version to 3.1.1, the first released line whose notes include upstream PR #2263 fixing #2252.
- Restore the exogenous-lag sit-sot carry path and remove the temporary pre-computed sequence workaround.
- Pass the known panel length to `scan` so missing-base lag panels remain supported when no input sequence is available.

## Validation
- `uv run pytest tests/test_graph_consistency.py -x -q` — 18 passed
- `uv run pytest tests/test_exog_lag_missing_base.py -x -v` — 4 passed
- `uv run python -m compileall -q pathmc`
- `make lint`

Closes #333

Made with [Cursor](https://cursor.com)

PR #404, opened 2026-08-07
Fix do() severance through the panel scan's temporal recursion (towards #327)

daimon-pymclabs:
## Sub-component

Addresses **item 2 only** of #327: *interventions through the scan (`pm.do`)*. Items 1, 3–12 are out of scope (several have their own open PRs: #388, #389, #390, #391; #387 is merged).

## Finding: confirmed bug, fixed

The scan-compiled panel model exists "so `pm.do()` handles interventions natively" (`compile.py` docstring). It does **not**, for endogenous variables.

Root cause: inside `_compile_scan_panel`'s `step_fn`, `lag()` and same-timestep contemporaneous references are resolved from Python-level `new_endo` / `prev_endo` dicts that are rebuilt fresh at every scan step from that step's own structural computation. They never re-read the outer free RV / `Deterministic` node named `var` that `pm.do()` performs graph surgery on. So intervening on an *endogenous* variable (e.g. one read via `lag(y)` by another equation, or contemporaneously by a sibling equation) was applied only cosmetically to the reported value at the intervened timestep — every other equation and every future timestep's `lag()` term kept using the model's natural, un-intervened dynamics. Exogenous interventions were unaffected (they're wired through real `pm.Data` scan sequences and already propagated correctly).

**Evidence (deterministic oracle, no MCMC — coefficients and `sigma` pinned to fixed posterior values):**

Model: `y ~ x` (β=2), `z ~ lag(y)` (β=3), single unit, 6 timesteps, `x = [1..6]`.

`do(set={"y": [10, 20, 30, 40, 50, 60]}, simulate_over="time")`:

| | z (before fix) | z (after fix) | closed-form analytic |
|---|---|---|---|
| result | `[6, 6, 12, 18, 24, 30]` | `[6, 30, 60, 90, 120, 150]` | `[6, 30, 60, 90, 120, 150]` |

Before the fix, `z` is exactly `3 * (natural mu_y = 2*x)` shifted by one step — the intervention on `y` has zero effect on `z`, despite `z`'s only structural input being `lag(y)`. After the fix it matches `3 * y_{t-1}` under the intervention exactly (the standard SCM closed form for a severed structural equation).

## Fix

Added a dedicated `_do_intervene_{var}` override channel (one per endogenous var, incl. latent) wired directly into the scan as a NaN-sentinel sequence. `step_fn` switches to the override *inside* the recursion, immediately after computing each var's natural value — so every same-timestep and future-timestep consumer inside the scan sees the severed value. The channel is a plain shared variable (`pm.Data` rejects NaN outright as an imputation request) set/restored around each `do()` call (`simulate._scan_do_intervention`) so state does not leak between repeated calls on a reused model object. Exogenous-variable interventions are unchanged (still handled via the pre-existing `pm.Data`/`pm.do()` path).

## Tests

`tests/test_scan_intervention_oracle.py` — 16 deterministic, no-sampling tests (parametrized over `kind="mean"`/`"predictive"`), covering:
- intervening on an endogenous variable that is a `lag()` base for another equation (the core bug),
- the intervened variable's own reported value,
- a mid-recursion trajectory that matches natural dynamics for several periods then diverges, checking the downstream effect starts exactly one step later (not earlier/later),
- a no-intervention baseline matching natural recursion,
- repeated `do()` calls with different/no interventions on the same fitted model (no state leakage),
- the override channel resetting to all-NaN after each call,
- the pre-sample carry (t=0, baked from training data) correctly *not* being overridable by the intervention array, while later periods correctly are.

## Test plan

- Targeted (foreground): `test_scan_intervention_oracle.py`, `test_do*.py`, `test_counterfactual.py`, `test_transforms_do.py`, `test_panel*.py`, `test_compile.py`, `test_lag_syntax.py`, `test_graph_consistency.py`, `test_causal_queries.py`, `test_exog_lag_missing_base.py`, `test_stochastic_latent.py` — **226 passed**.
- Full suite: `pytest tests/` — **1406 passed, 1 skipped, 10 failed**. All 10 failures are pre-existing `ModuleNotFoundError: No module named 'pyarrow'` in polars-backend tests (`test_falsify.py`, `test_narwhals_backends.py`), unrelated to this change and present on `main` in this environment.
- `prek run --all-files` — all hooks pass (ruff-format reformatted the new test file once; re-ran clean).

## Overlap note

PR #398 (`fix/394-panel-do-dispatch`) also touches `pathmc/simulate.py`, but only adds a docstring comment to `_forced_generative_carry` — no functional overlap with this PR's changes (a new context manager `_scan_do_intervention` and edits to `run_do_panel_unified`). Should merge cleanly either order.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


drbenvincent:
## Triage + review

**Actions taken:** Resolved merge conflict with `main` (pushed `0bcdf13`), which also cleared the prior `prek` failure. CI: `prek` green; `test` still running on the merge push.

**Defect-focused review:** The scan `do()` severance fix is sound. Wiring `_do_intervene_{var}` shared variables into `step_fn` via `pt.switch` correctly severs structural recursion inside the scan — graph surgery on the outer RV cannot reach lag/contemporaneous references inside `step_fn`, and this channel fixes that. `_scan_do_intervention` context manager correctly sets shared vars directly (avoiding `pm.set_data` NaN→masked-array conversion). `run_do_panel_unified` wires both outer replacements and scan overrides for endogenous vars.

**Tests:** `tests/test_scan_intervention_oracle.py` — **16 passed** locally on tip.

**Recommendation:** **Approve-worthy** once `test` CI completes green. Branch should rebase/merge latest `main` before final merge (merge commit already integrated current `main`).

drbenvincent (approved):
LGTM. Scan do()-intervention override channel correctly severs temporal recursion inside step_fn; tests discriminate the failure mode. Merge after test CI confirms green.

drbenvincent:
## Autopilot triage summary

### Merge conflicts (resolved)
Merged `main` into `327-scan-intervention-oracle` and resolved the conflict in `pathmc/compile.py`.

**Resolution:** kept `main`'s exog-lag scan **carry** path (#395) rather than the PR's older `lagged_exog_sequences` workaround. Wired the PR's `_do_intervene_{var}` override channel on top of that layout and fixed `do_intervene_t` sequence indexing (no `n_exog_lag_seq` offset — exog lag lives in carry, not sequences).

Pushed as commit `0bcdf13`.

### CI (prek)
**Root cause:** `ruff-format` had not been applied to `tests/test_scan_intervention_oracle.py` in the original push.

**Fix:** ran `ruff format` on the changed files; `prek` now passes locally and on CI.

### Local verification
Scoped suite (oracle + do/panel/compile/lag tests): **183 passed** in ~2m45s.

Full CI `test` job was still running at comment time (~32m expected); it passed on the prior push and only merge-resolution + formatting changed.

### Code review (defect-focused)

**Correctness:** The fix is sound — `pm.do()` graph surgery cannot sever edges inside the scan's `step_fn` recursion; the NaN-sentinel shared override + `pt.switch` inside the endogenous loop is the right place to sever. `_scan_do_intervention` correctly avoids `pm.set_data` (masked-array / numba issues) and restores state on exit. Context managers wrap both `mean` and `predictive` paths including latent deterministics.

**No blocking defects found.**

**Minor test gap (non-blocking):** tests cover `lag(y)` propagation thoroughly but not a contemporaneous sibling equation (another endogenous var reading `y` at the same timestep). Worth a follow-up test if we want parity with the PR description's stated scope.

### Recommendation
Merge-ready once full `test` CI is green. Draft status intentionally left unchanged.

drbenvincent (approved):
Conflicts resolved cleanly (main's exog-lag carry + do-intervention override channel), prek fixed, and full CI is green after merge commit `0bcdf13`. Code review found no blocking defects; oracle tests give strong coverage of the lag-propagation bug. Ready for human review / marking out of draft.

PR #405, opened 2026-08-08
Add single-page docs freeze target

### LINKED ISSUE #382: Add a make freeze-page target so the single-page freeze incantation isn't duplicated in docs
## Context

#381 consolidated Jupyter kernel registration behind `make jupyter-kernel` and scoped the kernelspec to the venv with `--sys-prefix`, replacing the four hand-copied `ipykernel install` invocations with one source of truth.

That left one smaller instance of the same duplication. Freezing a *single* page has no Make target, so both `AGENTS.md` and `CONTRIBUTING.md` now spell out the incantation by hand:

```bash
make jupyter-kernel
JUPYTER_PATH=$PWD/.venv/share/jupyter uv run great-docs freeze docs/examples/<notebook_name>.qmd
git add _freeze/
```

The `$PWD/.venv/share/jupyter` path is hardcoded in two prose files while the `Makefile` already computes the same value as `VENV_JUPYTER`. If that layout ever changes (a different venv location, a `--user` fallback, an upstream `great-docs` change), the Makefile gets updated and the two docs copies silently rot — which is the failure mode #381 set out to remove.

## Proposal

Add a single-page freeze target alongside `docs` / `refreeze-docs`:

```make
freeze-page: jupyter-kernel ## Re-execute one doc page and refresh its _freeze/ cache (PAGE=docs/examples/.../my_page.qmd)
	@test -n "$(PAGE)" || { echo "Usage: make freeze-page PAGE=docs/examples/01-foundations/my_page.qmd"; exit 1; }
	JUPYTER_PATH=$(VENV_JUPYTER) uv run great-docs freeze $(PAGE)
	@echo "Freeze cache refreshed. Commit with: git add _freeze/"
```

Then both docs collapse to:

```bash
make freeze-page PAGE=docs/examples/01-foundations/my_page.qmd
git add _freeze/
```

This also drops the separate `make jupyter-kernel` line from the documented workflow, since the target already depends on it.

## Scope

- `Makefile`: add `freeze-page`, add it to `.PHONY`
- `CONTRIBUTING.md`: replace the two-command block in the docs section
- `AGENTS.md`: same replacement

Small, docs/dev-ops only, no user-facing behavior change.

alloutflo:
## Summary

- add a `freeze-page` Make target that reuses `VENV_JUPYTER` and registers the project kernel automatically
- validate `PAGE` and quote the selected page path before passing it to Great Docs
- update the contributor and agent docs to use the new single source of truth

## Test plan

- `make help`
- `make -o jupyter-kernel freeze-page` (expected usage error)
- `make -n freeze-page PAGE=docs/examples/01-foundations/my_page.qmd`
- `make freeze-page PAGE=docs/user_guide/15-standardized-effects.qmd`
- `make check_lint`
- `make lint`
- `make test-fast` (1,356 passed, 2 skipped)
- `make docs`
- `make check-build BUILD_CHECK_DIR=/tmp/pathmc-build-check-382`

The full `make test` run reached 1,455 passing tests before a local Command Line Tools failure (`fatal error: 'vector' file not found`) while compiling a PyTensor extension. The same test fails identically on unchanged `upstream/main`, so it is unrelated to this docs/dev-ops-only change.

Closes #382

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,11 +55,10 @@ This package targets **PyMC ≥ 6.0, ArviZ ≥ 1.1, PyTensor ≥ 3.0, and NumPy 
 
 The site is built with [Great Docs](https://posit-dev.github.io/great-docs/) via `great-docs.yml`: `uv run great-docs build`. The output dir `great-docs/_site/` is ephemeral and gitignored — never edit it. Edit sources instead: `docs/user_guide/*.qmd`, `docs/examples/**/*.qmd`, `great-docs.yml`, and `pathmc/skills/pathmc/SKILL.md`.
 
-Notebooks are frozen (`freeze: true`): the committed `_freeze/` cache supplies cell outputs and CI never spawns a kernel, so **local previews show the last-frozen output, not in-progress edits.** After editing an executable page, refresh and commit the cache (local freezing needs the kernel once via `make jupyter-kernel`):
+Notebooks are frozen (`freeze: true`): the committed `_freeze/` cache supplies cell outputs and CI never spawns a kernel, so **local previews show the last-frozen output, not in-progress edits.** After editing an executable page, refresh and commit the cache:
 
 ```bash
-make jupyter-kernel
-JUPYTER_PATH=$PWD/.venv/share/jupyter uv run great-docs freeze docs/examples/01-foundations/my_page.qmd
+make freeze-page PAGE=docs/examples/01-foundations/my_page.qmd
 git add _freeze/
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ make lint
 
 The documentation site is built with [Great Docs](https://posit-dev.github.io/great-docs/), with [Quarto](https://quarto.org/docs/get-started/) as the underlying renderer. Install Quarto separately before running the docs commands.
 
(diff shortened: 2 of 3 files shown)


drbenvincent (approved):
Thanks for picking this up, @alloutflo — really nice first contribution! This is exactly what #382 asked for: a focused change that removes duplicated docs incantations and centralises the single-page freeze workflow behind `VENV_JUPYTER`, matching the patterns already used by `docs` and `refreeze-docs`.

**What looks great**
- Clean scope: only `Makefile`, `AGENTS.md`, and `CONTRIBUTING.md` as specified in the issue.
- `freeze-page` follows existing conventions (`jupyter-kernel` dependency, helpful usage message, trailing commit reminder).
- Quoting `"$(PAGE)"` is a small improvement over the issue snippet — handles paths with spaces safely.
- Thorough test plan in the PR description; I also verified `make help`, `make -n freeze-page PAGE=...`, and the missing-`PAGE` usage error locally.

**Optional polish (not blocking)**
- `.github/workflows/docs.yml` and `great-docs.yml` still mention `great-docs freeze <page>` in comments. Since `AGENTS.md` now documents `make freeze-page`, you could update those comments for consistency in a follow-up — totally fine to leave as-is and merge this as-is.

No security concerns from my side either — this is developer-only tooling with no CI exposure.

Happy to merge once CI is green. Welcome aboard, and thanks again! 🎉

PR #407, opened 2026-08-09
Guard effect() and standardized() against non-Gaussian families

### LINKED ISSUE #149: [HIGH] effect() and standardized() invalid for non-Gaussian models — missing guards
## Summary

Two functions in `effects.py` produce uninterpretable results for non-Gaussian models without any warning:

### 1. `compute_path_effect` (product-of-coefficients)

Multiplies coefficient draws along a path. This equals the natural indirect effect **only under linearity**. For a path X→M→Y where M ~ Bernoulli, the coefficients are on the logit scale and their product is not the mediated effect on any meaningful scale.

### 2. `build_standardized_effects` (STDYX)

Computes `coef * sd(X) / sd(Y)` without checking the family. For Bernoulli outcomes, the coefficient is on the logit scale and sd(Y) = sqrt(p*(1-p)), making the "standardized" value uninterpretable.

## Impact

A user who specifies `families={"M": "bernoulli"}` and calls `.effect("X -> M -> Y")` or `.standardized()` will get a confident-looking but **meaningless** value. There is no error, no warning — just a wrong number.

## Root cause

Neither function has access to `families` information. The `families` dict is stored on `PathModel` but not passed through to these effect functions.

## Fix

1. Pass `families` through to `compute_path_effect` and `build_standardized_effects`
2. For `effect()`: raise `NotImplementedError` or emit a warning when any edge in the path involves a non-Gaussian family
3. For `standardized()`: skip or warn for non-Gaussian outcomes, similar to how latent variables are already skipped
4. Add a note to the `effect()` docstring that this method assumes linear structural equations

## Reference

Technical audit report: `docs/dev/technical_audit_2026_03_16.md` finding H3.

drbenvincent:
## Summary

Fixes #149. Prevents `effect()` and `standardized()` from producing uninterpretable results for non-Gaussian models (Bernoulli, Poisson, NegBinomial).

- **`effect(path)`**: raises `NotImplementedError` when any node on the path has a non-Gaussian family, directing users to `do()`-based simulation instead
- **`standardized()`**: skips non-Gaussian outcomes (analogous to how latent variables are already skipped) and emits a `UserWarning` listing the skipped variables
- Both functions now accept a `families` parameter, threaded through from `PathModel`
- `effect()` docstring documents the linearity assumption

## Design decisions

- **Hard error for `effect()`** rather than a warning: the product-of-coefficients value is genuinely meaningless for non-linear links — there's no "approximately right" answer.
- **Soft skip for `standardized()`**: follows the existing latent-variable pattern (skip and continue), since the function already returns partial results. Warning ensures discoverability.
- Backward compatible: both new parameters default to `None`, so existing direct callers of `compute_path_effect` / `build_standardized_effects` are unaffected.

## Test plan

- [x] 7 unit tests for `_guard_non_gaussian_path` (Gaussian passes, each non-Gaussian family raises, message content)
- [x] 4 integration tests via `PathModel` with mock sampling (Bernoulli mediator raises, Gaussian passes, standardized skips with warning, Gaussian no warning)
- [x] Existing `test_effects.py` and `test_standardized.py` pass unchanged
- [x] Full `make test-fast` suite (612 pass, 1 pre-existing skip for missing pyarrow)
- [x] `make lint` passes (ruff, ruff-format, mypy, license)


drbenvincent:
## Review: fixes the core issue, but `effect()` guard is too aggressive

### Core issue (#149)

`effect()` multiplies regression coefficients along a path (Wright's tracing rule). That product is only meaningful when every coefficient lives on a **compatible linear scale**. For non-Gaussian outcomes, coefficients are on the link-function scale, so multiplying them (or applying STDYX `coef * sd(X)/sd(Y)`) yields confident-looking nonsense.

Root cause was correct in the issue: `PathModel` stores `self._families` but never passed it into `compute_path_effect` / `build_standardized_effects`.

### What works well

- **Threading `families` through `PathModel.effect()` / `.standardized()`** — addresses the root cause.
- **`standardized()` skip + `UserWarning`** — matches the existing latent-variable pattern; skipping by **outcome** (`lhs`) is the right granularity.
- **`effect()` hard `NotImplementedError`** — appropriate for a genuinely meaningless quantity; message points users to `do()` simulation.
- **Family set** `{bernoulli, poisson, negbinomial}` aligns with discrete families elsewhere in `compile.py`; `studentt` correctly omitted (identity link on the mean).
- **Tests + CI** — good unit coverage for the helper; integration tests exercise the `PathModel` path. All checks green.

### Bug: `_guard_non_gaussian_path` checks all path nodes, not edge targets

The helper docstring says *"endogenous node"*, but the implementation flags **every** node on the path:

```python
bad = [n for n in nodes if families.get(n, "gaussian") in _NON_GAUSSIAN_FAMILIES]
```

Each edge coefficient is on the **target** regression's link scale. The source node's family is irrelevant to that edge's coefficient.

**False positives:**

1. **`M -> Y`** with `families={"M": "bernoulli"}` and Gaussian `Y` — single edge `Y ~ M`; coefficient is on `Y`'s identity-link scale. This is a valid direct effect, but the guard raises because `M` is on the path.
2. **`X -> Y`** with `families={"X": "poisson"}` — exogenous `X`'s family does not enter the `Y ~ X` coefficient; guard raises on `X`.

Repro on the branch:

```python
_guard_non_gaussian_path(["M", "Y"], {"M": "bernoulli", "Y": "gaussian"}, "M -> Y")  # raises — should not
_guard_non_gaussian_path(["X", "Y"], {"X": "poisson", "Y": "gaussian"}, "X -> Y")    # raises — should not
```

The integration tests don't cover `effect("M -> Y")` on a Bernoulli mediator model (only the multi-hop `X -> M -> Y` case).

**Suggested fix:** only guard when the path has **multiple edges** and **any edge target** is non-Gaussian:

```python
if len(nodes) < 3:
    return  # single-edge path: one coefficient, no cross-scale product
targets = nodes[1:]
bad = [n for n in targets if families.get(n, "gaussian") in _NON_GAUSSIAN_FAMILIES]
```

This correctly blocks `X -> M(bernoulli) -> Y`, `X -> M -> Y(bernoulli)`, etc., while allowing single-edge paths.

Also flip `test_exogenous_non_gaussian_still_raises` — exogenous nodes should **not** trigger the guard.

### Minor gaps (non-blocking)

- **`PathModel.standardized()` docstring** doesn't mention that non-Gaussian outcomes are skipped with a warning (unlike `effect()`, which was updated).
- **Defined parameters** (`indirect := a * b`) in `effects_summary()` can still multiply link-scale coefficients with no guard — outside this PR's scope, but worth a follow-up issue if not already tracked.

### Verdict

**Not fully resolved** — merge after fixing the over-broad `effect()` guard (and adjusting tests). The `standardized()` side looks correct for the issue as scoped.

drbenvincent:
## Review summary

The narrowed `_guard_non_gaussian_path` logic looks correct: it only blocks multi-edge paths (`len(nodes) >= 3`) where an **edge target** has a non-Gaussian family, and correctly ignores the source node's family. That matches the product-of-coefficients rationale and addresses the prior over-aggressive guard. `standardized()` skipping non-Gaussian outcomes with a warning is also reasonable. Unit + integration coverage for the guard helper is solid.

### 🟡 Should fix

1. **`PathModel.effect()` docstring is stale after the narrowing** (`pathmc/_model.py`, Raises section): it still says `NotImplementedError` is raised when *"any node on the path has a non-Gaussian family"`. The implementation only raises for **multi-edge** paths whose **edge targets** are non-Gaussian (single-edge paths like `M -> Y` with Bernoulli `M` are explicitly allowed). Please align the `Raises` text (and the opening paragraph that broadly tells users to use `do()` for all non-Gaussian models) with `_guard_non_gaussian_path` / `compute_path_effect` docstrings.

2. **Weak assertion in `test_standardized_skips_bernoulli_with_warning`**: the check `if "b" in result.index or "c" in result.index: assert "a" not in result.index` is conditional — if neither `b` nor `c` appears (e.g. sampling noise), the test passes without verifying that coefficient `a` (X→M, non-Gaussian outcome) was skipped. Prefer unconditional assertions, e.g. `assert "a" not in result.index` plus `assert {"b", "c"} & set(result.index)` (or similar).

### 🟢 Nitpick

- `_NON_GAUSSIAN_FAMILIES` duplicates the discrete-family set already used in `compile.py` (`{"bernoulli", "poisson", "negbinomial"}`). A shared constant would prevent drift if a new family is added later.

No correctness or regression blockers found in the guard logic itself.

drbenvincent:
Reviewed — no actionable issues found. Approving.

Issue #123, opened 2026-03-13
Replace custom matplotlib plots with ArviZ where possible


drbenvincent:
## Summary

pathmc has extensive custom matplotlib plotting code — mostly in documentation notebooks, with one instance in the library itself (`SensitivityResult.plot()`). Many of these custom plots re-implement functionality that ArviZ already provides (`plot_posterior`, `plot_kde`, `plot_ppc`, `plot_forest`, `plot_hdi`). Replacing them would reduce boilerplate, improve visual consistency, and give users a familiar ArviZ-based experience.

This issue tracks the opportunities and proposes a phased approach.

> **Post-consolidation update (March 2026):** Issue #122 consolidated the docs from ~30 example notebooks down to 18. This issue has been refreshed to reflect the current notebook inventory. The core proposal is unchanged.

---

### ArviZ usage today

| Usage | Where |
|-------|-------|
| `az.summary()` | `model.py`, tests |
| `az.hdi()` | `effects.py`, `simulate.py`, `sensitivity.py` |
| `az.InferenceData` | type hints and return values throughout |
| `az.plot_posterior()` | `did.qmd` (1 notebook) |
| `az.plot_kde()` | `data_simulation.qmd` (1 notebook) |

ArviZ is used heavily for computation but almost never for plotting.

### Custom plotting in the library

| Location | What | ArviZ replacement? |
|----------|------|---------------------|
| `sensitivity.py` — `SensitivityResult.plot()` | Contour plot of bias-adjusted ATE | No — ArviZ has no sensitivity contour plot |
| `introspect.py` — `build_dag_viz()` | Graphviz DAG | No — ArviZ has no DAG rendering |

These two are domain-specific and should remain custom.

### Custom plotting in documentation notebooks

15 notebooks currently contain custom plotting code. The patterns fall into four categories.

---

## Opportunity 1: Posterior KDE → `az.plot_posterior` or `az.plot_kde`

**Pattern**: Extract `.posterior[var].values.flatten()`, compute `gaussian_kde`, plot with `fill_between` + `axvline` for mean.

**Affected notebooks** (7 use `gaussian_kde`):

| Notebook | What's plotted | Suggested ArviZ function |
|----------|----------------|--------------------------|
| `examples/vaccine_surrogates.qmd` | KDE of do() effect draws (total, indirect, direct) | `az.plot_kde` |
[...]

PR #409, opened 2026-08-09
Improve fix-bug skill: fixer comments, scope selection, closing semantics

drbenvincent:
## Summary

- Add **fixer summary comments** (`<!-- fix-bug-fix-summary -->`) — required after initial PR creation and after each review-fix push, so agents document root cause, approach, and what's left out of scope
- Add **umbrella/meta issue scoping** — one bug per PR; pick highest-priority sub-issue or one checklist item; record `SCOPE_ISSUE` before coding
- Improve **closing semantics** — dedicated `## Closes` section in PR body; `Fixes` vs `Part of` vs `Related to` table; under-close by default on meta issues
- Point reviewers at fixer summary comments for intent context (still judge from the diff)

## Test plan

- [x] Skill markdown only — no code changes
- [ ] Manual: run `fix bug #<umbrella-issue>` and confirm agent scopes to one item and posts a fix summary comment


daimon-pymclabs:
## Review: fix-bug skill improvements

Reviewed the diff and the skill as a whole. The direction is right — a durable fixer-summary comment is the correct place to park scope state for resume, and defaulting to under-closing (`Part of`) on meta issues is the safe call. Below are the things I'd fix before merging, then some whole-skill observations.

### 🔴 Round-count fallback now double-counts fixer summaries

Phase 1 keeps this fallback:

> **Fallback when markers are missing**: Use `max(marker_count, substantive_review_comment_count)` as round estimate.

This PR adds a *required* fixer summary comment after initial push **and after every review-fix push**. Those are substantive PR comments. After two real review rounds a PR carries 2 round markers plus ~3 fix summaries, so the fallback estimates round 3 and escalates immediately, without a third review ever running. The failure is silent: it looks like a legitimate exhaustion.

Fix: define the fallback explicitly as comments that carry neither `fix-bug-fix-summary` nor `fix-bug-escalation`, e.g.

```bash
gh pr view $PR_NUMBER --json comments -q \
  '[.comments[].body | select(test("fix-bug-fix-summary|fix-bug-escalation") | not)] | length'
```

### 🔴 The sub-issue detection command does not detect sub-issues

```bash
gh issue list --search "is:issue $ISSUE_NUMBER"
```

`is:issue 327` is a free-text search for the token `327`, not a parent→child query. I ran it against #327 in this repo: returns `[]`. It will also produce false positives on any issue whose body happens to contain that number.

GitHub's sub-issues endpoint is what you want, and it works here (verified — returns `[]` for #327 rather than 404, i.e. valid endpoint, no children):

```bash
gh api repos/pymc-labs/pathmc/issues/$ISSUE_NUMBER/sub_issues \
  --jq '.[] | {number, title, state}'
```

Keep a cross-reference scan as a *secondary* signal if you like, but the primary lookup should be the API.

### 🔴 `$SCOPE_ISSUE` never expands inside the heredocs

Every template in the skill uses `<<'EOF'` (quoted), which suppresses all expansion. The new lines say:

```
Fixes #SCOPE_ISSUE. Part of #PARENT if umbrella. Why this approach.
```

An agent following this literally posts the string `Fixes #SCOPE_ISSUE`, and GitHub links nothing. The rest of the document uses `#N` / `<angle bracket>` placeholders, so this is also internally inconsistent — an agent has to guess which tokens are shell variables and which are fill-ins.

Suggest one convention throughout: `#<scope-issue>` and `#<parent-issue>`, matching the existing `<1-2 sentences on the approach.>` style. Don't unquote the heredocs; the bodies contain backticks.

### 🟡 Step 9 clobbers the PR body

`gh pr edit $PR_NUMBER --body "..."` is a full replacement. Anything a human added since — ticked checkboxes, extra context, a linked discussion — is silently destroyed. Given this workflow's whole safety story is "never overwrite human work" (Robustness notes), this is the one place it does exactly that. Either read the current body first (`gh pr view $PR_NUMBER --json body -q .body`) and edit it, or restrict step 9 to editing only the `## Closes` block.

### 🟡 Approval and reviewer-crash are the same signal

Pre-existing, but this PR makes it matter more. The reviewer signals approval by posting a comment *without* a marker; a reviewer that dies before posting also leaves no marker. Robustness notes acknowledge the distinction ("If it returns without posting a comment... count as a failed round") but nothing on the PR records it, so a *resumed* session cannot tell an approved PR from a crashed reviewer. Now that you're teaching resume logic to reconstruct state from comment markers, add the missing one: `<!-- fix-bug-approved -->` on the approval comment. Then absence of any comment unambiguously means reviewer failure.

### 🟡 The new reviewer criterion exceeds the reviewer's permitted commands

The reviewer prompt states: "You NEVER modify code, push, or run commands other than reading the diff and the repo's `AGENTS.md` / `CONTRIBUTING.md`." The new scope-alignment criterion requires comparing the diff against the PR **Closes** section and the fixer summary comment — neither is in the diff. Add `gh pr view $PR_NUMBER --json body,comments` to the explicitly permitted reads, otherwise a well-behaved reviewer will refuse the check.

Related, minor: feeding the fixer's stated intent to the reviewer is a mild anchoring risk. The "still judge from the diff" caveat helps; I'd strengthen it to something like "if the diff and the summary disagree, the diff is the truth and the disagreement is itself a 🔴 finding."

### 🟡 Umbrella issues: nothing handles the remainder

"One bug per PR" is right, but the workflow now terminates after fixing one checklist item with no defined behaviour for the other N-1. From the user's side, `fix bug #327` appears to have half-failed. Worth adding a closing step: on success, report which item was taken and enumerate what remains (in the final message to the user, and optionally as a comment on the parent). The `**Not in scope**` field in the fix summary is close, but it's addressed at the next automation pass rather than the human who invoked the loop.

### 🟢 Repo integration is drifting

Not this PR's fault, but it lands on it: `CONTRIBUTING.md:30` still says the repo "ships **five** Great Docs Agent Skills under `.agents/skills/`" and pins them via `skills-lock.json`. There are now six directories, and `fix-bug` is hand-written, unlisted in the lockfile, and undocumented. Since CONTRIBUTING tells contributors to refresh with `npx skills add https://posit-dev.github.io/great-docs/`, it's worth stating explicitly that `fix-bug` is repo-local and not managed by that command (and confirming the refresh doesn't prune unmanaged directories).

### 🟢 Nits

- Frontmatter example `"fix bug PR #408"` now points at the merged PR that added the skill. Use a neutral number to avoid implying it's a live work item.
- The `## Closes` heading contains `Part of` lines, which reads contradictory. `## Scope` or `## Issue links` describes the content better.
- `gh issue view $ISSUE_NUMBER --json title,body,labels,comments` in the new block drops `state` (present in the Phase 1 call). Cheap to include, and closed sub-issues matter for the selection table.

### Overall

The scoping and closing-semantics work is a real improvement and I'd merge it after the three 🔴 items. The round-count regression is the one that would bite in production, since it degrades into premature escalation rather than a visible error.

drbenvincent:
## Review feedback addressed (`1343877`)

Thanks for the thorough review — all items below are addressed in the latest push.

### 🔴 Must-fix items

**Round-count fallback** — Fallback now excludes orchestration markers (`fix-bug-fix-summary`, `fix-bug-escalation`, `fix-bug-approved`) so fixer summaries no longer trigger premature escalation.

**Sub-issue detection** — Replaced the broken `gh issue list --search` with the `sub_issues` API via `gh api "repos/$REPO/issues/$N/sub_issues"`. Body scanning stays as a secondary signal.

**Placeholder convention** — Replaced `$SCOPE_ISSUE` with `#<scope-issue>` / `#<parent-issue>` throughout, matching existing `<angle bracket>` fill-ins. Added an explicit note that heredocs stay quoted.

### 🟡 Should-fix items

**Step 9 body edits** — Must read the current body and update only `## Issue links`; never replace the full description.

**Approval marker** — Reviewer posts `<!-- fix-bug-approved -->` on approval. Loop control and resume logic key off this; no comment = reviewer failure, not approval.

**Reviewer permissions** — Explicitly allows `gh pr view --json body,comments`. Diff-vs-summary disagreement is a 🔴 finding.

**Umbrella wrap-up** — New subsection on approval: tell the user what was fixed and what remains; optionally comment on the parent issue.

### 🟢 Nits + repo integration

- Frontmatter example PR changed from #408 → #123
- `## Closes` renamed to `## Issue links`
- `state` included in scope-selection `gh issue view`
- `CONTRIBUTING.md` updated: six skills, `fix-bug` is repo-local and not in `skills-lock.json`

### Design choice

Parent-issue comment in umbrella wrap-up is **optional** (per your "optionally comment on the parent" suggestion) so the agent doesn't spam tracking issues unless it has useful remainder info. Happy to make that required if you prefer.

daimon-pymclabs:
## Re-review: all items addressed

Checked the fix commit (`1343877`) against the diff rather than the summary. All eight items are genuinely resolved:

| Item | Status |
|---|---|
| 🔴 Round-count fallback | Fixed — explicit jq excluding `fix-bug-fix-summary` / `fix-bug-escalation` / `fix-bug-approved`, with a note on *why* |
| 🔴 Sub-issue detection | Fixed — `gh api "repos/$REPO/issues/N/sub_issues"` as primary, body scan demoted to secondary |
| 🔴 Placeholder expansion | Fixed — `#<scope-issue>` / `#<parent-issue>` throughout, plus an explicit "heredocs use `<<'EOF'`, no expansion" note |
| 🟡 PR body clobbering | Fixed — read-then-merge, scoped to `## Issue links` only |
| 🟡 Approval vs crash | Fixed — `fix-bug-approved` marker, wired into loop control, state table, robustness notes, and the copy-to-another-repo prefix list |
| 🟡 Reviewer allowlist | Fixed — `gh pr view --json body,comments` permitted; anchoring hardened to "diff is truth, disagreement is a 🔴" |
| 🟡 Umbrella remainder | Fixed — new wrap-up section |
| 🟢 Docs / nits | Fixed — CONTRIBUTING says six, `## Issue links` rename applied consistently, `state` added to the issue query, `#408` → `#123` |

Two small things, neither blocking:

**🟢 State-table row ordering.** The new `fix-bug-approved` row sits second, above `Human pushed or commented since last automation activity → escalate`. If an agent matches top-down, an approved PR that a human later commented on (say, requesting changes post-approval) hits the exit row first and terminates silently instead of escalating. Cheap fix: move the human-intervention row to the top of the table, or note that it is evaluated first regardless of position.

**🟢 One unverified claim in CONTRIBUTING.** The new text asserts that `npx skills add` "updates only the lockfile-managed skills; it does not add or remove `fix-bug`". That's the behaviour you'd want, but I haven't confirmed it against the skills CLI. If it hasn't been tested, softening to "is not expected to touch `fix-bug`" avoids putting an untested guarantee in contributor docs.

**Verdict: ready to approve** once `test` goes green (everything else — `prek`, both `zizmor` jobs, docs and dist builds — is passing). Neither 🟢 needs to block the merge; both are one-line follow-ups.

PR #416, opened 2026-08-09
Fix: reject duplicate coefficient labels at parse time

### LINKED ISSUE #152: [MEDIUM] Duplicate coefficient labels silently overwritten in effects extraction
## Summary

In `extract_labeled_draws` (`effects.py`), if two different regression equations use the same label name (e.g., `M ~ a*X; Y ~ a*Z`), the second silently overwrites the first in the `labeled_draws` dict. Any defined parameter using that label (e.g., `indirect := a*b`) would use the wrong coefficient.

## Context

In lavaan, reusing a label constrains the two coefficients to equality. pathmc does not implement equality constraints, so reusing a label is undefined behavior that produces silently incorrect results.

## Fix

Raise an error in the parser (`parse.py`) when a label is reused across equations. This catches the error early with a clear message:

```
DuplicateEquationError: Label 'a' used in both 'M ~ a*X' and 'Y ~ a*Z'. 
Each label must be unique. Use different labels (e.g., 'a1', 'a2') for 
distinct coefficients.
```

## Reference

Technical audit report: `docs/dev/technical_audit_2026_03_16.md` finding M2.

drbenvincent:
## Summary
Raise `DuplicateEquationError` in `parse_spec` when the same coefficient label appears in more than one regression equation, preventing silently wrong defined-parameter evaluation.

## Issue links
- Fixes #152

## Test plan
- [x] `uv run pytest tests/test_parse.py -x -v`
- [x] `make lint`


drbenvincent:
## Fix summary

**Scope**: Fixes #152
**Root cause**: `parse_spec` only checked duplicate LHS variables; reused coefficient labels were allowed and later overwrote each other in `extract_labeled_draws`.
**Approach**: Track seen labels while parsing regressions and raise `DuplicateEquationError` with the two conflicting equations. Also reject duplicate labels within a single equation.
**Not in scope**: lavaan-style equality constraints for shared labels (undefined in pathmc).

drbenvincent:
## Review (independent)

Solid fix for #152: duplicate coefficient labels are now rejected at parse time, matching the only place `extract_labeled_draws` sources labels (`reg.terms`). Error messages name the conflict and suggest distinct labels, consistent with `AGENTS.md`. Scope aligns with the issue and fix summary; intentionally breaking previously silent overwrite behavior is correct.

### Findings

**🟡 Should fix — within-equation duplicate label path untested**

The PR adds a distinct branch for reusing a label twice in one equation (`labels_in_stmt` / `"used more than once in '{stmt}'"`), and the fix summary calls this out explicitly, but `tests/test_parse.py` only covers the cross-equation case (`DUPLICATE_LABEL_SPEC`). Please add a small test, e.g. `parse_spec("Y ~ a*X + a*Z")` expecting `DuplicateEquationError` (or at least a match on `"used more than once"`), so the two error paths don't drift.

### Checked — no issues

- **Correctness**: Root cause addressed; `seen_labels` is updated in the regression branch only, which is the sole label source for `extract_labeled_draws`.
- **Regressions**: No legitimate lavaan-style equality-constraint use case in pathmc; rejecting duplicates is the right behavior.
- **Style / errors**: Type hints, docstring updates, and actionable messages look good.
- **Performance**: O(terms) per equation; negligible.

drbenvincent:
## Fix summary (round 2)

**Scope**: Fixes #152
**Addressed review**: Added `test_duplicate_label_within_equation_raises` for `Y ~ a*X + a*Z` to cover the within-equation duplicate-label branch.

drbenvincent:
Reviewed — no actionable issues found. Approving.

Round 2 re-check: the within-equation duplicate-label test (`test_duplicate_label_within_equation_raises` for `Y ~ a*X + a*Z`) covers the `labels_in_stmt` branch that round 1 flagged. Cross-equation coverage remains via `DUPLICATE_LABEL_SPEC`.

### Checked — no issues

- **Correctness**: Duplicate labels are rejected at parse time in the only source path (`reg.terms`); within-equation duplicates are caught before cross-equation lookup.
- **Scope / regressions**: Aligns with #152 and the fix summary; no legitimate shared-label equality-constraint use case in pathmc.
- **Tests**: Both error paths (`used more than once` vs `used in both`) are now covered; matches existing `pytest.raises(Exception, …)` style in this file.
- **Style / errors**: Docstring updates, actionable messages with fix hints, and O(terms) overhead are all fine.

PR #417, opened 2026-08-09
Fix partial-pooling intercept and residual-block wiring identification

### LINKED ISSUE #217: Pre-existing identification problems exposed by PyMC 6 RNG change: partial-pooling intercepts and residual-cov block wiring

drbenvincent:
## Summary

- Auto-drop the fixed formula `Intercept` when `pooling="partial"` (or dict intercept) is active on panel models, so `mu_alpha` serves as the sole global intercept and sampling no longer diverges.
- Separate estimation vs generative wiring for downstream equations that reference residual-covariance block members: likelihoods use observed block components (identified OLS-scale coefficients) while `mu_{var}` deterministics retain structural means for `do()` propagation.

## Issue links

- Fixes #217

## Test plan

- [x] `make test-fast` (1461 passed)
- [x] `make lint` passed
- [x] Real-sampling check: `beta_Y[M1]` posterior sd collapses from ~9.6 to ~0.1 on parallel-mediators fixture
- [x] Real-sampling check: panel lag `do()` contrast positive with identified partial-pooling parameterization
- [x] Updated partial-pooling tests (warning → auto-drop behavior)
- [x] Pinned oracle posteriors in panel lag fixtures for mock-sampling compatibility

### DIFF
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -235,6 +235,7 @@ class PathModel:
             families=self._families,
             pooling=pooling,
             latent=self._latent,
+            panel_info=panel_info,
         )
         self._priors = merge_priors(defaults, priors)
 
@@ -256,13 +257,17 @@ class PathModel:
                 elif t.variable not in data.columns:
                     missing.append(t.variable)
             if missing:
-                cols = get_predictor_columns(reg)
+                cols = get_predictor_columns(
+                    reg, pooling=pooling, panel_info=panel_info
+                )
                 self._design_matrices[reg.lhs] = nw.from_dict(
                     {c: np.array([], dtype=float) for c in cols},
                     backend=data.implementation,
                 )
             else:
-                self._design_matrices[reg.lhs] = build_design_matrix(reg, data)
+                self._design_matrices[reg.lhs] = build_design_matrix(
+                    reg, data, pooling=pooling, panel_info=panel_info
+                )
 
         self._compile()
 
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -184,7 +184,54 @@ class MuSpec:
     slots: list[PredictorSlot]
 
 
-def get_predictor_columns(reg: Regression) -> list[str]:
+# ---------------------------------------------------------------------------
+# Helpers: pooling / random effects
+# ---------------------------------------------------------------------------
+
+
+def _has_random_intercepts(pooling: str | dict | None) -> bool:
+    """Whether pooling spec requests random intercepts."""
+    if pooling == "partial":
+        return True
+    if isinstance(pooling, dict):
+        return pooling.get("intercept", False)
+    return False
+
+
+def _get_slope_vars(pooling: str | dict | None) -> list[str]:
+    """Extract variables that should get random slopes."""
+    if isinstance(pooling, dict):
+        return list(pooling.get("slopes", []))
+    return []
+
+
+def _drops_formula_intercept(
+    pooling: str | dict | None,
+    panel_info: PanelInfo | None,
+) -> bool:
+    """Whether the fixed formula intercept is absorbed by ``mu_alpha``."""
+    return _has_random_intercepts(pooling) and panel_info is not None
+
+
+def _effective_has_intercept(
+    reg: Regression,
+    pooling: str | dict | None = None,
+    panel_info: PanelInfo | None = None,
+) -> bool:
+    """Whether a regression keeps its formula intercept in the design matrix."""
+    if not reg.has_intercept:
+        return False
+    if _drops_formula_intercept(pooling, panel_info):
+        return False
+    return True
+
+
+def get_predictor_columns(
+    reg: Regression,
+    *,
+    pooling: str | dict | None = None,
+    panel_info: PanelInfo | None = None,
+) -> list[str]:
     """Return *all* predictor column names for a regression equation.
 
     Includes both free and fixed-coefficient predictors.
@@ -193,6 +240,10 @@ def get_predictor_columns(reg: Regression) -> list[str]:
     ----------
     reg : Regression
         Parsed regression with terms and intercept flag.
+    pooling, panel_info
+        When partial pooling with random intercepts is active on a panel
+        model, the fixed ``Intercept`` column is dropped and ``mu_alpha``
+        serves as the global intercept.
 
     Returns
     -------
@@ -200,19 +251,24 @@ def get_predictor_columns(reg: Regression) -> list[str]:
         Column names including ``"Intercept"`` when applicable.
     """
     cols: list[str] = []
-    if reg.has_intercept:
+    if _effective_has_intercept(reg, pooling, panel_info):
         cols.append("Intercept")
     cols.extend(t.variable for t in reg.terms)
     return cols
 
 
-def get_free_predictor_columns(reg: Regression) -> list[str]:
+def get_free_predictor_columns(
+    reg: Regression,
+    *,
+    pooling: str | dict | None = None,
+    panel_info: PanelInfo | None = None,
+) -> list[str]:
     """Return predictor column names that have free (estimated) coefficients.
 
     Fixed-value terms (e.g. ``1*X``) are excluded.
     """
     cols: list[str] = []
-    if reg.has_intercept:
+    if _effective_has_intercept(reg, pooling, panel_info):
         cols.append("Intercept")
     # HSGP terms carry their own basis weights (not a scalar beta column), so
     # they are excluded here to keep ``beta`` sized to the plain/free terms.
@@ -243,7 +299,12 @@ def _term_base_vars(term: Term) -> list[str]:
 # ---------------------------------------------------------------------------
 
 
-def build_mu_specs(spec: Spec) -> dict[str, MuSpec]:
+def build_mu_specs(
+    spec: Spec,
+    *,
+    pooling: str | dict | None = None,
+    panel_info: PanelInfo | None = None,
+) -> dict[str, MuSpec]:
     """Convert Spec regressions into MuSpec intermediate representations.
 
     Pure transformation with no PyMC dependency.  Each ``Regression``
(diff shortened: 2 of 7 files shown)


drbenvincent:
## Fix summary

**Scope**: Fixes #217
**Root cause**: (1) Partial pooling kept both `beta[Intercept]` and free `mu_alpha`, creating an unidentified intercept; (2) downstream equations wired block members through predicted means (`mu_M1`) rather than observed data, inducing exact collinearity with exogenous parents.
**Approach**: Auto-drop the formula intercept when random intercepts are active on panel models; compile separate `mu_est` (observed block wiring for likelihood) and `mu_gen` (structural wiring in `mu_{var}` for `do()`).
**Not in scope**: Non-centered random-intercept parameterization; broader simulate.py refactor beyond the dual-wiring compile change.

drbenvincent:
## Review (round 1)

Solid fix for the two #217 root causes described in the fix summary: intercept auto-drop is threaded consistently through design-matrix construction, `build_mu_specs`, coords, `default_priors`, cross-sectional compile, and scan-panel compile; dual `mu_gen` / `mu_est` wiring for residual-block members is a reasonable split between generative (`do()`) and estimation (likelihood) paths.

### Findings

**🟡 Missing regression test for residual-block dual wiring**

The intercept auto-drop path is well covered (`test_partial_pooling_warning.py`, coord assertions, lag/scan path). The second root cause — downstream equations referencing residual-covariance block members must use observed data in the likelihood but structural means in `mu_{var}` — has **no automated test** in this PR. The PR body cites manual real-sampling checks (`beta_Y[M1]` sd collapse), but nothing in CI will catch a regression in `_references_block_members` / `prefer_observed_block_members`. Please add a focused compile or mock-sampling test (e.g. parallel-mediators + residual block fixture) that asserts:
- likelihood wiring uses observed block members (identified `beta`),
- `mu_{var}` deterministic still references structural means for `do()` propagation.

**🟡 `prefer_observed_block_members` may not cover lag / interaction block references**

`_references_block_members` returns `True` for `lag` and `interaction` slots when a block member appears, which triggers the dual-resolver build, but `prefer_observed_block_members` only overrides plain-name resolution inside `_resolve_var`. If `resolve(slot)` handles `lag` / `interaction` / `transform` slots via a separate code path, `mu_est` and `mu_gen` could be identical for those predictors — leaving the collinearity bug unfixed for `Y ~ lag(M1)`-style equations. Either confirm those slot kinds already resolve to observed data (and add a test), or extend the observed-data override into the lag/interaction resolution paths.

**🟡 Panel fixture posterior pinning weakens test signal**

`test_panel_do.py` and `test_panel_by_time.py` now overwrite the mock-sampled posterior with oracle DGP values after `fit()`. That unblocks `do()` tests under the new identified parameterization, but the tests no longer exercise whether mock/real sampling produces sensible posteriors under auto-dropped intercepts. Consider keeping the pin for `do()` propagation tests but adding at least one test that validates compile-time coords / graph structure without pinning, or a lightweight assertion that `mu_alpha_*` is present and `Intercept` is absent from `{lhs}_predictors` in the lag panel path (partially covered elsewhere, but not tied to the `do()` fixtures).

### 🟢 Nits (non-blocking)

- `tests/test_partial_pooling_warning.py` still has a "warning" filename/module docstring heritage; consider renaming to `test_partial_pooling_intercept.py` in a follow-up.
- `get_predictor_columns` / `get_free_predictor_columns` now need `pooling` + `panel_info` for accurate results; worth a quick audit of any out-of-package callers (not touched in this diff).

### What looks good

- Intercept drop is correctly gated on `_has_random_intercepts(pooling) and panel_info is not None`, so cross-sectional models without panel are unaffected.
- Random intercepts/slopes are added to both `mu_gen` and `mu_est` before splitting likelihood vs deterministic — correct.
- Removing the warn-only path in favor of auto-drop is a better UX and matches the stated scope for #217.
- `_model.py` design-matrix construction now passes `pooling` / `panel_info`, keeping introspection matrices aligned with the compiled model.

drbenvincent:
## Fix summary (round 2)

**Scope**: Fixes #217 — addressing round-1 review findings.
**Changes**:
- Filter block members out of `active_endogenous` when `prefer_observed_block_members=True`, so transforms and interactions also resolve via observed data (not only plain slots).
- Added `TestBlockMemberDualWiring` compile-time graph tests (plain + interaction predictors).
- Added `test_panel_lag_compile_drops_intercept_and_keeps_mu_alpha` without posterior pinning; documented why `panel_lag_model` fixture still pins for `do()` propagation tests.

drbenvincent:
## Review (round 2)

Round-1 findings are addressed. This looks ready to merge.

### Round-1 follow-up — resolved

1. **Residual-block dual wiring** — `TestBlockMemberDualWiring` adds compile-time graph assertions for plain (`Y ~ M1 + M2`) and interaction (`Y ~ M1:M2`) predictors: likelihood `mu` is distinct from `mu_Y`, structural means appear only in the deterministic, and observed wiring is used in the likelihood. Good regression coverage for the #217 collinearity fix.

2. **`prefer_observed_block_members` lag/interaction/transform** — Filtering block members out of `active_endogenous` (and passing that map into `_apply_transform_chain` and interaction `_resolve_var` calls) is the right fix; it covers nested transform chains, not just plain slots. Interaction test exercises the non-plain path.

3. **Panel posterior pinning** — `test_panel_lag_compile_drops_intercept_and_keeps_mu_alpha` validates intercept drop + `mu_alpha_*` presence without pinning; fixture docstring explains why `do()` tests still pin oracle posteriors.

### What still looks good from round 1

- Intercept auto-drop gated on `_has_random_intercepts(pooling) and panel_info is not None`.
- Random intercepts/slopes added to both `mu_gen` and `mu_est` before the likelihood/deterministic split.
- `_model.py` design-matrix construction threads `pooling` / `panel_info` consistently with compile.
- Warn-only path replaced by auto-drop — better UX and matches scope.

### 🟢 Nits (non-blocking, follow-up)

- `introspect.build_priors()` still calls `get_free_predictor_columns(reg)` without `pooling` / `panel_info`; `model.priors()` may over-report intercept-related `beta_*` entries on panel partial-pooling models (e.g. intercept-only formulas). Worth threading `panel_info` in a follow-up.
- No dedicated compile test for transform-on-block-member (`adstock(M1)` etc.); interaction coverage makes regression unlikely, but a transform case would close the loop.
- `tests/test_partial_pooling_warning.py` filename still reflects the old warn-only behavior.

No 🔴 or 🟡 blockers from this round.

Issue #145, opened 2026-03-16
[CRITICAL] Non-Gaussian intermediaries produce wrong results in scan panel models


drbenvincent:
## Summary

Inside the scan step function, non-Gaussian endogenous variables store the **response-scale** value in the carry state:

- Bernoulli: `sigmoid(mu)` — continuous in (0, 1)
- Poisson/NegBinomial: `exp(mu)` — continuous positive rate

In the cross-sectional path, a Bernoulli free RV takes values in {0, 1} and a Poisson free RV takes integer values. These are **structurally different models**.

## Impact

If a Bernoulli variable M is a mediator with downstream dependents in a scan-compiled panel, the downstream equation computes `mu_Y = beta * sigmoid(mu_M)` instead of `beta * M` where M ∈ {0, 1}. The same issue applies to Poisson predictors.

## Scope

- **Affects:** Panel models where a Bernoulli/Poisson variable is used as a predictor in another equation within a scan-compiled model.
- **Does NOT affect:** Models with non-Gaussian terminal outcomes only (no downstream equations). Pure MMM models with all-Gaussian outcomes are unaffected.

## Root cause

Same architectural tension as the noise-free carry state issue: the scan body can't wire through free RVs emitted outside the scan. For non-Gaussian variables, the nonlinear link function makes the approximation much worse — downstream equations see `sigmoid(mu)` instead of `{0, 1}`.

## Reference

Technical audit report: `docs/dev/technical_audit_2026_03_16.md` finding C2.

PR #418, opened 2026-08-09
Fix: stochastic discrete carry for non-Gaussian panel intermediaries

### LINKED ISSUE #191: Track stochastic split-scan compiler for non-Gaussian panel intermediaries
Follow-up to #145. The immediate safeguard should prevent scan-compiled panel models from silently using response-scale means for Bernoulli, Poisson, or NegativeBinomial intermediaries, but a full behavioral fix needs a deeper compiler change.

Investigation notes: PyMC can register and observe random variables produced inside `pytensor.scan`, so a stochastic scan body is viable for logp construction. However, replacing one scan output with `pm.do()` does not rewrite sibling outputs produced by the same scan body, so a one-scan stochastic rewrite would still leave downstream interventions structurally wrong.

Suggested direction: split scan compilation along the contemporaneous topological order or into dependency stages so downstream equations consume named upstream sequence variables rather than sibling outputs hidden inside the same `scan` op. The design should preserve `pm.observe()` for fitting, `pm.do()` for interventions, and response-scale deterministics for Bernoulli/count outcomes.

drbenvincent:
## Summary
- Non-Gaussian endogenous variables used as predictors in scan-compiled panel models now carry **sampled** values through the temporal recursion instead of response-scale means (probabilities/rates).
- Bernoulli mediators use per-timestep uniform innovations; Poisson/NegativeBinomial use a scan-scoped RNG passed as a non-sequence input.
- Removes the compile-time rejection guard added in #145; terminal discrete outcomes are unchanged.

## Issue links
- Fixes #191

## Test plan
- [x] `uv run pytest tests/test_scan_non_gaussian_mediators.py tests/test_panel_smoke.py::TestScanNonGaussianValidation -v`
- [x] `make test-fast` (1470 passed)
- [x] `make lint`

### DIFF
--- a/pathmc/compile.py
+++ b/pathmc/compile.py
@@ -480,7 +480,6 @@ def compile_to_pymc(
     _reject_nan_predictors(data, graph_info)
 
     if panel_info is not None and _has_temporal_deps(spec, graph_info):
-        _validate_scan_non_gaussian_intermediaries(spec, families, latent)
         return _compile_scan_panel(
             spec=spec,
             data=data,
@@ -1409,12 +1408,17 @@ def _scan_term_base_vars(term: Term) -> list[str]:
     return _term_base_vars(term)
 
 
-def _validate_scan_non_gaussian_intermediaries(
+def _scan_discrete_sample_vars(
     spec: Spec,
     families: dict[str, str],
     latent: set[str],
-) -> None:
-    """Reject scan models that would pass response means downstream."""
+) -> set[str]:
+    """Return discrete endogenous vars that must carry sampled values in scan.
+
+    Bernoulli/Poisson/NegativeBinomial nodes used as predictors (including
+    via ``lag()`` or transforms) must propagate *sampled* values through the
+    scan carry, not response-scale means (probabilities or rates).
+    """
     discrete_families = {"bernoulli", "poisson", "negbinomial"}
     non_gaussian = {
         reg.lhs
@@ -1423,31 +1427,15 @@ def _validate_scan_non_gaussian_intermediaries(
         and families.get(reg.lhs, "gaussian") in discrete_families
     }
     if not non_gaussian:
-        return
+        return set()
 
-    downstream: dict[str, set[str]] = {}
+    used_as_predictor: set[str] = set()
     for reg in spec.regressions:
         for term in reg.terms:
             for base_var in _scan_term_base_vars(term):
                 if base_var in non_gaussian:
-                    downstream.setdefault(base_var, set()).add(reg.lhs)
-
-    if not downstream:
-        return
-
-    details = "; ".join(
-        f"'{var}' ({families.get(var, 'gaussian')}) -> {', '.join(sorted(targets))}"
-        for var, targets in sorted(downstream.items())
-    )
-    raise ValueError(
-        "Scan-compiled panel models do not support non-Gaussian endogenous "
-        "variables as predictors in downstream equations. The current scan "
-        "compiler would propagate response means (probabilities or rates) "
-        f"instead of sampled Bernoulli/count values: {details}. Make these "
-        "variables terminal outcomes, remove temporal terms so the "
-        "cross-sectional compiler can be used, or follow #191 for the "
-        "split-scan stochastic compiler work."
-    )
+                    used_as_predictor.add(base_var)
+    return used_as_predictor
 
 
 # Bound applied to the linear predictor ``mu`` before exponentiating it into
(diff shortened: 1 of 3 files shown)


drbenvincent:
## Fix summary

**Scope**: Fixes #191
**Root cause**: The scan step function stored sigmoid(mu) / exp(mu) in the carry state for Bernoulli/Poisson/NegBinomial variables, so downstream equations saw continuous response-scale means instead of discrete samples.
**Approach**: Mirror the existing Gaussian stochastic-carry pattern — identify discrete vars used as predictors via `_scan_discrete_sample_vars`, sample inside `step_fn` (uniform innovations for Bernoulli, scan-scoped RNG for count families), and thread response-scale means through `carry_mu` for the outer likelihood RVs. Removed the #145 rejection guard.
**Not in scope**: Full split-scan stage compilation (only needed if do()-intervention sibling rewriting regresses; the existing `_do_intervene_*` channel from #404 handles contemporaneous/lagged severance within a single scan).

drbenvincent:
## Review (round 1)

Overall the approach is sound: `_scan_discrete_sample_vars` replaces the #145 rejection guard, Bernoulli mediators get per-timestep uniform innovations, count families sample inside `step_fn` via a scan-scoped RNG, and `carry_mu` still threads response-scale means to the outer likelihood RVs while `new_endo` carries sampled values for downstream resolution — mirroring the existing Gaussian stochastic-carry split. Scope matches the fixer summary and the diff.

### 🟡 Should fix — no behavioral regression test for the actual bug

All updated tests only assert **compilation / node presence** (`discrete_uniforms_M`, `mu_N`, `_use_observed_carry`). None verify the semantics that #191/#145 cared about: that downstream `mu` (or equivalent) consumes **sampled** 0/1 / count values rather than response-scale means.

A minimal guard would compile a tiny two-equation panel (e.g. `M ~ X\nY ~ M + lag(Y)`, `families={"M":"bernoulli"}`), evaluate the scan `step_fn` graph (or compare `mu_Y` against hand-built expressions) at fixed inputs, and assert `mu_Y` depends on a Bernoulli **draw** (uniform threshold) and **not** solely on `sigmoid(mu_M)`. Without this, a future refactor could restore mean-propagation while keeping the same node names and the suite would still pass.

Related: `_assert_discrete_sample_nodes` is effectively a no-op for `poisson` / `negbinomial` (only checks `mu_M` / `_use_observed_carry`, same as any scan model). The parametrized non-Bernoulli cases therefore add little beyond “it compiles.”

### 🟢 Nitpicks (non-blocking)

- `TestValidatorStaysSilentWhenSafe` still names the old validator; consider renaming now that rejection is gone.
- Terminal-outcome tests in `test_scan_non_gaussian_mediators.py` could mirror `test_panel_smoke.py` and assert `discrete_uniforms_M` is **absent** when `M` is never used as a predictor.

No other correctness or regression issues spotted in the diff.

drbenvincent:
## Fix summary (round 2)

**Scope**: Fixes #191 — addressed reviewer round 1.
**Changes**: Added `TestDownstreamMuUsesSampledValues` — evaluates `mu_Y` with all-0 vs all-1 `discrete_uniforms_M` and asserts they differ at t=0 (would be identical if still consuming sigmoid(mu_M)). Renamed `TestValidatorStaysSilentWhenSafe` → `TestTerminalDiscreteOutcomesUnchanged`; added negative assertion that terminal scan outcomes do not create `discrete_uniforms_*` nodes.

drbenvincent:
## Review (round 2)

Round 1 feedback is addressed. The new `TestDownstreamMuUsesSampledValues::test_bernoulli_mediator_mu_y_differs_with_uniform_draw` compiles a minimal two-equation panel, evaluates `mu_Y` via `pytensor.function` with all-0 vs all-1 `discrete_uniforms_M`, and asserts `mu_Y` at t=0 differs — which would fail if downstream still consumed `sigmoid(mu_M)` instead of a Bernoulli draw. The round 1 nitpicks are also done: `TestTerminalDiscreteOutcomesUnchanged` rename and the negative `discrete_uniforms_M` assertion on terminal scan outcomes (mirrored in `test_panel_smoke.py`).

### Implementation

The `carry_mu` / `new_endo` split is correct: `carry_mu` threads response-scale means to outer likelihood RVs (`pm.Bernoulli(p=mu_all)`, etc.) while `new_endo` carries sampled 0/1 or count values for downstream resolution within the scan step. Bernoulli uniform innovations, scan-scoped RNG for count families, and observed-carry switching for lagged discrete mediators all mirror the existing Gaussian stochastic-carry pattern. Topological ordering via `endogenous_order` preserves contemporaneous mediator resolution (e.g. `N ~ M` before `Y ~ N`).

### Tests

All 46 related tests pass locally (`test_scan_non_gaussian_mediators.py` + `TestScanNonGaussianValidation`). Scope is tight: rejection guard removed, stochastic carry added, compilation + behavioral regression covered.

### Non-blocking follow-ups (optional)

- `_assert_discrete_sample_nodes` still adds little for `poisson` / `negbinomial` (no uniform node to check); a count-family behavioral analogue (varying scan RNG state) would close that gap but is not required for this fix.
- `pt.random.poisson` / `negative_binomial` emit PyTensor `return_next_rng` deprecation warnings at compile time — harmless now, worth silencing upstream when convenient.

No actionable correctness, regression, or scope issues found. LGTM.

Issue #333, opened 2026-06-22
Revert exog-lag scan-carry workaround once PyTensor #2252 is fixed


drbenvincent:
## Context

PR #331 fixed the zeroed-out lag posteriors reported in #316 by replacing a trivial-echo `scan` sit_sot carry with a pre-computed lagged sequence (a concatenation of the init row and the shifted `pm.Data` tensor). This eliminates the carry structure that triggered a PyTensor scan-merge optimizer bug.

The upstream bug is tracked at **pymc-devs/pytensor#2252**: when two scan nodes with structurally identical inner functions are compiled together, the optimizer can incorrectly alias sit_sot carry outputs if any carry has a trivially-echo update.

## What to do when the upstream fix lands

Once pymc-devs/pytensor#2252 is fixed and a new PyTensor release is available:

1. Bump the minimum PyTensor version in `pyproject.toml` to include the fix.
2. Revert the pre-compute block in `_compile_scan_panel` (compile.py ~line 1395) back to a scan sit_sot carry.
3. Remove the TODO comment and the upstream-link comment from compile.py.
4. Re-run `tests/test_graph_consistency.py` to confirm all cells still pass.

## Reference

- pathmc #316 (original user report)
- pathmc #331 (workaround PR)
- pymc-devs/pytensor#2252 (upstream bug)

PR #435, opened 2026-08-12
Re-adopt pymc-marketing transform delegation and expose MMM variants (#214)

### LINKED ISSUE #214: Re-adopt pymc-marketing transform backends once it supports PyMC 6

drbenvincent:
## Summary

- Evaluated re-adopting `pymc-marketing` for built-in transforms now that upstream ships PyMC 6 support (`1.0.0`).
- **Full hard dependency is blocked**, but not because pathmc requires PyMC 6.2 specifically — `pyproject.toml` pins `pymc>=6.0,<7`. The conflict is `pytensor>=3.1.1` (required since #333/#395 for exog-lag scan carry) vs `pymc-marketing` 1.0.0's `pymc<6.1` pin, which transitively constrains `pytensor<3.1`.
- Added `pathmc/_pmm_backend.py` with lazy xtensor bridging helpers that delegate to `pymc_marketing.mmm.transformers` when a compatible install is present, falling back to vendored pytensor kernels otherwise.
- Exposed additional MMM transform variants through the registry: `delayed_adstock`, `weibull_adstock`, and `michaelis_menten` (the main motivation for re-adoption per #214).
- Refactored `Adstock` onto a shared convolution base class; geometric adstock retains scan `step()` support with the existing non-default `l_max`/`normalize` guard.

Closes #214

## Review fixes (post-Bugbot)

1. **Rectangular panel validation** — `_requires_rectangular_panel()` now triggers for all convolution adstock transforms (`adstock`, `delayed_adstock`, `weibull_adstock`), not just the scan path.
2. **Scan error messages** — `delayed_adstock` / `weibull_adstock` raise a clear vectorized-only error instead of suggesting a default config that still fails.
3. **Theta prior** — `DelayedAdstock.emit_prior` now emits `Beta(2, 5)` for `theta` as documented in `param_specs`.

## Test plan

- [x] `uv run pytest tests/test_adstock_parity.py tests/test_mmm_transforms.py tests/test_transforms_parse.py tests/test_panel_shape_validation.py -v` — 83+ passed
- [x] `make lint` — all hooks pass

### DIFF
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -31,7 +31,7 @@ import xarray as xr
 from narwhals.stable.v1.typing import IntoFrame, IntoFrameT
 
 from pathmc.compile import (
-    _has_temporal_deps,
+    _requires_rectangular_panel,
     build_design_matrix,
     compile_to_pymc,
     get_predictor_columns,
@@ -1861,13 +1861,13 @@ def model(
                 "panel= requires data. Provide data= alongside panel=, "
                 "or omit panel= for data-free DAG exploration."
             )
-        # Only the scan compiler reshapes rows to a dense (n_times,
-        # n_units) grid, so only it needs a rectangular panel. Non-temporal
-        # panel models take the row-wise compiler and may be unbalanced.
+        # The scan compiler and vectorized convolution transforms reshape rows
+        # to a dense (n_times, n_units) grid; only row-wise panel models with
+        # no lag/adstock convolution may be unbalanced.
         panel_info = build_panel_info(
             nw_data,
             panel,
-            require_rectangular=_has_temporal_deps(spec, graph_info),
+            require_rectangular=_requires_rectangular_panel(spec, graph_info),
         )
 
     path_model = PathModel(
--- a/pathmc/_pmm_backend.py
+++ b/pathmc/_pmm_backend.py
@@ -13,11 +13,13 @@
 #   limitations under the License.
 """Optional delegation to ``pymc_marketing.mmm.transformers``.
 
-``pymc-marketing`` 1.0.0 supports PyMC 6.0.x but not the PyMC 6.2+ /
-PyTensor 3.1+ stack pathmc currently requires. When a compatible
-``pymc-marketing`` release is installed, the built-in transforms delegate
-through the xtensor bridging helpers below; otherwise ``pathmc.transforms``
-falls back to vendored pytensor kernels with matching numerics.
+``pymc-marketing`` 1.0.0 pins ``pymc<6.1``, which transitively constrains
+``pytensor<3.1``. pathmc requires ``pytensor>=3.1.1`` for exog-lag scan carry
+(#333), so the two stacks cannot be installed together until upstream relaxes
+its pins. When a compatible ``pymc-marketing`` release is installed, the
+built-in transforms delegate through the xtensor bridging helpers below;
+otherwise ``pathmc.transforms`` falls back to vendored pytensor kernels with
+matching numerics.
 """
 
 from __future__ import annotations
--- a/pathmc/_pmm_backend.py
+++ b/pathmc/_pmm_backend.py
@@ -1,0 +1,158 @@
+#   Copyright 2025 - 2026 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Optional delegation to ``pymc_marketing.mmm.transformers``.
+
+``pymc-marketing`` 1.0.0 supports PyMC 6.0.x but not the PyMC 6.2+ /
+PyTensor 3.1+ stack pathmc currently requires. When a compatible
+``pymc-marketing`` release is installed, the built-in transforms delegate
+through the xtensor bridging helpers below; otherwise ``pathmc.transforms``
+falls back to vendored pytensor kernels with matching numerics.
+"""
+
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import Any
+
+import pytensor.tensor as pt
+from pytensor.xtensor.type import as_xtensor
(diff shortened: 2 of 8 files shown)


drbenvincent:
## Review response

Thanks for the review. Addressed all three findings in e33fc86:

### 1. Panel reshape skips rectangular checks (high)
**Fix:** Added `_requires_rectangular_panel()` in `compile.py` — returns true for any convolution adstock transform (`adstock`, `delayed_adstock`, `weibull_adstock`), not just scan-path temporal deps. `pathmc.model()` now passes this to `build_panel_info(require_rectangular=...)`. Regression test: `test_ragged_panel_with_delayed_adstock_raises`.

### 2. Misleading scan error for variants (medium)
**Fix:** `_ConvAdstockBase.step()` now raises immediately with a vectorized-only message for `DelayedAdstock` / `WeibullAdstock` (no longer suggests switching to default `l_max`/`normalize`).

### 3. Theta prior string not applied (medium)
**Fix:** `DelayedAdstock.emit_prior()` checks for `Beta(2, 5)` in `param_specs` and emits `pm.Beta(..., alpha=2, beta=5)`.

---

**Re: PyMC version pin** — pathmc does **not** require PyMC 6.2 specifically (`pymc>=6.0,<7`). The blocker for hard-dep `pymc-marketing` is the `pytensor>=3.1.1` floor (#333), which is incompatible with `pymc-marketing` 1.0.0's `pymc<6.1` transitive constraint. Updated module docstrings accordingly.

juanitorduz:
@drbenvincent I am thinking about this: will pathmc depend on pymc-marketing, or vice versa? See https://github.com/pymc-labs/pymc-marketing/pull/2642 🫠

drbenvincent:
@juanitorduz Good question — I read your concern less as “this PR will break imports” and more as the broader coupling between the two packages going forward.

For *this* PR specifically, the delegation is lazy and optional (no hard dep on `pymc-marketing`; vendored kernels remain the default), so the usual circular-import pitfalls are largely avoidable. That said, even done carefully it still adds coordination overhead — version pins, parity between vendored vs delegated code, and a fuzzier story for users about who owns what.

I’m very open to dropping the `pymc-marketing` imports from pathmc altogether. PathMC is meant to be domain-general SEM/path analysis; it feels more natural for `pymc-marketing` to sit higher in the stack and depend on PyMC + pathmc (as in #2642), rather than pathmc reaching back upstream for MMM-specific transforms. Not a final call — just exploring options — but that direction resonates with me.

juanitorduz:
ok! at it is a lazy import it seems like a good step :) Thanks for the clarification

juanitorduz (approved):

drbenvincent (commented):
## Review: good to merge (minor nits)

Reviewed `e33fc86` locally — targeted tests pass (101/101), and `prek`/docs/wheel checks are green. Full CI `test` job was still running when I checked (likely slow MCMC), but nothing in the diff suggests it should fail.

### What's solid

- **Core feature (#214):** `delayed_adstock`, `weibull_adstock`, and `michaelis_menten` are registered, kernels have oracle/reference tests, and `DelayedAdstock`'s `Beta(2, 5)` theta prior is wired correctly.
- **Panel validation fix:** `_requires_rectangular_panel()` correctly extends the rectangular-panel guard to all convolution adstocks, not just the scan path. The regression test for ragged panels + `delayed_adstock` is the right check.
- **Refactor is clean:** `_ConvAdstockBase` + shared `_apply_conv_panel` removes duplication without changing geometric adstock behaviour. Geometric adstock still has its scan `step()` with the existing `l_max`/`normalize` guard.
- **pymc-marketing delegation is safe (if mostly inert today):** Lazy imports, no declared dependency, vendored fallbacks always run in CI because the stacks can't co-install yet.

### Minor nits (non-blocking)

1. **Phantom extra in error message** — Weibull CDF fallback says `install pathmc[marketing]`, but no such extra exists in `pyproject.toml`. Should say `pip install pymc-marketing` or similar.
2. **Scan-path limitation** — `delayed_adstock` / `weibull_adstock` are vectorized-only. If a panel model hits the scan compiler (because of `lag()` or geometric `adstock()` anywhere in the spec), these transforms will raise `NotImplementedError` at compile time. Acceptable for v1, but worth documenting eventually. The main MMM case — panel + convolution adstock only — uses the vectorized path and works.
3. **Docstring trimming** — A lot of parameter docs were removed from `Transform` base methods. Small regression vs our docstring norms.
4. **No user-guide update** — New DSL transforms aren't documented in `docs/`. Fine for closing #214 if that's follow-up scope.
5. **Delegation code is currently dead** — `_pmm_backend.py` can't run until pymc-marketing relaxes its pins. Given the thread above, a follow-up to drop it may be preferable to carrying unused complexity.

### Summary

| Area | Status |
|------|--------|
| Correctness (kernels, priors, panel guard) | ✅ |
| Tests | ✅ (101 targeted; CI full suite pending) |
| Lint / types | ✅ (`prek` green) |
| Architecture / coupling | ⚠️ acceptable; revisit per discussion above |
| Docs / error messages | Minor gaps only |

**Recommendation:** merge once CI `test` goes green. None of the nits need to block; the `pathmc[marketing]` typo is the only one worth a one-line fix if we want to tidy before merge.

Issue #437, opened 2026-08-13
Feature: AdjustmentModel — construction, fit, and backdoor estimands


drbenvincent:
## Summary

Add `PathModel.adjustment_model(...)` returning a separate `AdjustmentModel` facade around a **reduced single-equation** `PathModel`. Users can inspect the adjustment set and formula, `fit()`, and compute backdoor-adjusted `ate` / `att` / `atu` / `cate` via the inner model's existing `do()` path.

This is the first shippable slice of #380. It does **not** implement `predictions` / `comparisons` / `slopes` (those are #358, then a follow-up wiring issue).

Parent: #380. Locked decisions live there — do not re-open compiler vs raw PyMC, Bambi, or silent set selection.

### Problem Statement

pathmc can compute backdoor adjustment sets but cannot fit the corresponding outcome regression. Users who want a focused total-effect estimator must leave the package. The compiler already accepts a single equation (`Y ~ X + Z`); this issue makes that handoff explicit, validated, and queryable.

### Solution

A small facade, not a second compiler:

1. Parse a treatment–outcome query from the **original** causal `PathModel`.
2. Resolve and validate a backdoor adjustment set against that DAG.
3. Build a reduced spec (generated or user `formula=`).
4. Construct an ordinary data-bound `PathModel` via `pathmc.model(...)`.
5. Delegate `fit()` and `ate`/`att`/`atu`/`cate` to that inner model.
6. Keep the original `GraphInfo` on the facade so the reduced graph is never mistaken for the causal system.

### User Stories

1. As an analyst, I want `model.adjustment_model("X -> Y")` after specifying a DAG, so that pathmc picks the unique minimal adjustment set and builds `Y ~ X + Z` for me.
2. As an analyst with several minimal sets, I want a clear error listing them, so that I must pass `adjustment_set=` instead of getting a silent scientific choice.
3. As an analyst, I want to pass `adjustment_set=` and/or `formula="Y ~ X + Z + X:Z"`, so that I can use an alternative valid set or a richer outcome model.
4. As an analyst with a data-free DAG, I want to pass `data=` at `adjustment_model()`, so that identification can happen before I attach data.
5. As an analyst, I want `adjusted.fit()` then `adjusted.ate(values=(0, 1))` to return the existing `EstimandResult`, so that structural and adjustment ATEs share one result API.
[...]

Issue #380, opened 2026-08-06
Meta: Native regression adjustment and shared interpret layer


drbenvincent:
## Summary

Add **native regression adjustment** as a second causal estimation strategy in pathmc, plus a **shared interpret layer** (`predictions` / `comparisons` / `slopes`) used by both structural `PathModel` and the new `AdjustmentModel`.

This issue is the epic. Implementation lives in child issues. Competing proposal #370 (Bambi-backed adjustment) remains closed.

## Two estimation strategies

1. **Structural estimation (today):** fit the full system and propagate interventions with `pm.do()`.
2. **Regression adjustment (this epic):** use the causal DAG only for identification, fit one native pathmc/PyMC outcome equation, and standardize posterior predictions over a declared population.

Durable boundary:

> causal DAG → identification → estimation strategy → canonical causal result

## Locked decisions (2026-08-13)

Do not re-litigate these in child issues.

| Decision | Choice |
|---|---|
| Outcome compiler | Reuse pathmc. `AdjustmentModel` wraps a reduced single-equation `PathModel`. Do **not** hand-roll a `pm.Model`. |
| Constructor | `PathModel.adjustment_model("X -> Y")` and `adjustment_model(treatment="X", outcome="Y")`. Data-free parents pass `data=` here. `"X -> M -> Y"` errors and points at structural `effect()`. |
| Scope | Ambitious: ATE/ATT/ATU/CATE convenience **and** `predictions`/`comparisons`/`slopes`. Split across children, not one PR. |
| Interpret layer | #358 is a child of this epic. One shared query layer on **both** `PathModel` and `AdjustmentModel`. Public container is **xarray-backed result objects**, not Polars DataFrames. Averages still collapse to `EstimandResult`. |
| Causal gating | Dual-mode. `ate`/`att`/`atu`/`cate` are validated causal queries. `predictions`/`comparisons`/`slopes` may be associational; results carry causal vs associational metadata. No silent overclaim. |
| Reduced formula | Auto-generate `Y ~ X + Z1 + Z2`. Inherit data, outcome family, and only compatible priors (`sigma_Y` and family extras — **not** `beta_Y`). User may pass `formula=` and `priors=`. Extra pre-treatment covariates allowed if they are not descendants/colliders/latents. Treatment is included even with no direct `X → Y` edge. Panel / lags / `~~` / other equations are **not** inherited. |
[...]

Issue #438, opened 2026-08-13
Feature: AdjustmentModel interpret API (predictions, comparisons, slopes)


drbenvincent:
## Summary

Give `AdjustmentModel` the same `predictions` / `comparisons` / `slopes` methods as `PathModel`, by **delegating** to the inner outcome `PathModel` and attaching **designated-treatment causal metadata**.

Parent: #380.

### Problem Statement

After #358, structural models can answer grid/slope/comparison queries. `AdjustmentModel` would still only have `ate`/`att`/`atu`/`cate`. Users should not learn two interpret APIs. Causal honesty differs: on an adjustment model, a slope for a confounder is associational even though the treatment ATE is identified.

### Solution

Thin wrappers on `AdjustmentModel` that call the inner `outcome_model` interpret methods, then set metadata:

- Interventional **and** `causal=True` only when the varied variable is the designated treatment (and the selected adjustment set remains the one validated at construction).
- Otherwise allow the call (dual-mode) but mark `causal=False` / associational.
- `predictions()` without intervening on the designated treatment is associational.
- Do not reimplement finite differences or xarray arithmetic.

### User Stories

1. As an analyst, I want `adjusted.comparisons(variable="X", contrast=(0, 1), average_by="all")` to match `adjusted.ate(values=(0, 1))` when `X` is the designated treatment.
2. As an analyst, I want `adjusted.slopes(wrt="X")` for a continuous designated treatment.
3. As an analyst, I want `adjusted.slopes(wrt="Z")` to still run when `Z` is an adjustment variable, but the result must not be labelled as a causal effect of `Z`.
4. As an analyst, I want a clear error if I pass a `variable`/`wrt` that is not in the reduced outcome formula.

### Implementation Decisions

- Reuse `InterpretResult` and `EstimandResult` from #358. After delegating, set:

```text
estimator = "regression_adjustment"
interventional = <whatever the inner result had>
identifiable = True   # AdjustmentModel only exists for an identified query
causal = interventional and (varied variable == designated treatment)
```

Also set `adjustment_set` on the facade result if easy (`frozenset` stored on `AdjustmentModel`; copy into `dataset.attrs["adjustment_set"]` as a sorted tuple).
- `outcome` / `variable` / `wrt` default to the designated query when omitted.
- `average_by` / `conditional` semantics are exactly #358; do not fork them.
- Do not reimplement finite differences or xarray arithmetic.

### Testing Decisions

[...]

PR #443, opened 2026-08-13
Add shared interpret query layer on PathModel

### LINKED ISSUE #358: Feature: Shared predictions/comparisons/slopes query layer
## Summary

Add a shared causal **query layer** on structural `PathModel`: `predictions()`, `comparisons()`, and `slopes()`, built on a **unit-preserving** response-mean draw container (`chain`, `draw`, `unit`). Public returns are **xarray-backed result objects** (not Polars). Population averages still collapse to the existing `EstimandResult`.

This issue is now a **child of #380**. It lands the interpret layer on `PathModel` first so `AdjustmentModel` can reuse it in a follow-up. Do not implement `AdjustmentModel` here.

## What changed vs the original #358 write-up

Locked in the #380 design discussion:

- **Return type:** xarray result objects with `.dataset`, not tidy Polars DataFrames.
- **Causal gating:** dual-mode metadata, not a hard `is_identifiable()` refusal. `ate()` today already runs `do()` without an identification check; this layer stays consistent with that.
- **Shared contract:** the same method names and draw container will later be used by `AdjustmentModel`. Design the primitive so the adjustment follow-up can delegate.
- **Plots:** a thin `.plot()` on `InterpretResult` is in scope (matplotlib, like other pathmc results). Do not add Seaborn or copy Bambi's `plot_*` wrappers.
- **`datagrid()`:** required, small helper (pinned below). Defer `effect_profile()`.

### Problem Statement

`ate()` / `cate()` cover two-point average contrasts but not continuous slopes, ratio/lift contrasts, or predictions on an arbitrary covariate grid. Bambi `interpret` and `marginaleffects` have this vocabulary; pathmc should too, with draws and pathmc result types, and without a Bambi dependency.

### Solution

1. Extract a **unit-preserving** response-mean prediction primitive from the current `kind="mean"` `do()` path. Today `run_do_pymc` averages over observation rows for g-computation (`mu.mean(_obs_dims(mu))`). **Do not change that averaging** — existing `do()` / `ate()` must stay bit-compatible. Add a parallel path (flag or new function) that keeps a `unit` dim.
2. Implement `predictions` / `comparisons` / `slopes` on `PathModel` using that primitive.
3. Contrast or finite-difference **draw-wise at unit level**, then standardize over the requested population.
4. Attach metadata: estimand name, variables varied, whether the result is **interventional** (graph surgery / `do()`) or **associational** (grid predictions with no intervention), and `is_identifiable(...)` when a treatment/outcome pair is defined.

### User Stories

1. As an analyst, I want `m.comparisons(outcome="Y", variable="X", contrast=(0, 1), comparison="diff")` to match `m.ate("Y", "X", values=(0, 1))` within mock-sample noise, so the new API is a strict generalization.
2. As an analyst, I want `comparison="ratio"` and `"lift"` as well as `"diff"`.
3. As an analyst, I want `m.slopes(outcome="Y", wrt="X", slope="dydx")` as a finite-difference interventional slope, so continuous treatments have an AME.
4. As an analyst, I want `m.predictions(outcome="Y", newdata=...)` for fitted response means on a grid, marked associational unless an intervention `set=` is supplied.
5. As an analyst, I want `average_by=None` to keep unit-level draws and `average_by="all"` to collapse to `EstimandResult`-shaped `(chain, draw)` when the quantity is a single contrast.
6. As a notebook user, I want `result.plot()` for a simple matplotlib summary of the result object.
7. As a test author, I want most tests to use `mock_pymc_sample` or hand-built xarray draws, not MCMC.

### Proposed API (pinned)

```python
from pathmc import datagrid, InterpretResult, EstimandResult

pred = m.predictions(outcome="Y", set={"X": 1})          # InterpretResult
pred = m.predictions(outcome="Y", newdata=datagrid(df, X=[0, 1], Z=[0, 1, 2]))

comp = m.comparisons(
    outcome="Y",
    variable="X",
    contrast=(0, 1),
    comparison="diff",   # "diff" | "ratio" | "lift"
    conditional=None,    # dict[str, float] scalars only
    average_by="all",    # "all" | None
)

slopes = m.slopes(
    outcome="Y",
    wrt="X",
    slope="dydx",        # "dydx" | "eyex" | "eydx" | "dyex"
    eps=1e-4,
    conditional=None,
    average_by="all",
)

grid = m.datagrid(X=[0, 1], Z=[0, 1, 2])   # uses the fitted model frame
grid = datagrid(df, X=[0, 1])              # module-level equivalent
```

Pinned signatures:

```python
def predictions(
    self,
    outcome: str,
    *,
    set: dict[str, float] | None = None,
    newdata: IntoFrame | None = None,
) -> InterpretResult: ...

def comparisons(
    self,
    outcome: str,
    variable: str,
    *,
    contrast: tuple[float, float] = (0.0, 1.0),
    comparison: Literal["diff", "ratio", "lift"] = "diff",
    conditional: dict[str, float] | None = None,
    average_by: Literal["all"] | None = "all",
) -> EstimandResult | InterpretResult: ...

def slopes(
    self,
    outcome: str,
    wrt: str,
    *,
    slope: Literal["dydx", "eyex", "eydx", "dyex"] = "dydx",
    eps: float = 1e-4,
    conditional: dict[str, float] | None = None,
    average_by: Literal["all"] | None = "all",
) -> EstimandResult | InterpretResult: ...
```

- `comparisons(..., comparison="diff", average_by="all")` with `contrast=(lo, hi)` **is** the `ate(outcome, variable, values=(lo, hi))` estimand (same `EstimandResult` type; draws agree under mock sample).
- `conditional` values must be scalars in this issue (same as `cate()`). A list/array value raises `TypeError` pointing at `predictions(newdata=datagrid(...))`.
- `average_by="all"` (default) collapses `unit` after the draw-wise contrast and returns **`EstimandResult`**.
- `average_by=None` keeps `unit` and returns **`InterpretResult`**.
- `predictions()` always returns **`InterpretResult`** (it is not a contrast).
- One new public class only: **`InterpretResult`**. Do not add `PredictionResult` / `ComparisonResult` / `SlopeResult`.

### Draw contract

Canonical unit-level container:

```text
xarray.Dataset
  data variable: <outcome>   # response-mean (inverse-link applied), not residual noise
  dims: chain, draw, unit
  optional extra dims only when conditional grids require them (document the dim name)
```

After `average_by="all"`: drop `unit` by taking the mean over `unit` **after** the contrast; wrap in `EstimandResult` (`estimand` label `"ATE"` for `diff`, `"ratio"` for `ratio`, `"lift"` for `lift`, `"dydx"` / `"eyex"` / `"eydx"` / `"dyex"` for slopes).

Do **not** average HDI endpoints across units. Contrast first, then average, then summarize.

`kind="predictive"` residual-noise predictions are out of scope. Default is response-mean (`kind="mean"`).

### Causal vs associational metadata (dual-mode, pinned fields)

Same four fields as #437 adds on `EstimandResult`, plus two on `InterpretResult` only:

| Field | Type | PathModel sets |
|---|---|---|
| `estimator` | `str` | `"structural"` |
| `interventional` | `bool` | `True` iff `do()` / `set=` was used |
| `identifiable` | `bool \| None` | `is_identifiable(variable, outcome)` when both are DAG nodes, else `None` |
| `causal` | `bool` | equal to `interventional` on PathModel (intervention under the specified SCM) |
| `quantity` | `str` | `"prediction"` \| `"comparison"` \| `"slope"` (`InterpretResult` only) |
| `variable` | `str \| None` | varied name (`variable` / `wrt` / the `set` key if a single intervention) |

Copy the four shared fields onto `dataset.attrs`. Do not refuse unidentified queries.

If this PR touches `EstimandResult.__init__` before #437, add the four optional kwargs with the defaults in the table (same schema as #437). If #437 already added them, only set values.

`AdjustmentModel` (#438) will set `estimator="regression_adjustment"` and `causal=True` only when the varied variable is the designated treatment. PathModel code must not hardcode `causal is True`; it assigns `causal = interventional` so #438 can overwrite `causal` without forking arithmetic.

### Implementation Decisions (pinned)

- **Module:** `pathmc/interpret.py` holds `InterpretResult`, `datagrid()`, and the PathModel-method implementations (imported/called from `_model.py`). No Bambi imports.
- **Primitive:** add `average_units: bool = True` to `run_do_pymc` in `simulate.py`. Default `True` preserves current `do()` / `ate()` averaging. Interpret calls `average_units=False` and keeps a `unit` dim. Do not change the default.
- **Array interventions:** if `set[var]` is a 1-D ndarray of length `N`, use it as the per-row intervention; if it is a scalar, keep `np.full(N, val)`. Today `run_do_pymc` always does `np.full(N, val)` — fix that here. Required for unit-level `slopes` (`x_i` vs `x_i+eps`).
- **Result type:** exactly one new class, `InterpretResult`, using `_DrawStorageMixin` / `ResultReprMixin`. Public: `.dataset`, `.mean()`, `.hdi()`, `.draws()`, `.plot()`, plus the metadata fields above.
- **`datagrid(data, **cols)`:** cartesian product of the provided column values. Unspecified columns: numeric → column mean, otherwise → mode. Returns pandas. `PathModel.datagrid(**cols)` calls it on the fitted frame. Export `datagrid` and `InterpretResult` from `pathmc/__init__.py`.
- **Finite differences:** unit-level `set={wrt: x}` vs `set={wrt: x+eps}` with `x` the observed (or `newdata`) vector, then apply `slope` scaling draw-wise. Default `dydx`. Document that positivity is local.
- **Panel / `simulate_over="time"`:** out of scope. Raise `NotImplementedError` on panel models with a message to use cross-sectional models or `do(simulate_over="time")`.
- **`ate()`:** do not reimplement via `comparisons` in this issue. Gold-test agreement is enough.
- **`.plot()`:** lazy matplotlib import. Plot the marginal posterior of the outcome data var (mean over `unit` first if that dim exists). Return `matplotlib.figure.Figure`. Accept `ax=`. This is not a unit-level spaghetti plot.
- **API reference:** add `interpret.InterpretResult` (and `datagrid` if it is public) to `great-docs.yml` `reference` → Result objects.

### Testing Decisions

New gate file `tests/test_interpret.py`. Use `mock_pymc_sample` and/or synthetic xarray.

Must cover:

1. Unit-level mean draws have dims `chain`, `draw`, `unit`; `unit` length equals N (or the model frame length).
2. `comparisons(..., comparison="diff", average_by="all")` returns `EstimandResult` and agrees with `ate()` on a simple linear-Gaussian model (mock sample). Prefer identical xarray after averaging over asserting loose numeric tolerance.
3. `average_by=None` returns `InterpretResult` with a `unit` dim.
4. `predictions()` returns `InterpretResult`.
5. Ratio and lift are draw-wise functions of the two potential-outcome arrays, then averaged when `average_by="all"`.
6. Bernoulli (or other inverse-link) example: response-scale contrast is not the treatment coefficient.
7. `slopes(..., slope="dydx")` smoke: finite difference runs and returns finite values; uses per-row `x` and `x+eps` arrays.
8. `predictions` without `set=` has `interventional is False` and `causal is False`; with `set=` both are `True`.
9. `comparisons` records `identifiable` True/False from `is_identifiable` without raising.
10. `datagrid` crossed shape equals the product of supplied unique values; unspecified numeric column is constant (the mean).
11. `conditional={"Z": [0, 1]}` (a list) raises `TypeError`.
12. Panel model raises `NotImplementedError` with a fix message.
13. `.plot()` returns a matplotlib Figure; `Agg` backend; optional `ax=` (mirror `tests/test_falsify.py` plot tests).
14. `run_do_pymc(..., set={"X": array_of_len_N})` uses the per-row values (not `np.full`).
15. No Polars in the public return type. No class named `PredictionResult` / `ComparisonResult`.

### Out of Scope

- `AdjustmentModel` constructor or methods (#380 construction child + interpret wiring child)
- Hard identification refusal
- Polars tidy-draws public API
- Custom callable `comparison=` / `slope=` functions (follow-up unless trivial)
- `effect_profile()` convenience wrapper (follow-up)
- Mediation-aware slopes vs total-effect slopes as a separate API (structural `do()` already respects the DAG; do not add a second mediation-specific slopes entrypoint)
- Seaborn, Bambi
- Docs site pages (#439); docstring on the new methods is required
- `AdjustmentModel` interpret wrappers (#438)

### Blocked by

Nothing. Can proceed in parallel with #437. Parent: #380. Consumer: #438.

drbenvincent:
## Summary

- Add `pathmc/interpret.py` with `InterpretResult`, `datagrid()`, and unit-preserving `predictions` / `comparisons` / `slopes` wired on `PathModel`.
- Extend `run_do_pymc(..., average_units=True)` (default preserves `do()` / `ate()` averaging) and support per-row ndarray interventions in `set=`.
- Add causal metadata fields on `EstimandResult` and `InterpretResult`; `comparisons(..., comparison="diff", average_by="all")` matches `ate()` draws.

Part of #380. Fixes #358.

Note: #437 (`AdjustmentModel` / `EstimandResult` metadata) may land in parallel and also touch `_model.py` — both changes should be kept on rebase.

## Test plan

- [x] `uv run pytest tests/test_interpret.py tests/test_do.py tests/test_estimand_result_api.py -x -v`
- [x] `make test-fast`
- [x] `make lint`


drbenvincent (commented):
## Spec fidelity (InterpretResult / PathModel interpret layer)

**Met**

- Single public result class: `InterpretResult` only (`PredictionResult` / `ComparisonResult` absent; test enforces).
- Public returns are xarray-backed (`InterpretResult`, `EstimandResult`); no Polars in return types.
- `run_do_pymc(..., average_units=True)` remains the default; `_run_do()` does not pass `average_units=False`, so existing `do()` / `ate()` averaging path is preserved. `comparisons(..., comparison="diff", average_by="all")` matches `ate()` on the fork fixture (`xr.testing.assert_allclose`).
- Per-row array interventions: `_intervention_array` / `_broadcast_intervention` plus `test_run_do_pymc_array_intervention_per_row`.
- Return types: `predictions()` → `InterpretResult`; `comparisons` / `slopes` with `average_by="all"` → `EstimandResult`, `average_by=None` → `InterpretResult` with `unit` dim.
- `datagrid()` (module + `PathModel.datagrid`) returns pandas; crossed product and unspecified numeric → mean covered in tests.
- Scalar-only `conditional`: list values raise `TypeError` pointing at `datagrid` (tested).
- Metadata on `InterpretResult` and `dataset.attrs`: `estimator`, `interventional`, `identifiable`, `causal`, `quantity`, `variable`. Associational vs interventional predictions set `interventional`/`causal` correctly (tested).
- No `AdjustmentModel` in the diff.

**Violations**

1. **Pinned causal assignment pattern** — Implementation decisions require PathModel code to set `causal = interventional`, not hardcode `causal=True`, so #438 can override `causal` without forking arithmetic. `_wrap_interpret` follows this (`causal = interventional`), but `_wrap_estimand` passes `interventional=True, causal=True` as separate literals. Behavior is correct on PathModel today (both `True`), but the hardcoded assignment does not match the pinned contract.

## Tests vs issue checklist

Most checklist items are covered (22 tests pass locally). Gaps:

2. **Checklist #5 (ratio / lift)** — Tests only assert finite `EstimandResult` draws. The issue requires verifying that ratio/lift are **draw-wise** functions of the two potential-outcome unit arrays, then averaged over `unit` when `average_by="all"` (contrast before average, not average-before-contrast). A synthetic-xarray or hand-built unit-level check would close this.

Minor (not blocking alone): panel `NotImplementedError` is only exercised for `predictions()`, not `comparisons()` / `slopes()` (same guard exists in code).

## Verdict

**Request changes** — implementation is close and the core interpret layer matches the pinned API, but please fix the `_wrap_estimand` causal assignment pattern and add checklist #5 coverage for ratio/lift ordering before merge.

drbenvincent (commented):
## Round 2 review — spec fidelity & tests

### Round 1 fixes (resolved)

- **`causal = interventional` in `_wrap_estimand`** — fixed; `_wrap_estimand` now mirrors `_wrap_interpret` (`causal = interventional` before constructing `EstimandResult`).
- **Checklist #5 (ratio/lift ordering)** — `test_ratio_and_lift_contrast_before_average` verifies draw-wise `_apply_comparison` then `.mean("unit")`, matching the pinned contrast-before-average contract.
- **Panel `NotImplementedError` coverage** — `test_comparisons_raises` and `test_slopes_raises` added alongside `test_predictions_raises`.

### Tests vs issue checklist

All 15 pinned testing-decision items are covered; `uv run pytest tests/test_interpret.py tests/test_do.py tests/test_estimand_result_api.py -x -v` passes (64 tests).

### Remaining spec gap (blocking)

**`predictions(..., newdata=...)` is broken** — user story 4 and the pinned API example both require this:

```python
pred = m.predictions(outcome="Y", newdata=datagrid(df, X=[0, 1], Z=[0, 1, 2]))
```

Any call with `newdata` sets `swap_data=True`, and `_temporary_query_data` calls `pm.set_data` on every column present in `gen_model.named_vars`, including endogenous nodes like `X` that are `TensorVariable`s, not `pm.Data` `SharedVariable`s. Result: `TypeError: The variable X must be a SharedVariable`.

Even bypassing `swap_data`, a grid shorter than the fitted `N` fails inside `run_do_pymc` (`SpecifyShape: dim 0 of input has shape 4, expected 100`) because the generative model is compiled to the fit-time observation count.

Suggested direction (not prescriptive): restrict `_temporary_query_data` to `SharedVariable` columns only, and either (a) support arbitrary-length grids by rebuilding/resizing the generative evaluation path, or (b) document and enforce `len(newdata) == len(fitted_frame)` until grid sizing is implemented. Add an integration test for the pinned `datagrid` + `predictions` example.

### Minor (non-blocking)

- `import matplotlib.axes` at module import time; spec asked for lazy matplotlib in `.plot()` only.
- No test that `PathModel.datagrid(**cols)` delegates (module-level `datagrid` is tested).
- `predictions(newdata=...)` and `slopes` alternate kinds (`eyex`, etc.) are untested beyond `dydx`; acceptable per checklist.

### Verdict

**Request changes** — round 1 feedback is fully addressed and the interpret layer matches the pinned API for comparisons/slopes/metadata, but `predictions(newdata=...)` must work (or the public API must be narrowed) before merge.

drbenvincent (commented):
## Round 3 review — spec fidelity & tests

### Round 2 fix (resolved)

**`predictions(..., newdata=...)`** — commit 338a6c3 recompiles on the query frame via `_temporary_query_frame` instead of `pm.set_data` on endogenous vars. `test_newdata_datagrid` exercises the pinned `datagrid` + `predictions` example and asserts `unit` length matches the grid (6 rows).

### Spec fidelity

All pinned API items are met:

- Single public result class (`InterpretResult`); xarray-backed returns; `datagrid` module + `PathModel.datagrid`.
- `run_do_pymc(..., average_units=True)` default unchanged; interpret path uses `average_units=False` with `unit` dim.
- Per-row ndarray interventions via `_intervention_array` / `_broadcast_intervention`.
- Return types: `predictions` → `InterpretResult`; `comparisons`/`slopes` with `average_by="all"` → `EstimandResult`, `average_by=None` → `InterpretResult`.
- `comparisons(..., comparison="diff", average_by="all")` matches `ate()` (`xr.testing.assert_allclose`).
- Ratio/lift: draw-wise contrast before unit average (`test_ratio_and_lift_contrast_before_average`).
- Metadata: `estimator`, `interventional`, `identifiable`, `causal`, `quantity`, `variable` on results and `dataset.attrs`; `causal = interventional` in both `_wrap_interpret` and `_wrap_estimand`.
- Scalar-only `conditional` with `TypeError` pointing at `datagrid`; panel `NotImplementedError` on all three methods.
- Public exports: `InterpretResult`, `datagrid` in `pathmc.__init__.py` and API reference.

### Tests vs issue checklist

All 15 pinned testing-decision items covered in `tests/test_interpret.py`. Gate suite passes locally: 66 tests (`test_interpret`, `test_do`, `test_estimand_result_api`).

### Minor (non-blocking)

- `import matplotlib.axes` at module import time; `pyplot` is lazy inside `.plot()` only.

### Verdict

**Approve** (spec met, checklist complete). GitHub blocks self-approval on this PR; merge-ready from a spec/test perspective.

PR #454, opened 2026-08-13
Validate the bin count in `FalsificationResult.plot`, delegate the rest (#284)

### LINKED ISSUE #284: Polish: validate non-integer plot(bins) and add a few falsification edge-case tests
### Summary

A small grab-bag of low-risk polish items for the falsification feature (good first issue).

### 1. `FalsificationResult.plot(bins=...)` non-integer validation

`pathmc/falsify.py:307` rejects `bins < 1` but accepts a non-integer like `2.5`, which then fails later inside matplotlib with a less clear error. The type hint already says `int | None`.

```python
if bins is not None and (not isinstance(bins, int) or bins < 1):
    raise ValueError(f"bins must be a positive integer or None, got {bins!r}.")
```

### 2. A few explicit edge-case tests worth locking in

The suite (`tests/test_falsify.py`) is thorough, but these explicit cases are nice to pin:

- A 3-variable residual-covariance block (`X ~~ Y`, `Y ~~ Z`) is rejected (currently covered only transitively by the `residual_blocks` check).
- `plot(bins=2.5)` raises a clear `ValueError` (after fix #1).

### Acceptance criteria

- [ ] `plot(bins=2.5)` raises a clear pathmc `ValueError`.
- [ ] `plot(bins=0)`, `bins=-1` still raise (unchanged); `bins=None`, `bins=10` still work.
- [ ] New tests added under `tests/test_falsify.py` (do not modify existing tests' intent).
- [ ] `make lint` clean.

### Notes for picker

- Module: `pathmc/falsify.py`; tests: `tests/test_falsify.py`.
- Run `uv run pytest tests/test_falsify.py -q` and `make lint`.

anevolbap:
Closes #284.

`plot()` guarded `bins` with a bare `bins < 1`. A float like 2.5 passed that check and failed later inside matplotlib, which is the reported symptom. The same expression also raises `TypeError: '<' not supported between instances of 'str' and 'int'` for a binning strategy name, and likewise for a list or tuple of bin edges, so neither ever reached matplotlib even though `Axes.hist` accepts both.

The fix checks only what the bin count owns:

- a non-positive integer raises `ValueError`, as before
- a type `Axes.hist` cannot take at all (float, bool, set) raises `ValueError`
- a strategy name or an edge sequence passes through, and matplotlib validates it

Delegating the last case keeps matplotlib's more precise reporting: `bins="nonsense"` gives `'nonsense' is not a valid estimator for bins`, which re-implementing the rules here would lose.

The annotation widens to `int | str | Sequence[float] | np.ndarray | None` to match what the parameter accepts.

Tests cover the rejected types, the non-positive counts, the accepted counts including `np.int64`, the strategy names and edge sequences that previously raised `TypeError`, and matplotlib's estimator error. Also adds the three-variable `~~` block case from the issue.

One follow-up question: `refute.py:395` guards its own `bins` with the same bare `bins < 1`, and `interpret.py:214` declares `bins: int | None` but passes it to `Axes.hist` unvalidated. Worth a separate issue, or out of scope?


drbenvincent (approved):
## Standards

Clean, focused change. The validation logic is well-scoped: pathmc owns the integer bin-count contract, matplotlib owns strategy names and edge sequences. The explicit `bool` guard is the right call (`True` subclasses `int`), and accepting `np.integer` is a nice touch beyond plain `int`.

Docstring and type annotation now match actual behaviour. Error messages name the problem and suggest the fix. Tests are parametrized and readable; the shared `result` fixture avoids repetition without over-abstracting.

One tiny nit (non-blocking): `test_invalid_bins_type_rejected` matches `"bins must be a positive integer"`, which is a substring of the longer type-error message rather than the exact wording. Fine for now, but a tighter `match` on the full type-error string would make the test more precise.

Lint passes locally (`make lint`).

## Spec

Closes #284 completely:

- `plot(bins=2.5)` → clear `ValueError` before matplotlib
- `bins=0` / `bins=-1` still rejected; `None` / `10` still work
- Three-variable `~~` block test pinned explicitly
- Bonus coverage: strategy names, edge sequences, `np.int64`, matplotlib's estimator error — all sensible extensions of the same bug surface

The issue's suggested fix (`not isinstance(bins, int)`) would have rejected valid `np.int64` inputs; this implementation is strictly better.

## Follow-up

Agree with the PR description: the same bare `bins < 1` pattern in `refute.py` and the unvalidated `bins` passthrough in `interpret.py` are worth a separate issue rather than expanding this PR.

**Summary:** 1 minor nit (Standards), 0 blockers (Spec). LGTM.

anevolbap:
Filed the follow-up as #460, implemented in #461 (draft, stacked on this branch).

daimon-pymclabs (approved):
## Follow-up review

Re-reviewed at `365a212` after the 2026-08-16 approval. The one nit from that review is addressed, the code still holds against current `main`, and the only thing standing between this and merge is CI.

### Nit resolved

`02da81e` splits the old shared `match="bins"` into one anchored pattern per error path:

- `test_non_positive_bin_count_rejected` → `^bins must be a positive integer, got {bins}\.$`
- `test_invalid_bins_type_rejected` → anchored on the full type-error prefix

That was the exact concern last time (a substring match that a type error would also satisfy). Both messages are now pinned to their own path, so the two branches can't silently swap.

### Drift check against `main`

The branch is behind `main` by a handful of commits (`main` is now at #471), but the delta this PR actually introduces is still only `pathmc/falsify.py` (+29/-5) and `tests/test_falsify.py` (+47/-4). Nothing on `main` has touched `falsify.plot` or the bins path since the 2026-09-12 merge, GitHub reports `MERGEABLE`, and the validation logic reads the same as it did at approval:

- `bool` rejected before the `int` branch (still correct, `True` is an `int`)
- `np.integer` accepted alongside `int`
- `str` passes the `Sequence` check, so strategy names delegate as intended
- `np.ndarray` listed explicitly because it isn't a `Sequence` ABC

### The one blocker: CI never finished

The `test` job for `365a212` has been `in_progress` since 2026-09-12T18:13Z, i.e. about a week. It isn't failing, it's stuck. Every other check on the head commit is green (`prek`, `Build Docs`, `zizmor`, dists), but there is no completed test run on this head, so the new parametrized tests have never actually been observed to pass in CI. I could not run them locally either (this sandbox has no `matplotlib`/`pymc`), so this review is a read of the diff plus the check statuses, not an execution.

Please push a rebase onto current `main` (or re-run the workflow) and merge on green.

### Minor, non-blocking

1. **Docstring `Raises` section.** It lists only `ValueError`. For the delegated cases the exception comes from matplotlib and isn't guaranteed to be a `ValueError` for every malformed sequence. Either widen the wording to "whatever `Axes.hist` raises" or leave it, given the surrounding sentence already says matplotlib validates those.
2. **Figures aren't closed.** The parametrized tests create roughly eight new unclosed figures. This matches what the rest of `TestResultDisplay` already does, so it's consistent rather than a regression, but it does move the module closer to matplotlib's 20-open-figure warning. A `plt.close(fig)` in the `result` fixture's teardown would cost one line.

### Verdict

Approving again. Merge once the `test` job completes green on a fresh head.

### Related

#461 (the `refute.py` / `interpret.py` follow-up, filed as #460) is still a draft stacked on this branch, so it'll want a rebase after this lands.

drbenvincent:
Sorry about the delay @anevolbap - I've not had much time to dedicate to open source recently

Issue #465, opened 2026-08-21
Adjustment models silently drop parent priors set via set_priors() and transform-parameter priors


juanitorduz:
**Version:** pathmc 0.3.0 (branch `fix/463-adjustment-beta-prior-override`), pymc 6.2.0, Python 3.12, macOS

Found while fixing #463. That issue is about the `beta_{outcome}` guard being unescapable; this one is about two neighbouring paths where a parent's custom prior is silently replaced by a default with **no guard and no warning at all**. Both are confirmed by running the snippets below.

## 1. `set_priors()` beta overrides bypass the guard entirely

The guard reads only `parent._construction["priors"]` — the original `model()` argument:

```python
construction_priors = (
    parent._construction.get("priors") if parent._construction else None
)
```

A beta prior set after construction never appears there, so it passes the guard and is then not inherited:

```python
m = pathmc.model("X ~ Z\nY ~ X + Z", data=df)
m.set_priors({"beta_Y": Prior("Normal", mu=0, sigma=2.0)})
m._priors["beta_Y"]                                   # Normal(mu=0, sigma=2.0)
adj = m.adjustment_model("X -> Y")                    # no raise
adj.outcome_model._priors["beta_Y"]                   # Normal(mu=0, sigma=10)  <- silent default
```

Passing the same prior to `model(priors=...)` instead raises. Same user intent, opposite behaviour.

## 2. Transform-parameter priors are dropped under a custom `formula=`

`_validate_reduced_spec` rejects `lag()` and `hsgp()` terms but not transforms, so a reduced formula may legitimately contain `adstock()` / `logistic_saturation()`. `_inherit_outcome_priors` only copies `_OUTCOME_PRIOR_SUFFIXES` (`sigma`, `nu`, `alpha_disp`), so transform params fall back to defaults:

```python
spec = "X ~ Z\nY ~ logistic_saturation(X, lam=lam_x) + Z"
m = pathmc.model(spec, data=df, priors={"lam_x": Prior("HalfNormal", sigma=7.0)})
m._priors["lam_x"]                                    # HalfNormal(sigma=7.0)

adj = m.adjustment_model(
    "X -> Y", formula="Y ~ logistic_saturation(X, lam=lam_x) + Z"
)
adj.outcome_model._priors["lam_x"]                    # HalfNormal(sigma=1)  <- silent default
```

Unlike the beta case, a transform parameter has no shape dependence on the predictor set, so there is no structural reason not to inherit it.

[...]

PR #466, opened 2026-08-21
Make the adjustment_model() priors= escape hatch reachable

### LINKED ISSUE #463: `adjustment_model()` raises on any parent with a custom `beta_{outcome}` prior; the error's suggested `priors=` workaround is unreachable
**Version:** pathmc 0.3.0, pymc 6.0.1, Python 3.14, macOS

## What happens

`PathModel.adjustment_model()` raises `ValueError` whenever the *structural* model was built with a custom prior on `beta_{outcome}`. The message advises passing `priors=` to `adjustment_model()`, but doing so raises the identical error, because the guard inspects the parent's priors and runs before the user's `priors=` argument is ever consulted.

## Reproducer

```python
import numpy as np, pandas as pd, pathmc
from pathmc import Prior

n = 200
Z = rng.normal(size=n)
X = 0.6 * Z + rng.normal(scale=0.5, size=n)
Y = 0.5 * X + 0.8 * Z + rng.normal(scale=0.5, size=n)
df = pd.DataFrame({"X": X, "Y": Y, "Z": Z})
spec = "X ~ Z\nY ~ X + Z"

# Works: parent has default priors
pathmc.model(spec, data=df).adjustment_model("X -> Y")            # -> AdjustmentModel

# Fails: parent has a custom beta_Y prior
m = pathmc.model(spec, data=df, priors={"beta_Y": Prior("Normal", mu=0, sigma=2)})
m.adjustment_model("X -> Y")                                       # -> ValueError
m.adjustment_model("X -> Y", priors={"beta_Y": Prior("Normal", mu=0, sigma=2)})  # -> same ValueError
```

Both of the last two calls produce:

```
ValueError: The structural model has a custom prior on 'beta_Y'. Adjustment models do
not inherit coefficient priors. Pass priors= on adjustment_model() for the reduced equation.
```

## Cause

In `pathmc/adjustment.py`, `construction_priors` is read from the **parent** (line 395) and passed to `_inherit_outcome_priors` (line 400), which raises at line 237 if `beta_{outcome}` is present:

```python
construction_priors = (
    parent._construction.get("priors") if parent._construction else None
)
inherited_priors = _inherit_outcome_priors(
    parent._priors,
    construction_priors,
    outcome_name,
)
```

The user's `priors=` argument is only merged at lines 408 to 409, after the raise:

```python
merged_priors = merge_priors(reduced_defaults, inherited_priors)
if priors is not None:
    merged_priors = merge_priors(merged_priors, priors)
```

So the escape hatch the error message names can never be reached.

## Expected

Passing `priors={"beta_{outcome}": ...}` to `adjustment_model()` should satisfy the guard and be used for the reduced equation.

Alternatively, if coefficient priors genuinely must never be inherited, drop the guard and simply do not inherit them. The reduced equation's predictor set can legitimately differ from the structural one, and appropriate defaults are already computed at line 407 via `default_priors(reduced_spec, ...)`.

## Workaround

Build the reduced equation as its own model:

```python
reduced = pathmc.model("Y ~ X + Z", data=df, priors={"beta_Y": ..., "sigma_Y": ...})
```

---

Found while porting [Introduction to Causal Inference with Probabilistic Programming Languages](https://juanitorduz.github.io/intro_causal_inference_ppl_pymc/) to PathMC. That post fixes specific coefficient priors in order to compare against a hand-rolled PyMC model, so matching priors on both the structural and the adjustment model is exactly the use case that hits this. Thanks for the library, the 0.3.0 adjustment and interpret layers are a great addition.

juanitorduz:
## Summary

`PathModel.adjustment_model()` raised on any structural parent built with a custom `beta_{outcome}` prior, and the `priors=` workaround its own error message advertised raised the identical error. The guard lived inside `_inherit_outcome_priors`, which runs before the caller's `priors=` argument is merged, so the escape hatch was unreachable.

The guard is now its own helper, `_reject_uninheritable_beta_prior(construction_priors, user_priors, outcome)`, called from `from_path_model` with the user's overrides in hand. It returns early when `beta_{outcome}` is supplied, so that prior is used for the reduced equation. `_inherit_outcome_priors` lost the guard and the `construction_priors` parameter it only needed for it, leaving it a pure dispersion-prior copier.

Coefficient priors are still never inherited — `beta` is a vector over the outcome's predictors and a reduced equation's predictor set can legitimately differ from the structural one — so the guard still fires when no override is passed. Its message now names the concrete key to pass.

## Issue links

- Fixes #463
- Follow-up filed as #465 for two neighbouring silent-drop paths found while fixing this (a `beta` set via `set_priors()` bypasses the guard entirely, and transform-parameter priors are dropped under a custom `formula=`). Both are out of scope here by design.

## Test plan

- [x] Three cases added to `TestPriorInheritance`; the two escape-hatch cases were confirmed red against the unmodified implementation
- [x] No existing assertion changed — `test_beta_override_on_parent_raises` still passes as-is
- [x] Targeted: `uv run pytest tests/test_adjustment.py tests/test_adjustment_interpret.py tests/test_priors.py` — 89 passed
- [x] `make test-fast` — 1556 passed, 2 skipped
- [x] `make lint` — clean

Verified against pymc 6.2.0 / pytensor 3.2.4 / arviz 1.2.0 (the reporter was on pymc 6.0.1).

### DIFF
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -876,8 +876,11 @@ class PathModel:
         families : dict[str, str] | None
             Per-variable families for the reduced model only.
         priors : dict | None
-            Prior overrides for the reduced model, merged with inherited
-            outcome dispersion priors.
+            Prior overrides for the reduced equation, merged with the
+            outcome dispersion priors inherited from this model.
+            Coefficient priors are never inherited, since the reduced
+            predictor set can differ from the structural one — pass
+            ``beta_{outcome}`` here to set it on the reduced equation.
 
         Returns
         -------
--- a/pathmc/adjustment.py
+++ b/pathmc/adjustment.py
@@ -227,20 +227,31 @@ def _validate_reduced_spec(
             )
 
 
+def _reject_uninheritable_beta_prior(
+    construction_priors: dict[str, Any] | None,
+    user_priors: dict[str, Any] | None,
+    outcome: str,
+) -> None:
+    """Raise when a parent beta prior would be silently replaced by defaults."""
+    beta_key = f"beta_{outcome}"
+    if not construction_priors or beta_key not in construction_priors:
+        return
+    if user_priors is not None and beta_key in user_priors:
+        return
+    raise ValueError(
+        f"The structural model has a custom prior on '{beta_key}'. "
+        f"Adjustment models do not inherit coefficient priors because the "
+        f"reduced predictor set can differ. Pass "
+        f"priors={{'{beta_key}': ...}} to adjustment_model() to set it on "
+        f"the reduced equation."
+    )
+
+
 def _inherit_outcome_priors(
     parent_priors: dict[str, Any],
-    construction_priors: dict[str, Any] | None,
     outcome: str,
 ) -> dict[str, Any]:
-    """Copy outcome dispersion priors from the parent; reject beta overrides."""
-    beta_key = f"beta_{outcome}"
-    if construction_priors and beta_key in construction_priors:
-        raise ValueError(
-            f"The structural model has a custom prior on '{beta_key}'. "
-            f"Adjustment models do not inherit coefficient priors. Pass "
-            f"priors= on adjustment_model() for the reduced equation."
-        )
-
+    """Copy outcome dispersion priors from the parent."""
     inherited: dict[str, Any] = {}
     for suffix in _OUTCOME_PRIOR_SUFFIXES:
         key = f"{suffix}_{outcome}"
@@ -395,11 +406,12 @@ class AdjustmentModel:
         construction_priors = (
             parent._construction.get("priors") if parent._construction else None
         )
-        inherited_priors = _inherit_outcome_priors(
-            parent._priors,
+        _reject_uninheritable_beta_prior(
             construction_priors,
+            priors,
             outcome_name,
         )
+        inherited_priors = _inherit_outcome_priors(parent._priors, outcome_name)
         reduced_defaults = default_priors(
             reduced_spec,
             families=outcome_families or None,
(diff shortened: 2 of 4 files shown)


daimon-pymclabs:
## Review

Reviewed statically against `3bed387`. I read the code but did not execute the suite (no pymc available in my environment), so the reported `89 passed` / `1556 passed` numbers are unverified on my side — CI closes that gap.

**The core change is correct.** Splitting the guard out of `_inherit_outcome_priors` and calling it from `from_path_model` with `priors` in scope is the minimal correct fix: the old code raised unconditionally before `merge_priors(merged_priors, priors)` ever ran, so the advertised workaround could not fire. `_inherit_outcome_priors` losing `construction_priors` is a real simplification rather than churn, and it is the only other caller-visible change. `_OUTCOME_PRIOR_SUFFIXES` is `("sigma", "nu", "alpha_disp")`, so `beta` was never on the inherit path and the guard is the only thing whose behavior changed.

Mergeable as-is. The items below are follow-ups, not merge conditions.

### 1. The `_construction` hole is wider than #465 records

The guard reads `parent._construction["priors"]`, and `_construction` is only populated by `pathmc.model()` (`_model.py:2099`). A `PathModel` built by direct construction leaves it `None`, so a custom `beta_{outcome}` passed to the constructor sails past the guard and is silently replaced by the default — the same failure mode as the `set_priors()` case, which #465 does document. Worth adding a line to #465 so the eventual fix covers both entry points.

Checking `parent._priors[beta_key]` against the default prior instead of consulting `_construction` would close all of these at once, but that changes the guard's source of truth and is correctly out of scope here.

### 2. The message-format test is weaker than the change it covers

The point of the message rewrite is that it now names the concrete key to pass. All three raise-path tests use `match="beta_Y"`, which the *old* message also satisfied, so none of them pin the improvement. Tightening `test_parent_beta_override_without_user_beta_still_raises` to something like:

```python
with pytest.raises(ValueError, match=r"priors=\{'beta_Y'"):
```

would actually assert the new behavior.

### 3. Minor

`adjustment_model()`'s docstring gained the prior-inheritance prose but still has no `Raises` section, and this `ValueError` is now documented user-facing behavior.

### On the deferrals

Punting `set_priors()` and the transform-parameter drop to #465 is the right call — fixing them here would mean changing the guard's source of truth, which carries its own blast radius.

drbenvincent:
Quick reminder ping @juanitorduz

PR #468, opened 2026-08-24
Meta #426: MMM generative workflows as stress-test for panel + simulate APIs

drbenvincent:
Closes #427, closes #428, closes #429, closes #430, closes #431, closes #432, closes #433. Implements Phase 1 of #88 (tracked separately there). Meta: #426.

## Summary

MMM/funnel generative workflows as an integration stress-test for pathmc's panel + simulate APIs. One commit-checkpoint per issue, each reviewed in-loop; full run report with per-issue evidence is on the [tracking issue](https://github.com/pymc-labs/pathmc/issues/426).

### `simulate()` stack
- **#427 — panel support:** `simulate()` gains `panel=`/`pooling=` forwarded to `compile_to_pymc()` like `model()`. Scan-compiled specs forward-simulate in one draw over the `pm.do`-fixed model; stochastic carry states are reconstructed as `mu + sigma*innovations` so simulated chains stay DAG-consistent for simulate-and-recover.
- **#428 — transforms:** gate tests proving adstock/saturation chains round-trip exactly through panel `simulate()` (zero-noise match vs a NumPy reference using pathmc's actual saturation formula, atol 1e-8); unreachable `Adstock._apply_pymc_panel` deleted.
- **#429 — parameter discovery:** new `pathmc.simulate_params_template(spec, data, panel=, pooling=, families=, latent=)` returns `{name: {kind, shape, dtype}}` from the same zero-filled placeholder compile `simulate()` uses.
- **#434 — residual covariances + HSGP:** cross-sectional `~~` blocks draw correlated residuals from packed `chol_{block}`; cross-sectional `hsgp()` terms flow through the existing `pm.do` path. Both `NotImplementedError` guards removed (panel combos stay rejected).

### Panel expressiveness
- **#430 — multi-dimensional units:** `panel={"unit": ["geo", "brand"], "time": "week"}` builds a composite unit key (`North|Acme`) and pools over its Cartesian product; rectangularity validated with actionable errors; public results keep the flat `(chain, draw, unit)` dims contract.
- **#431 — structured pooling:** `pooling={"by_var": {"tv": {"coefficient": ("geo",)}, "theta_tv": "none"}}` pools individual predictors/transform params over requested dimension subsets (`mu_{var}_{dim}`/`sigma_{var}_{dim}` hyperpriors) or forces per-cell params. Flat pooling dict unchanged. Transform-leaf predictors rejected at parse time (pooling raw columns under transformed terms would silently zero the media signal).
[...]

PR #471, opened 2026-09-10
feat: generative residual-covariance realizations for descendant equations

### LINKED ISSUE #469: simulate() generative residual-covariance realizations for descendant equations
## Summary

Follow-up to PR #468 review finding B1(b): make residual-block realizations first-class in the generative graph so descendants of `~~` block members receive the realized correlated noise, not zero-filled data columns or mean-only structure.

Related: #434, #468

## Design requirements

1. `compile_to_pymc(..., generative: bool = False)` — only `simulate()` / `simulate_params_template()` pass `True`. Estimation must not grow extra residual RVs.
2. Compile residual blocks *before* any non-block consumer (explicit schedule; topo order is not enough — e.g. `Y1 ~~ Y2`, `Z ~ Y1` admits `[X, Y1, Z, Y2]`).
3. Descendants must read **realized noisy values**, not `mu_{var}`. Estimation reads the observed noisy column; generating from `mu` and estimating against realized values is attenuation.
4. Drop the numpy post-hoc Cholesky in `simulate()`.
5. Tests: hand-written NumPy reference for `Z ~ Y1` slope (truth 3.0) plus a recovery HDI check on `beta_Z`.
6. Keep estimation dual-wiring green (`TestBlockMemberDualWiring` in `tests/test_residual_cov.py`, issue #217): `prefer_observed_block_members=True` for likelihood, noisy generative RVs only on the `generative=True` path.
7. `PathModel.do` / `run_do_cross_sectional` currently treat block members as endogenous ∩ not-free-RV ∩ has `mu_{var}` and intervene on `mu_{var}`. A free generative RV named `var` will change that heuristic — update `simulate.py` in this follow-up, not in #468.

## Context

PR #468 ships scoped `NotImplementedError` guards for (a) descendants of block members and (b) scan-panel + `~~` combos. This issue tracks the real generative rewrite.

drbenvincent:
Closes #469

## Summary
- `compile_to_pymc(..., generative=True)` (used only by `simulate()` / `simulate_params_template()`) emits a free joint residual RV so descendants of `~~` block members read realized correlated noise, not `mu_{var}` or zero-filled columns.
- Residual blocks are compiled before any non-block consumer (topo order alone can place `Z` before `Y2`).
- Drop the NumPy post-hoc Cholesky in `simulate()`. Estimation dual-wiring (`prefer_observed_block_members=True`) is unchanged.
- `do()` identifies block members from `graph_info.residual_blocks` so the heuristic does not depend on whether members are free RVs.

## Test plan
- [x] NumPy oracle: OLS of `Z` on realized `Y1` recovers slope 3.0 (and is not the attenuated `mu_Y1` slope)
- [x] Recovery HDI/mean check on `beta_Z`
- [x] `TestBlockMemberDualWiring` still green
- [x] Within-block and scan-panel `~~` still raise
- [x] `make lint`

### DIFF
--- a/docs/user_guide/18-panel-data.qmd
+++ b/docs/user_guide/18-panel-data.qmd
@@ -256,7 +256,7 @@ User-facing inputs and outputs use business units: `do(set=)` values are divided
 
 ## Residual covariance simulation
 
-Cross-sectional residual-covariance blocks (`Y1 ~~ Y2`) can be simulated only when no equation reads a block member. Until correlated residual draws are represented inside the generative graph, `simulate()` rejects descendants—including directed edges between members of the same block—instead of silently generating a different DAG. Scan-compiled panel models cannot yet combine `simulate()` with residual covariances.
+Cross-sectional residual-covariance blocks (`Y1 ~~ Y2`) are realized inside the generative graph, so descendants such as `Z ~ Y1` receive the noisy block draws rather than `mu_Y1`. Directed edges *within* the same `~~` block (`Y2 ~ Y1` together with `Y1 ~~ Y2`) still raise `NotImplementedError`, as do scan-compiled panel models that combine `simulate()` with residual covariances.
 
 ## Latent dynamics with sparse measurements
 
--- a/pathmc/_model.py
+++ b/pathmc/_model.py
@@ -2514,9 +2514,10 @@ def simulate(
         (``~~``), supply ``"chol_{block_name}"`` as the packed
         lower-triangular Cholesky vector of length ``k(k+1)/2``
         (PyMC ``LKJCholeskyCov`` packing); block members' simulated
-        columns share the correlated residuals. ``~~`` simulation
-        supports blocks whose members are terminal in the DAG; a
-        descendant of a block member raises ``NotImplementedError``.
+        columns share the correlated residuals. Descendants of block
+        members receive the realized noisy draws. Within-block directed
+        edges (one member regressing on another) still raise
+        ``NotImplementedError``.
         For ``hsgp()``
         terms, supply ``"ell_{lhs}_{var}"``, ``"eta_{lhs}_{var}"``, and
         ``"beta_hsgp_{lhs}_{var}"`` (length ``m``). Transform
@@ -2578,9 +2579,8 @@ def simulate(
     ------
     ValueError
         If required parameter values are missing from *params*, if a
-        ``chol_{block_name}`` vector has the wrong length for its block,
-        or if the spec contains ``lag()`` terms but ``panel=`` was
-        omitted.
+        parameter array has the wrong shape, or if the spec contains
+        ``lag()`` terms but ``panel=`` was omitted.
 
     Examples
     --------
@@ -2600,7 +2600,7 @@ def simulate(
     """
     spec = parse_spec(spec_string)
     latent_set = set(latent) if latent else set()
-    block_var_set, blocks = _identify_residual_blocks(spec)
+    block_var_set, _ = _identify_residual_blocks(spec)
     graph_info = build_graph(spec, latent=latent_set)
 
     endogenous_lhs = [reg.lhs for reg in spec.regressions]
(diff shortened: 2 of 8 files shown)


drbenvincent (commented):
## Review (round 1)

Closes #469 in spirit: `generative=True` emits a free joint RV, descendants of a *non-block* LHS read realized `Y1` rather than `mu_Y1`, the NumPy post-hoc Cholesky is gone, estimation dual-wiring stays on `prefer_observed_block_members=True`, and `do()` keys block members from `graph_info.residual_blocks`. The NumPy OLS oracle plus the recovery mean/HDI-regime check match the issue’s test plan.

The advertised contract is still wrong for any descendant that is itself a `~~` member of a *different* block, and the compile schedule does not implement that case. That is a correctness/contract miss against #469 items 2–3 (“descendants read realized noisy values”; only *within-block* directed edges stay NI).

---

### `pathmc/_model.py`

**~2616–2631 — NI predicate is “any ~~ member”, not “same block”.**  
`block_deps = (block_var_set & read_vars) - {reg.lhs}` uses the union of *all* residual blocks. A spec like:

```text
Y1 ~ X
Y2 ~ X
W1 ~ Y1
W2 ~ X
Y1 ~~ Y2
W1 ~~ W2
```

raises `NotImplementedError` claiming W1 reads another member of the *same* `~~` block. W1 is a non-within-block descendant of Y1 — exactly the class this issue exists to generate.

**Fix:** Restrict the check to the residual component that contains `reg.lhs` (the `blocks` list already returned by `_identify_residual_blocks`; you currently discard it with `_`). Keep the NI only for a directed edge *inside* that component. Update the error to name that component. Add a simulate test for the two-block spec above (OLS of W1 on Y1 recovers the structural slope, not the attenuated `mu_Y1` slope).

---

### `pathmc/compile.py`

**~703–711 — Residual blocks are not compiled before a *block-member* consumer of another block.**  
`_ensure_residual_block` runs for referenced blocks only when `var not in block_vars`. When the consumer is itself a block member, the loop only records membership and compiles *that* block once all of *its* members are seen. It never compiles an upstream block first.

So even after fixing the `_model.py` guard, the two-block spec can still compile `W1 ~~ W2` while `Y1` is still the zero-filled data column (topo e.g. `Y1, W1, W2, Y2`). That reintroduces the attenuation bug #469 is about.

**Fix:** Before `_compile_residual_block` for `bidx`, recursively `_ensure_residual_block` for every other block referenced by any member’s `MuSpec` (same helper you already have: `_block_indices_referenced`). If a referenced block is the current one, that is the existing within-block NI.

**~481–488 — Stale function contract.**  
The docstring still says every endogenous variable is a free RV. On `generative=True`, block members are `Deterministic` slices of `{block}_joint`. Say that, and that estimation (`generative=False`) is unchanged.

**~865–907 — Duplicated slot walk (judgement).**  
`_block_indices_referenced` and `_references_block_members` copy the same kind cascade (`plain` / `interaction` / `lag` / `transform`). One helper that yields predictor names would keep HSGP/`hsgp` handling from drifting (the new helper lists `hsgp`; `_references_block_members` does not). Not a merge blocker.

---

### `pathmc/residuals.py`

**~87–92 — Stale class docstring.**  
Still “Emits the block as a single observed MvNormal.” It now emits either `{block}_obs` or `{block}_joint` depending on `observed=`.

---

### `pathmc/simulate.py`

`block_vars = {v for block in graph_info.residual_blocks for v in block}` is the right #469.7 replacement for the free-RV heuristic. No issue here.

---

### `tests/test_simulate_data.py`

**~273–303 — Recovery test vs its docstring.**  
Docstring: “puts 3.0 in the beta_Z HDI.” Body never asserts `lo <= 3.0 <= hi`; it checks mean and `lo > 2.0`. Fine given 50-draw HDIs, but the docstring overclaims. Either assert coverage with enough draws or say “HDI sits in the realized-Y1 regime, not the attenuated one.”

Missing: two-block descendant coverage (see `_model.py`).

The non-block `Z ~ Y1` NumPy oracle is the right shape (independent DGP + simulate OLS, not simulate-and-recover of the same draws).

---

## Verdict

**Request changes.** Do not merge until the within-block NI is actually same-component, the compiler compiles upstream `~~` blocks before a downstream block that reads them, and there is a test that would have caught the two-block case. The #469 happy path (`Z ~ Y1` with `Y1 ~~ Y2`) looks correctly implemented.

drbenvincent (commented):
## Review (round 2)

**APPROVE**

Round 1 blockers are addressed. Same-component NI, recursive upstream `~~` compile, the two-block OLS test, and the recovery/residuals/`compile_to_pymc` opening docstrings all land. #469 items 1–7 look implemented.

---

### Round 1 follow-up

- **`_model.py` NI is same-component.** `var_to_block` / `own_block` only flags directed edges inside the residual component that contains `reg.lhs`. Cross-block `W1 ~ Y1` with `Y1 ~~ Y2` and `W1 ~~ W2` is no longer rejected. Error text names the same-block members.
- **`compile.py` recurses before compiling a block.** `_ensure_residual_block` walks `_block_indices_referenced` on every member `MuSpec` and compiles other blocks first. That covers topo like `Y1, W1, W2, Y2`.
- **`compile_to_pymc` / `LKJResidual` contracts** now distinguish observed `{block}_obs` vs free `{block}_joint` + Deterministic slices.
- **Tests.** `test_residual_cov_cross_block_descendant_numpy_reference` exercises the two-block spec (OLS of W1 on Y1 ≈ 3.0, not the attenuated `mu_Y1` slope). Recovery docstring matches the mean / `lo > 2.0` checks.

---

### Remaining (non-blocking)

**`pathmc/compile.py` module docstring + Returns (leftover from round 1 docstring ask).**  
The function *summary* is updated; the module docstring (lines 16–19) and Returns (“all endogenous vars are free RVs”) still describe the pre-`~~` world. Estimation block members are still not free RVs; generative block members are Deterministic slices. One-paragraph fix.

**`docs/user_guide/18-panel-data.qmd` § Residual covariance simulation.**  
Still says cross-sectional `~~` can be simulated only when no equation reads a block member, and that `simulate()` rejects descendants until correlated draws live in the generative graph. That is the #468 guard this PR removes. Scan-panel + `~~` remains NI; descendants of a *different* block now work. Update that paragraph in this PR or a docs follow-up so the site does not contradict `simulate()`’s docstring.

**Duplicated slot walk (same judgement as round 1).**  
`_block_indices_referenced` lists `hsgp`; `_references_block_members` still does not. Not a merge blocker.

**Two-block test is simulate-OLS only.**  
Unlike `test_residual_cov_descendant_numpy_reference`, there is no independent NumPy DGP oracle. Fine for the round 1 ask; a hand-written `w1 = 3*y1 + e` check would match the package test bar more tightly.

---

## Verdict

**Approve.** Merge the generative `~~` rewrite. Please fix the stale `compile_to_pymc` Returns/module docstring and the user-guide paragraph so the advertised contract stays aligned with the code.

PR #473, opened 2026-09-19
fix: return business-unit datagrids on scaled models

### LINKED ISSUE #470: datagrid() on a scaled model returns mixed units
## Summary

`PathModel.datagrid()` builds from the internal-scale fitted frame (`model._data`). User-supplied `**cols` stay in business units while every covariate held at its mean is the *scaled* column mean. Feeding that grid to `predictions(newdata=)` — the composition the docs show — is silently wrong.

Surfaced in adversarial review of #468 (round 6). Not blocking that PR: round 5 made `predictions(newdata=)` accept business-unit grids, so the two halves of the documented workflow now disagree by construction. Follow-up from the #426 / #432 scaling work.

## Repro (from #468 round 6)

```python
m = pathmc.model(
    "sales ~ tv + radio",
    data=df,
    scaling=pathmc.Scaling(
        target={"method": "max"},
        channel={"method": "max"},
    ),
)
m.fit()
m.datagrid(tv=[500.0])
# [{'tv': 500.0, 'radio': 0.622, 'sales': 0.694}]
# business-unit means: tv ~485, radio ~311, sales ~3523

predictions(m.datagrid(tv=[500]))          # 2087  (wrong)
predictions(hand-built business grid)      # 3566
m.do(set={"tv": 500})                      # 3566
```

A single-predictor model looks fine (nothing is held at a scaled mean). The failure appears as soon as a second scaled covariate is held constant.

## Spec

- `PathModel.datagrid()` returns a grid in **business units** for every numeric column that has a fitted scale factor.
- User-supplied `**cols` are already business units; do **not** inverse-transform them (that would turn `tv=[500]` into `500 * factor`).
- Columns filled from `_data` (held at mean / mode) must be inverse-transformed with `fitted_scaling` before the grid is returned.
- `pathmc.datagrid(df, ...)` on a caller-supplied frame is unchanged (that frame is not the internal scaled `_data`).
- Oracle: `m.predictions(outcome, newdata=m.datagrid(tv=[raw]))` matches `m.do(set={"tv": raw})` on a two-predictor scaled model, and matches a hand-built grid whose held columns are the raw (unscaled) means.
- Docs (`PathModel.datagrid`, interpret guide) state that the bound method returns business units on a scaled model.

## Non-goals

- A general `to_internal` / `to_business` seam with a parametrised contract test over every public method. That is the right longer-term follow-up (see #468 round 6) and must not grow this issue. This ticket is the `datagrid()` hole only.
- Changing `predictions(newdata=)` again; that path already treats the grid as business units.

## Test plan

- [ ] Two-predictor scaled model: `datagrid(tv=[raw])` held `radio` equals the raw radio mean, not the scaled mean
- [ ] `predictions(newdata=m.datagrid(tv=[raw]))` equals `do(set={"tv": raw})` (same oracle as #468 N2, with a second channel)
- [ ] User-supplied column is not multiplied by the factor
- [ ] Unscaled models: `datagrid()` behaviour unchanged
- [ ] `make test-fast` and `make lint`

## Links

- PR: https://github.com/pymc-labs/pathmc/pull/468 (round 6 follow-up)
- Parent meta: #426
- Scaling feature: #432

drbenvincent:
Closes #470

## Summary
- return business units from PathModel.datagrid on scaled models
- preserve supplied grid values while inverse-scaling held columns
- document the bound helper and add a two-covariate prediction/do oracle


drbenvincent (commented):
## Review: changes requested

### `pathmc/_model.py:2269` — per-unit scaling can make `datagrid()` raise for numeric unit keys

`_datagrid()` fills every unspecified numeric column with its mean. For a panel that uses a numeric unit identifier, that includes the scaling dimension (for example, `geo=[1, 1, 2, 2]` becomes `geo=1.5`). When a second scaled predictor is held at its mean, `inverse_transform_column()` asks `_per_row()` for the factor keyed by `geo=1.5`, which is not an estimation-time unit and raises `KeyError`. This is a regression for `PathModel.datagrid()` and leaves the per-unit scaling path untested.

Repro: fit `"y ~ x + z"` with `panel={"unit": "geo", "time": "time"}`, numeric `geo` values 1 and 2, and `Scaling(channel={"method": "max", "dims": ("geo",)})`; then `m.datagrid(x=[50.0])` raises while inverse-transforming held `z`.

Build the grid so scale dimensions retain a real fitted unit key (or explicitly reject/require those dimensions instead of silently averaging them), and add a per-unit scaling regression test with a numeric unit id and two scaled predictors. The test should assert that `datagrid(x=[...])` returns business-unit `z` rather than raising.

Focused test run: `uv run pytest tests/test_interpret.py -x -v` (29 passed).

drbenvincent (commented):
## Review: approved

The prior per-unit scaling issue is resolved. `PathModel.datagrid()` now inverse-transforms each held scaled column from the fitted rows before taking its business-unit mean, rather than attempting an inverse transform against the synthetic grid. This avoids looking up a scale factor for the numeric mean of a unit key (for example, `geo=1.5`). The added regression covers a numeric per-unit key with two scaled predictors and verifies the held predictor is returned in business units.

The implementation also preserves user-supplied values, keeps unscaled `datagrid()` behavior unchanged, documents the business-unit contract, and exercises the `predictions(newdata=m.datagrid(...)) == do(...)` oracle.

Validation: `uv run pytest tests/test_interpret.py -x -v` (30 passed); `make lint` (passed).

PR #476, opened 2026-09-19
Centralize scaling through a single unit-conversion seam

### LINKED ISSUE #472: Single to_internal/to_business seam for scaling, with a parametrised contract test
## Summary

Scaling has no single place that defines the unit convention. Each public method carries its own ad-hoc conversion, so the "estimation is internal-scale, user-facing values are business units" contract is enforced per call site rather than by construction. Every round of review on #468 found a fresh surface that had been missed: `do(set=)` (round 4), coefficient rescaling / scan-panel `predict()` / `slopes()` (round 5), `effects_summary()` `:=` params / `predictions(newdata=)` / `observed_data` (round 5 non-blocking), `datagrid()` (round 6, #470). Each was fixed in isolation. The pattern is the bug.

Proposal: one `to_internal` / `to_business` seam plus a parametrised contract test that every public method must pass, so the next new surface fails a test instead of shipping silently wrong numbers.

## Motivation

The failures share a shape and were all silent (plausible-looking numbers, no warning):

- direction confusion: `slopes()` divided endpoints that `_data` had already divided, evaluating at `x / F**2`
- missing factor key: `b*tv:radio` rescaled by 1.0 because `'tv:radio'` isn't a column, off by 4.9e5
- wrong factor granularity: scan-panel `predict()` used the global row-mean divisor instead of the per-geo one, because the dim was named `sales_dim_0` and only `"time"` was excluded
- partial coverage: labeled coefficients rescaled, `:=` defined params not, so `indirect` and `a * b` disagreed in the same table
- double application: `observed_data` unscaled again on a second `predict()` call

Local fixes cannot prevent the next one. Nothing today tells you which surfaces exist, which direction each needs, or which factor granularity applies.

## Spec

Introduce an explicit conversion seam owned by the fitted scaling object:

- `to_internal(obj, *, kind, dims=None)` and `to_business(obj, *, kind, dims=None)` handling `pandas` frames/series, `xarray`, and scalars.
- `kind` names the semantic role, not the container: `outcome`, `regressor`, `coefficient`, `elasticity`, `derived`. `elasticity` and dimensionless terms are explicitly identity.
- Factor resolution is one function: given a term and the dims present, return the divisor array (global, per-`by` group, or 1.0 for a term with no raw-column factor). Saturated and interaction terms resolve to 1.0 there rather than at each call site.
- Idempotence is enforced by tagging converted objects, so a second call is a no-op rather than a second division.
- Every public method delegates. No inline `* factor` or `/ factor` outside the seam; a lint or grep check in CI would make that testable.

## Test plan

A parametrised contract test over the public surface, so a new method must opt in:

- [ ] Registry listing every public method that returns numbers, with its expected units. A method absent from the registry fails the test.
- [ ] For each: scaled fit and equivalent unscaled fit on the same raw data agree to tolerance on business-unit output.
- [ ] Oracle cross-check per method against `do()` where one exists (already the pattern used in #468 rounds 5 and 6).
- [ ] Shapes matrix: bare term, `adstock()`, `logistic_saturation()`, `lag()`, interaction `a:b`, `:=` defined param.
- [ ] Granularity matrix: no scaling, global scaling, `by=` group scaling, scan panel with per-unit factors.
- [ ] Idempotence: calling the conversion twice equals calling it once, for each `kind`.
- [ ] Deletion check: removing any single conversion call site turns some test red. This is the check that would have caught the round-5 `predict()` block having zero coverage.
- [ ] `make test-fast` and `make lint`

## Non-goals

- Changing the user-facing convention itself. Business-unit outputs and business-unit intervention inputs stay as they are; this is about enforcing them in one place.
- Fixing `datagrid()`. That is #470 and should land independently.
- Any change to how estimation itself scales columns.

## Links

- PR: https://github.com/pymc-labs/pathmc/pull/468 (rounds 4 to 6 surfaced the pattern)
- Immediate hole: #470
- Parent meta: #426
- Scaling feature: #432

drbenvincent:
Closes #472

## Summary

- add `ScalingFactors.to_internal()` / `to_business()` as the single semantic unit-conversion boundary
- centralize raw-value, coefficient, interaction, transform, xarray, scan-panel, and scalar factor resolution
- tag converted pandas, Narwhals, xarray, NumPy, and scalar values so repeated conversion is idempotent
- use content-derived conversion provenance that includes the resolved factor shape and values, rather than Python object identity or caller role
- require exact row/scan shapes (or explicit trailing sample dimensions) and reject equal-size but ambiguous reshapes
- route model, simulation, effects, prediction, and interpretation scaling through the seam
- add a public-method unit registry, executable scaled-vs-unscaled contracts, and a recursive guard against factor access escaping `scaling.py`

## Intentional numerical fixes

- homogeneous lag/adstock coefficients use the underlying raw predictor factor; saturating transforms remain dimensionless
- grouped exogenous fills convert row-aligned values before subgroup selection and reduction
- CATE/ATT/ATU subgroup values are converted to internal units before matching fitted data
- placebo refits remain compiled on internal data while their ATE query uses the original model's business-unit boundary

## Validation

- `make test-fast` — 1767 passed, 3 skipped
- focused scaling/interpret/refute/effects/do suite — 191 passed
- scaling contract and public API suite — 64 passed
- exact adversarial reproductions for transposed scan and same-size row arrays now raise actionable errors
- `make lint` — ruff, format, mypy, metadata, and license checks passed


drbenvincent (commented):
## REQUEST_CHANGES — review round 1

The seam is a strong direction, but this round does not yet satisfy #472's behavioral contract. I found four reproducible correctness failures in the implementation plus a test-design gap that lets all four pass.

### `pathmc/scaling.py`

1. **[P1] Role-aware frame conversion falls back to every factor when the requested role has no columns (line 472).**

   `tuple(role_columns or (set(self.factors) & set(obj.columns)))` treats an empty role match as if role metadata were unavailable. With factors for `X` (regressor) and `Y` (outcome), `to_internal(frame[["geo", "X"]], kind="outcome")` divides `X`, even though `X` is explicitly registered only as a regressor. That silently violates the semantic `kind` contract. Fall back to all fitted columns only when `self.roles` itself is empty (legacy/manual factors); when roles are present, an empty match must remain empty.

2. **[P1] The supported MMM transforms still cannot rescale labeled coefficients (line 315).**

   The new centralized coefficient path consults `transform_impl.homogeneous`, but only `logistic_saturation` declares it. `adstock`, `delayed_adstock`, `weibull_adstock`, and `michaelis_menten` all inherit `None`, so a scaled model with `b*adstock(X, ...)` (or any of the other three) raises from `effects_summary()` instead of returning business-unit coefficients. This is especially important after the new PyMC-Marketing-backed transforms landed on `main`. Declare the three adstocks homogeneous and Michaelis-Menten non-homogeneous, then cover every built-in and nested chain in the coefficient contract matrix.

3. **[P1] Narwhals frame conversion is not idempotent (lines 474-487).**

   The pandas branch reads and writes conversion state, but the `nw.DataFrame` branch always divides/multiplies and returns an untagged frame. A second `to_internal()` therefore scales again; for factors `[10, 20]`, `X=[10, 20]` becomes `[1, 1]` and then `[0.1, 0.05]`. This is also the container used by `transform()` and the internal call sites. Either carry conversion state in a wrapper/sidecar that survives Narwhals operations, or make the seam's ownership explicit so repeated internal calls cannot occur; in either case add the required repeated-conversion test for this path.

### `pathmc/simulate.py`

4. **[P1] Converting a grouped exogenous mean with the mean factor returns the wrong business value (line 1319).**

   `_business_exog_value()` first averages the internally scaled values and then multiplies by the mean divisor. For heterogeneous factors, `mean(X / F) * mean(F) != mean(X)`, and subgroup results are worse because the factor is still averaged over the full frame. A two-row example with internal `X=[1, 5]` and factors `[10, 20]` returns `45` instead of the raw mean `55`; selecting the first row returns `15` instead of `10`. Convert the row-aligned values to business units first, apply `subgroup_indices`, and only then average.

### `tests/test_scaling_contract.py`

5. **[P1] The added “contract” registry does not execute the issue's public-surface contract (lines 212-241).**

   The registry test only compares method names, and the delegation test only searches source text for `to_internal`/`to_business`. It never runs the registered methods, never compares scaled and equivalent unscaled fits, and cannot prove the conversion call is live (dead code containing the string passes). The requested shape matrix is also incomplete: only a raw term, interaction, and logistic saturation are checked; there is no adstock, lag, HSGP, defined-param public output, grouped/scan public-method matrix, or deletion-sensitive behavioral assertion. That is why the four failures above coexist with 27/27 passing contract tests. Parameterize executable cases from the registry and assert business-unit outputs against independent oracles/equivalent unscaled fits across the issue's shape and granularity matrices.

The existing targeted suites pass, but the direct reproductions above show this is not ready to merge yet.

drbenvincent:
Round 1 findings addressed in d9562a9:

- role-aware frame conversion now preserves an empty role match instead of falling back to all scaled columns
- Narwhals frames carry conversion ownership, making repeated conversion idempotent
- all built-in transforms declare dimensional homogeneity (adstocks: homogeneous; saturations: non-homogeneous), with nested/lag/HSGP coefficient oracles
- grouped exogenous values are converted row-wise before subgroup selection and averaging
- the contract suite now includes executable business-unit oracles across do, ATE, comparisons, slopes, predictions, effects_summary, effect, and datagrid

Validation: 141 focused tests passed; make test-fast passed (1746 passed, 3 skipped); make lint passed.

tests/test_scaling_contract.py
```diff
@@ -0,0 +1,394 @@
...
+        channel_columns={"X"},
+    )
+
+    assert factors.roles == {
+        "X": frozenset({"regressor"}),
+        "Y": frozenset({"outcome"}),
+    }
+
+
+def test_public_business_surfaces_share_one_executable_linear_oracle(
```
drbenvincent: [P1] This new oracle is valuable, but it is still disconnected from `PUBLIC_METHOD_UNITS` and covers only one global/fixed/linear model. The issue requires the registry itself to drive executable assertions across every numeric surface and the requested term-shape/granularity matrix; source-string delegation checks are not deletion-sensitive. Please parameterize executable cases from the registry (including grouped/scan and the remaining business-facing methods) and use independent business-unit or equivalent-manual-scale oracles.

drbenvincent (commented):
## REQUEST_CHANGES — review round 2

The four concrete correctness defects from round 1 are fixed: role-empty frame conversion no longer falls back, all built-in transforms declare the expected homogeneity, Narwhals direct repeated conversion is idempotent, and grouped exogenous values are converted row-wise before reduction. The new regressions for those fixes pass, as do the existing defined-parameter and scan-panel tests.

One blocking part of #472 and the prior review remains unresolved.

### `tests/test_scaling_contract.py`

1. **[P1] The registry still is not an executable parametrized public-surface contract (line 358).**

   `PUBLIC_METHOD_UNITS` is only checked against method names, and `test_scaling_boundary_delegates_to_the_conversion_seam` still only searches source text. The new `test_public_business_surfaces_share_one_executable_linear_oracle` is useful, but it is one monolithic global/fixed/linear example covering eight surfaces; it is not parameterized from the registry, does not compare scaled and equivalent manually pre-scaled/unscaled fits, and does not cover every numeric business-facing registry entry. In particular, entries such as `predict`, `counterfactual`, `cate`, `att`, `atu`, `sensitivity`, and `refute_placebo` can remain in the registry without an executable unit assertion. The requested public-method granularity matrix is also still absent: grouped and scan behavior are covered only by isolated helper/legacy tests, while HSGP/lag/adstock are tested only at private coefficient-factor resolution rather than through registered public surfaces.

   Please make the registry carry executable cases (or map every numeric entry to a shared executable contract), parameterize the test over those cases, and add independent business-unit or equivalent-manual-scale assertions for the requested term-shape and global/grouped/scan dimensions. Replace the source-string deletion check with behavioral coverage of each conversion boundary so moving a call into dead code cannot satisfy the contract.

### Checks run

- `uv run pytest tests/test_scaling_contract.py tests/test_scaling.py -x -q` — 89 passed
- focused global-oracle, defined-parameter, scan-predict, and scan-do cases — 4 passed
- `git diff --check origin/main...HEAD` — passed

**Final verdict: REQUEST_CHANGES**

drbenvincent:
Round 2 finding addressed in 2f2e725.

The contract registry now carries executable cases, and the test is parameterized directly across every business-facing numeric PathModel surface: effects_summary, standardized, effect, predict, refute_placebo, do, counterfactual, ATE/CATE/ATT/ATU, sensitivity, prob, predictions, comparisons, slopes, and datagrid. Each case compares a scaled fit with an independently parameterized unscaled fit in business units. The former source-string assertions were removed.

Additional behavioral matrices cover:
- global scaling versus no scaling
- grouped per-unit factors through public do()
- scan-panel per-unit factors through public do() and effects_summary()
- bare, adstock, saturation, interaction, lag, HSGP, and := defined-parameter shapes
- Narwhals/pandas/xarray/scalar idempotence

The expanded harness found and fixed two additional boundary omissions: ATT/ATU subgroup matching now converts the business subgroup value row-wise, and placebo clone ATE queries temporarily inherit the original conversion seam while keeping refit data internal.

Validation: focused scaling/ATT/refute suite 140 passed; make test-fast 1759 passed, 3 skipped; make lint passed.

drbenvincent (commented):
## APPROVE — review round 3

Round 3 closes the executable-contract gap from the prior reviews. I found no remaining blocking correctness or architecture issue against #472.

### What I verified

- `PUBLIC_METHOD_CONTRACTS` is exhaustive over `PathModel`'s public methods, and every business-facing numeric entry is connected to an executable case. The parameterized harness runs all 17 numeric surfaces and compares a scaled fit with an independently parameterized unscaled fit in business units.
- The former source-string delegation assertions are gone. The replacement cases exercise live behavior, while the factor-arithmetic guard confirms the old helper/arithmetic routes do not escape `scaling.py`.
- The public shape matrix covers bare, adstock, saturation, interaction, lag, defined-parameter, and HSGP behavior. The granularity matrix covers no scaling, global scaling, grouped factors, and scan-panel factors.
- The grouped intervention path converts row-aligned values before reduction. I additionally checked ATT and ATU with heterogeneous per-geo treatment, covariate, and outcome factors; both exactly matched the manually unscaled model.
- Placebo refits remain on internal-scale data, while `_permute_and_refit` temporarily attaches the original scaling boundary only for the ATE query and restores the clone afterward.
- All prior production findings are fixed: empty role matches stay empty, built-in transform homogeneity is declared, Narwhals conversion is idempotent, and grouped exogenous fills convert before subgrouping/reduction.
- Existing tests were not weakened or edited; the PR adds `tests/test_scaling_contract.py`.

### Checks run

- `uv run pytest tests/test_scaling_contract.py -q` — 52 passed
- `uv run pytest tests/test_scaling.py tests/test_interpret.py tests/test_refute.py tests/test_effects.py tests/test_do.py -q` — 131 passed
- heterogeneous grouped ATT/ATU scaled-vs-unscaled probe — exact match for both estimands
- `git diff --check origin/main...HEAD` — passed
- CI: docs, packaging, lint/prek, and security checks passed; the main test job was still pending when this review was posted

**Final verdict: APPROVE**

daimon-pymclabs:
## Adversarial review

I tried to break the seam rather than confirm it works. The abstraction is the right one and the call-site cleanup is real, but three of the guarantees the PR description claims do not hold, and one of them replaces a loud failure with a silent wrong number. Ranked by how much I would want them fixed before merge.

All repros below were executed against `pathmc/scaling.py` as it stands on this branch (loaded in isolation, so no PyMC needed).

---

### 1. `reshape` on a size coincidence silently produces misaligned factors where the old code raised

`_factor_for` resolves a per-row factor and then, when the shapes do not match, falls back to `reshape` on a size match:

```python
matrix = per_row[scan.sort_idx].reshape(scan.n_units, scan.n_times).T
...
if arr.size == matrix.size:
    return matrix.reshape(arr.shape)
...
if arr.ndim > 0 and arr.size == len(per_row):
    return per_row.reshape(arr.shape)
```

Every call site this replaced used broadcasting (`mat / factor_mat`, `values * per_row`, `arr / per_row`), which raises on a shape mismatch. The seam now reinterprets the bytes instead.

```python
import numpy as np, pandas as pd, narwhals.stable.v1 as nw
from types import SimpleNamespace
from pathmc.scaling import ScalingFactors, _ScaleContext

df = nw.from_native(pd.DataFrame({
    "geo": ["a","a","b","b","c","c"], "x": [1.0]*6}), eager_only=True)
f = ScalingFactors(factors={"x": (("geo",), {("a",):1.0, ("b",):10.0, ("c",):100.0})})
scan = SimpleNamespace(sort_idx=np.arange(6), n_units=3, n_times=2)

# caller passes (n_units, n_times) instead of (n_times, n_units)
out = f.to_internal(np.ones((3,2)), kind="regressor",
                    dims=_ScaleContext(term="x", data=df, scan_info=scan))
print(1/np.asarray(out))
# [[  1.  10.]
#  [100.   1.]
#  [ 10. 100.]]    <- unit c's factor applied to unit a's row, no error
```

The non-scan branch has the same hole whenever a draws array happens to have `size == len(data)`:

```python
df2 = nw.from_native(pd.DataFrame({"geo":["a","a","b","b"], "x":[1.0]*4}), eager_only=True)
f2 = ScalingFactors(factors={"x": (("geo",), {("a",):1.0, ("b",):10.0})})
print(1/np.asarray(f2.to_internal(np.ones((2,2)), kind="regressor",
                                  dims=_ScaleContext(term="x", data=df2))))
# [[ 1.  1.]
#  [10. 10.]]      <- a (chain, draw) array silently treated as 4 panel rows
```

`_layout_unscaled_column` in `_model.py` has the identical `if mat.size == prod(sizes): reshape` pattern, and this PR widens its reach (see item 5).

Suggested fix: require an exact shape match or an explicit dim mapping, and raise otherwise. A size match is not evidence of an alignment.

### 2. The NumPy idempotency tag does not survive arithmetic or slicing

`_ConvertedArray` defines no `__array_finalize__`, so the `_pathmc_scale_units` attribute is dropped by the first operation on the value. Draws are ndarrays that get multiplied, sliced and stacked all over `simulate.py` and `effects.py`, so this is the container where the guarantee matters most and it is the one where it is weakest.

```python
f = ScalingFactors(factors={"x": (("geo",), {("a",):1.0, ("b",):10.0})})
conv = f.to_internal(np.array([8.,8.,8.,8.]), kind="regressor",
                     dims=_ScaleContext(term="x", data=df2))
print(type(conv).__name__)                  # _ConvertedArray  (tagged)
derived = conv * 1.0
print(getattr(derived, "_pathmc_scale_units", None))   # None  <- tag gone
print(np.asarray(f.to_internal(derived, kind="regressor",
                               dims=_ScaleContext(term="x", data=df2))))
# [2. 2. 0.5 0.5]   <- divided twice, no warning
```

`conv[:2]` loses it too. Either implement `__array_finalize__` (and accept that the tag then propagates to unrelated derived quantities, which has its own failure mode) or drop the array tagging and be explicit that arrays are unprotected, rather than shipping a guard that evaporates on first use.

### 3. Idempotency is scoped per `kind`, so the same column converted under two kinds converts twice

`_state_key` is `f"{id(owner)}:{kind}:{term}"`. `transform()` now tags **every** fitted column, outcome included, as `kind="regressor"`:

```python
return self.to_internal(df, kind="regressor",
                        dims=_ScaleContext(columns=tuple(self.factors)))
```

so a subsequent outcome-kind conversion of the same frame is not recognised as already done:

```python
p = pd.DataFrame({"geo":["a","a","b","b"], "y":[8.]*4})
het = ScalingFactors(factors={"y": (("geo",), {("a",):2.0, ("b",):4.0})})
once  = het.to_internal(p,    kind="regressor", dims=_ScaleContext(columns=("y",)))
twice = het.to_internal(once, kind="regressor", dims=_ScaleContext(columns=("y",)))
cross = het.to_internal(once, kind="outcome",   dims=_ScaleContext(columns=("y",)))
print(twice["y"].to_numpy())   # [4.  4.  2.  2. ]  idempotent within a kind
print(cross["y"].to_numpy())   # [2.  2.  0.5 0.5]  double-divided across kinds
```

No caller does frame-level outcome conversion today, so this is latent, not live. But "the outcome column is tagged as a regressor" is not an invariant I would want load-bearing, and it is precisely the class of bug the seam exists to prevent. The tag should key on the column and direction, not on the caller's declared `kind`.

### 4. The tag key is identity-scoped, and `fit_scaling` now hands out a fresh object every call

Two consequences:

- `fit_scaling` on an existing `ScalingFactors` returns `replace(scaling, roles=...)`, a new object with a new `id`. Every tag written by the previous instance is dead weight: the guard silently does nothing across that boundary. The `refute.py` change in this PR (saving `clone._scaling_factors`, swapping in the parent's, restoring in `finally`) is a manual workaround for exactly the double-conversion the tag was supposed to catch. That is worth calling out in the code as the reason, because the next person will try to delete it.
- `id()` is reused after GC. A stale tag left on a long-lived value can match a freshly allocated `ScalingFactors` at the same address and skip a conversion that should happen. I could not reproduce a collision in 500k allocations, so I am flagging this as fragility rather than a confirmed defect, but keying on content (a hash of `factors`) instead of `id` costs nothing and removes the question.

Related: the key omits `outcome`, `data`, `scan_info`, and `scalar`, all of which change the resulting multiplier. The key identifies the caller and the term, not the conversion, so it cannot distinguish "already done" from "a different conversion was requested".

### 5. `_unscale_predict_groups` drops its `var in factors` guard, widening `_layout_unscaled_column`'s reach

```diff
-        if var not in scaling_factors.factors or var not in data.columns:
+        if var not in data.columns:
```

Unscaled observed columns used to be left alone. They now round-trip through `_layout_unscaled_column`, which carries the same size-match `reshape` as item 1. The conversion itself is a no-op (`* 1.0`), so the only thing the removed guard bought was skipping a reshape that can now silently reorder a column. If the goal is uniformity, the layout call should still be skipped when the factor is 1.0, or `_layout_unscaled_column` should be tightened alongside item 1.

### 6. Role semantics diverge between fitted and manual `ScalingFactors`, silently

In `_convert_frame`:

```python
if self.roles:
    columns = tuple(c for c, roles in self.roles.items() if kind in roles and c in obj.columns)
else:
    columns = tuple(set(self.factors) & set(obj.columns))
```

A fitted object filters by `kind`; a legacy or hand-built one ignores `kind` entirely and converts everything. Same call, two behaviors, decided by whether `roles` happens to be populated. Users constructing `ScalingFactors` by hand (which the docstring invites, and which `fit_scaling` explicitly accepts) get the second. I would either require `roles` at construction, or make the fallback raise on a `kind` it cannot honour rather than quietly widening the column set.

### 7. The seam is enforced by a denylist with gaps

`test_factor_arithmetic_does_not_escape_scaling_module` greps four method names over `package.glob("*.py")`. It does not cover `.factors` membership tests, which still happen outside the seam at `_model.py:2283` and `interpret.py:491,494,509`; it does not cover `_scalar_factor` or `_per_row` reached via `getattr`; and `glob` (not `rglob`) means any future subpackage is unchecked.

More structurally: the single boundary is `to_internal(obj: Any, *, kind, dims: Any)` and its context object `_ScaleContext` is private but imported by four modules. `dims` accepting a frame, a mapping, or a private dataclass means neither the type checker nor the reader can tell a correct call from an incorrect one. Making `_ScaleContext` public and typing `dims: _ScaleContext | Mapping | IntoFrame | None` would get most of the safety the denylist is proxying for.

### 8. Behavior changes not mentioned in the PR description

These look like improvements, but they change numbers and none is called out:

- `_coefficient_factor` divides by `lag_of or variable` in the transform branch, where `_term_coefficient_scale` used `term.variable`. Coefficients on lagged **and** transformed terms change value. Intentional? If so it deserves a test naming that shape.
- `_business_exog_value` now converts per row and then reduces; the old path reduced and then multiplied by `mean_factor`. Different under heterogeneous factors. There is a test for it (`test_grouped_exogenous_fill_converts_rows_before_reducing`), just no mention in the summary.
- `cate` now converts `subgroup_value` to internal units before `np.isclose` against `_data`. That is a real bug fix on a scaled model and should get its own line rather than riding along.
- `_scale_scalar_intervention` lost its actionable message. The old error told the user what to do ("Use `do()` on the panel, or pass a value already in scaled units"); the generic `_scalar_factor` message suggests `dims={'data': frame, 'scalar': 'mean'}`, which is not something a user of `counterfactual()` can act on.

### 9. Dead code left behind

- `effects._labeled_coef_business_scale` and `effects._term_coefficient_scale` are now referenced only from `tests/test_scaling.py:829`. The docstring on `ScalingFactors.coefficient_to_business` still points at "the labeled-coefficient path in `pathmc.effects`" that no longer runs.
- `simulate._exog_value` has no remaining callers.

---

### What I checked and could not break

- Narwhals frame tagging works: the attribute sticks on the returned frame and `transform()` is genuinely idempotent within a kind.
- The pandas path does not mutate the caller's frame or its `attrs`.
- `mean_factor`, `coefficient_to_business` and `unscale_xarray` are numerically equivalent to their old bodies on the homogeneous and heterogeneous cases I tried, including the `Intercept` and not-in-`factors` short circuits.
- `build_effects_summary`'s rewrite from a dict comprehension to a spec walk covers the same label set, because `extract_labeled_draws` builds from the same `spec.regressions` terms. Duplicate labels across regressions now resolve last-wins in both the draws and the factor, which is more consistent than before.
- The new xarray coord-aligned branch in `_factor_for` matches `_row_keys`' `astype(str)` key convention on the string and integer coordinate cases.

Items 1 and 2 are the ones I would block on. 3 through 6 I would want addressed or explicitly deferred with a comment saying why.

drbenvincent:
Addressed the adversarial review in `6540272`.

1. **Shape coincidence:** removed both size-only reshapes. Row arrays must be exactly `(n_rows,)` or have `n_rows` as their trailing axis; scan arrays must be exactly `(n_times, n_units)` or have that exact trailing shape. Xarray alignment is likewise validated. `_layout_unscaled_column()` now accepts only exact layouts. The review's `(3, 2)` scan and `(2, 2)` row repros both raise actionable `ValueError`s.

2. **NumPy provenance:** `_ConvertedArray.__array_finalize__()` now copies conversion provenance through ufunc results, views, and slices. Repeating conversion after `converted * 1.0` is a no-op, and sliced views retain the tag.

3. **Cross-kind idempotence:** provenance is no longer keyed by `kind`. It is keyed by the semantic term and the fully resolved factor, so a column transformed under `regressor` is recognized as already internal when subsequently reached as `outcome`.

4. **Identity/context fragility:** removed `id(ScalingFactors)` from the tag. The key now uses a stable BLAKE2 fingerprint of the resolved factor's exact shape and values, which captures row order, scan layout, scalar reduction, and factor content. Equivalent fitted objects share provenance; different resolved alignments do not. `fit_scaling()` also returns an existing object unchanged when no roles need adding. The placebo boundary swap remains intentional: the clone was compiled from already-internal data without scaling metadata, so the original factors are attached only while accepting business-unit interventions and returning business-unit ATE draws.

5. **Observed-data widening:** restored the factor guard through `ScalingFactors.has_factor()`. Unscaled observed variables no longer enter the layout path.

6. **Manual role ambiguity:** implicit frame conversion now raises when a hand-built `ScalingFactors` has factors but no `roles`. Manual callers must either supply roles or explicitly select columns with `ScaleContext(columns=...)`; `transform()` already does the latter.

7. **Seam enforcement/types:** renamed the cross-module context to typed `ScaleContext` and typed `dims` as `ScaleContext | Mapping | IntoFrame | None`. It remains outside the package's explicit exports to preserve the existing public API contract. Callers no longer inspect `.factors`; they use encapsulated `columns_present_in()`, `required_dimensions()`, and `has_factor()`. The enforcement test now uses `rglob()` and also rejects `.factors` access outside `scaling.py`.

8. **Behavior documentation/messages:** the PR description now calls out homogeneous lag/transform coefficient scaling, row-before-reduction grouped fills, subgroup matching in internal units, and the placebo boundary. The scalar counterfactual failure again tells users to use `do()` on the panel or a globally scaled model.

9. **Dead code:** removed unused `simulate._exog_value`. I retained `effects._term_coefficient_scale` / `_labeled_coef_business_scale`: an existing regression test imports them directly, and repository policy says not to rewrite existing test behavior. They are thin delegates into `to_business()` and contain no independent factor arithmetic.

Validation:

- `make test-fast` — 1767 passed, 3 skipped
- focused scaling/interpret/refute/effects/do suite — 191 passed
- contract/public API suite — 64 passed
- `make lint` — passed, including mypy
- `git diff --check` — passed

daimon-pymclabs:
## Re-review of `6540272` — all three findings confirmed fixed

I re-ran the original repros against `pathmc/scaling.py` on this branch (module loaded in isolation, no PyMC), plus a few new ones aimed at the fixes themselves.

**1. Shape coincidence — fixed.** Both size-only reshapes are gone. The row repro (`(2, 2)` array against 4 rows) and the scan repro (`(3, 2)` against a `(2, 3)` time-major layout) now raise actionable `ValueError`s naming the expected shape. The legitimate broadcast case still works: a `(3, 4)` draws-by-rows array converts against 4 rows and gets the right per-row factors.

**2. NumPy provenance — fixed.** `__array_finalize__` carries the tag through ufunc results and views: after `converted * 1.0`, re-converting in the same direction is a no-op, and `converted[1:]` retains the tag.

**3. Cross-kind idempotence — fixed.** With the key no longer including `kind`, a column converted as `regressor` and then reached as `outcome` is recognized as already internal and left alone. The fingerprint change also does what it claims: an independently constructed `ScalingFactors` with identical factor values shares provenance, so the old `id()` fragility is gone.

Also verified from the list: the manual-roles frame path now raises instead of silently converting all scaled columns, the observed-data widening guard is back through `has_factor()` (`_model.py:236`), the dead `simulate._exog_value` is gone, and no module outside `scaling.py` touches `.factors` any more.

### One non-blocking asymmetry

Provenance now survives arithmetic for arrays but not for scalars. `_ConvertedFloat` has no equivalent of `__array_finalize__`, so any operation on a converted scalar returns a plain `float` and drops the tag:

```python
s = factors.to_internal(4.0, kind="outcome", dims=ScaleContext(term="y"))   # 2.0, tagged
factors.to_internal(s, kind="outcome", dims=ScaleContext(term="y"))          # 2.0  — no-op, correct
factors.to_internal(s + 0.0, kind="outcome", dims=ScaleContext(term="y"))    # 1.0  — silently divides twice
```

This is the same silent-wrong-number failure mode as finding 2, just on the scalar branch, and it predates this PR. It only bites if a derived scalar re-enters the seam, which I did not find a live call site for. Fine to merge as is and handle separately, but worth a line in the module docstring saying scalar conversion results are not safe to re-enter after arithmetic.

No remaining blocking issues from my side.

drbenvincent:
Follow-up: this PR does have user-facing behavior. It adds `ScalingFactors.to_internal()` / `to_business()` and fixes the unit contract for causal estimates, placebo refutation, and grouped `do()` fills. I added a user-guide section explaining the conversion API, alignment requirements, and business-unit behavior (commit d7ee88e). I also corrected the PR summary: `cate()` fixes intervention conditions; `att()` / `atu()` select observed treatment subgroups and currently support cross-sectional models only.

Validation: Great Docs build and `make lint` passed locally. The docs page uses display-only code, so no frozen notebook outputs needed updating.

daimon-pymclabs:
Checked `d7ee88e` against the code. The docs are accurate on the points I could test, with one sentence I would qualify.

Verified by executing the documented snippets (`pathmc/scaling.py` loaded in isolation, no PyMC):

- `factors.to_internal(df, kind="regressor")` converts exactly the fitted role columns and leaves `other` untouched; `to_business()` round-trips.
- Manual `ScalingFactors` with factors but no roles raises the documented `ValueError` on frame conversion, and `dims={"columns": ("tv",)}` selects explicitly as described. Plain mappings are accepted by `_coerce_context`, so the `dims={...}` form in the guide works and does not require constructing a `ScaleContext`.
- `dims={"term": "tv"}` works for unnamed individual values.
- The `att()` / `atu()` cross-sectional restriction the paragraph states matches the code: `_model.py` documents and raises `NotImplementedError` for panel models.

### The idempotence sentence is not true for scalars

> Repeating a conversion in the same direction on its result leaves the values unchanged.

This holds for pandas, Narwhals, xarray and NumPy, including after arithmetic (`converted * 1.0` keeps its tag via `__array_finalize__`). It does not hold for scalars, because `_ConvertedFloat` has no equivalent hook:

```python
s = factors.to_internal(4.0, kind="regressor", dims={"term": "tv"})   # 2.0, tagged
factors.to_internal(s,       kind="regressor", dims={"term": "tv"})   # 2.0  — no-op, correct
factors.to_internal(s + 0.0, kind="regressor", dims={"term": "tv"})   # 1.0  — divides twice
```

This is the same asymmetry I flagged as non-blocking on `6540272`, but the guide now promises the property to users in general terms, which raises the cost of leaving it unstated. Two ways out, either is fine: add a half-sentence to that paragraph saying a scalar result loses its conversion tag once it is used in arithmetic, or give `_ConvertedFloat` the same provenance-preserving behavior as `_ConvertedArray` so the sentence becomes unconditionally true.

Nothing else from me on the docs commit.

PR #477, opened 2026-09-19
Add first-class categorical predictor support

### LINKED ISSUE #372: Feature: first-class categorical predictor support
## Summary

pathmc has no first-class categorical predictor support. A user with a `region` column must pre-dummy their dataframe and list the resulting columns manually in the formula, which means they also hand-manage the reference level and get one independent prior per level with no shrinkage.

This is the one item in the basis-terms discussion (#371) that is a plain missing feature rather than a design improvement. It stands alone regardless of whether the basis refactor happens, though it lands much more cheaply on top of it.

Related: #371 (basis-term abstraction), #230 (design-matrix standardization).

## The mechanism, precisely

pathmc **already depends on patsy** (`pyproject.toml`), and `build_design_matrix` already calls it (`pathmc/compile.py:284`). The expansion machinery is present and structurally disconnected from the term system.

`build_design_matrix` passes patsy a formula built as a plain `" + ".join(t.variable for t in reg.terms)`, while `get_predictor_columns` (`pathmc/compile.py:102`) returns one name per term. For a string column patsy emits:

```
patsy.dmatrix("region", df)     -> ['Intercept', 'region[T.b]', 'region[T.c]']
patsy.dmatrix("0 + region", df) -> ['region[a]', 'region[b]', 'region[c]']
```

The term system asks for `region`; patsy produces `region[T.b]` and `region[T.c]`; nothing reconciles the two. So the dependency is already paid for and its main feature is unreachable.

Verification note: the patsy output above was executed. The downstream name-mismatch failure is **read from code, not executed** — no `pymc`/`uv` was available in the sandbox where this was traced. Reproducing the actual failure mode (and its error message, if any) should be the first step.

## Why it is a basis term

A factor is exactly the basis shape: one input column expands to K−1 (or K) design columns multiplying a coefficient vector the term owns. That is the same plumbing #362 built for HSGP and #371 proposes to generalize — the slot skip in `get_free_predictor_columns`, the `free_idx` short-circuit in `build_mu`, the per-term coefficient coord, the resolver returning an assembled contribution.

In the two-axis framing from #371:

- **Build contract:** pure **data basis**. Columns are numbers computed from observed data before the model exists. No parameter dependence, no symbolic path needed for the common case.
- **Coefficient prior:** this is where categoricals are interesting. The default should probably not be K−1 independent priors. Hierarchical shrinkage across levels is the standard answer and is the whole reason to have first-class support rather than pre-dummied columns. That makes categoricals a second consumer of the coefficient-prior axis in #371, alongside HSGP's `beta * sqrt_psd`.

## The fit-time state problem

As a data basis, a categorical term has state that must be frozen at fit and replayed at `predict()`/`do()`: **the category set and the reference level**. If new data contains an unseen level, or the levels arrive in a different order, silently re-deriving the contrast coding puts the counterfactual on a different design matrix with no error. This needs an explicit test, not just a note.

## `do()` semantics need a decision

Intervening on a categorical is a different problem from intervening on a numeric predictor. `do(set={"region": "b"})` means *selecting a level*, not setting a number. Open questions:

- Does `do()` accept a level label, a one-hot row, or both?
- What happens for an unseen level — error, or is that a legitimate counterfactual?
- How does this interact with the reference level under treatment coding?
- What does an `EstimandResult` over a categorical treatment look like — pairwise contrasts against the reference, or all pairwise?

This is a semantics question, not a materialization one, and it is the part that cannot be inherited from the basis machinery.

## Scope

In scope:

- `y ~ region` (and/or explicit `C(region)`) for a string or categorical dtype column, cross-sectional
- Reference-level control, with the chosen level recorded and surfaced in introspection
- Configurable coefficient prior: independent by default or hierarchical, with a documented recommendation
- Fit-time state (level set, reference) frozen and replayed under `predict()` and `do()`
- `do()` on a categorical predictor by level label, with the semantics above settled and documented
- Introspection: equation and LaTeX rendering that names levels rather than emitting an opaque vector
- Docs: a user-guide section and a worked example

Explicitly out of scope for a first pass:

- Panel/scan models
- Categorical **outcomes** (that is a likelihood change, a different layer entirely)
- Interactions between two categoricals, and categorical-by-numeric interactions
- Ordered factors and non-treatment contrast schemes (sum, Helmert, polynomial)
- The formulaic migration — deliberately deferred, see the note below

## Note on formulaic

The obvious question is whether to swap patsy for formulaic first, since formulaic does contrast coding with explicit reference levels and records stateful fit-time transforms. That is worth doing later and has been deliberately deferred; it should not block this issue. Two reasons it buys less than it appears to: it says nothing about the coefficient prior structure (the shrinkage question above), and nothing about `do()` semantics on a factor — which together are the bulk of the work here.

## Acceptance criteria

- [ ] The current failure mode is reproduced and recorded (what actually happens today with `y ~ region` on a string column)
- [ ] `y ~ region` compiles, samples, and recovers known level effects in a simulation test
- [ ] Reference level is explicit, controllable, and visible in introspection
- [ ] Coefficient prior structure is declared rather than hardcoded; hierarchical shrinkage is available
- [ ] Level set and reference level are frozen at fit and replayed identically under `predict()` and `do()`, with a test that fails on recomputation from new data
- [ ] Unseen levels at `predict()`/`do()` time produce a clear error naming the variable and the offending level
- [ ] `do()` on a categorical accepts a level label, with contrast semantics documented
- [ ] `beta` index alignment tested with a categorical mixed with plain predictors and a fixed-value term
- [ ] Equation and LaTeX introspection name the levels
- [ ] User-guide coverage plus a worked example, with `_freeze/` regenerated

drbenvincent:
Closes #372

## Summary
- infer string/category predictors and support explicit `C(..., reference=..., prior=...)` terms
- freeze treatment-coding state across prediction and label-based `do()` interventions, with clear unseen-level errors
- add independent and hierarchical categorical coefficient priors, introspection, simulation, tests, and docs

## Reproduced baseline failure
Before this change, `pathmc.model("y ~ region", data=df)` failed with `ValueError: could not convert string to float: 'a'`.\n\n## Validation\n- `uv run pytest tests/test_categorical.py -x -v`\n- `make test-fast` (1703 passed, 2 skipped)\n- `make lint`\n- `uv run great-docs build`\n- `make freeze-page PAGE=docs/examples/01-foundations/categorical_predictors.qmd`


drbenvincent (commented):
## Review round 1 — REQUEST_CHANGES

The implementation covers the main surface well, and the targeted categorical suite passes locally (`9 passed`). I found two blockers against the issue's definition of done / repository error contract, plus one documentation mismatch.

### `tests/test_categorical.py`

**Line 117 — the frozen-state regression test cannot detect recomputation.**

The issue explicitly requires “a test that fails on recomputation from new data.” This test changes row order, but its only assertion is the posterior-predictive row count. Rebuilding a fresh contrast matrix from `new_data` would produce the same shape, so the test would still pass while labels were mapped to the wrong coefficients. Please assert the actual fitted coding (for example, inside `_prediction_data`, verify that `_cat_y_region` maps `west`, `north`, `south` to the columns frozen from the fit), or assert label-specific predictions against a deterministic oracle. The same frozen state should be covered through `do()` as required by the acceptance criterion.

### `pathmc/categorical.py`

**Line 74 — an explicit categorical term with a missing source column escapes validation.**

`fit_categorical_terms()` silently continues when `term.variable` is absent even when the user wrote `C(region)`. Compilation then reaches `np.column_stack([])` and fails with `ValueError: need at least one array to concatenate`. I reproduced this with `pathmc.model("y ~ C(region)", data=pd.DataFrame({"y": [1.0, 2.0]}))`. This violates the repository rule that errors name the problem and suggest a fix. For an explicit categorical term, fail here with an error that names `region` and tells the user to add the predictor column (categorical endogenous/latent inputs are out of scope for this issue).

### `docs/user_guide/11-model-specification.qmd`

**Line 106 — the documented default reference rule is incomplete for pandas categorical data.**

The implementation sorts string values, but deliberately preserves `CategoricalDtype.categories` order. Therefore “the first sorted level” is not generally the reference for both input types named above. Please say that strings use the first sorted level while pandas categoricals use the first declared category (or otherwise document the exact implemented rule).

Once these are addressed, the implementation otherwise appears aligned with the requested compilation, priors, intervention, introspection, and docs surfaces.

drbenvincent (commented):
## Review round 2 — REQUEST_CHANGES

The three round-1 findings are addressed, and the focused categorical suite passes locally (`11 passed`). I found one remaining fit-state correctness issue and one documentation error.

### `pathmc/categorical.py`

**Line 44 — unused pandas categories are treated as fitted/observed levels.**

`_fit_levels()` copies `series.cat.categories` verbatim. A pandas categorical dtype can declare categories that do not occur in any fitting row, so this lets an unobserved label become the reference (if it is first) or receive a completely prior-driven coefficient. It also makes that label legal in `predict()` and `do()`, even though the issue requires unseen fit-time levels to be rejected and the user guide says unseen counterfactual levels are unsupported. For example, fitting values `['south', 'west']` with declared categories `['north', 'south', 'west']` currently chooses the unobserved `north` reference and accepts `do(set={'region': 'north'})`. Preserve the dtype's declared order, but filter it to categories actually present in the fitting rows; then validate an explicit reference against that observed set. Add a regression test covering an unused declared category at fit and its rejection during prediction/intervention.

### `docs/user_guide/11-model-specification.qmd`

**Line 142 — the intervention example calls an undefined variable.**

The section constructs the model as `model`, but the `do()` example later calls `fitted.do(...)`; there is no `fitted` binding in the page. Use `model.do(...)` (and, if execution is intended, make clear that `model.fit()` must run first) so the documented example is internally consistent.

Once the fitted level set is restricted to values actually seen during fitting and the example variable is corrected, the rest of the implementation appears aligned with the issue's compilation, priors, frozen-state replay, intervention, introspection, and documentation requirements.

drbenvincent (commented):
## Review round 3 — REQUEST_CHANGES

The round-2 findings are addressed, and the focused categorical suite passes locally (`12 passed`). I found one remaining implementation incompatibility and one documentation mismatch.

### `pathmc/_model.py`

**Lines 2537–2544 and 2786–2795 — explicit numeric categorical predictors are scaled after their levels are frozen.**

`fit_categorical_terms()` records the raw numeric labels before scaling, but both `model()` and `simulate()` include every numeric term in `channel_columns`. With `C(region)` where `region` contains labels `1` and `2`, `Scaling(channel={"method": "max"})` changes those values to `0.5` and `1.0`; categorical encoding then raises `Categorical predictor 'region' contains unseen level(s) [0.5]. Fitted levels are [1, 2].` I reproduced this in both entry points. A categorical label is not a numeric magnitude and must never be scaled. Exclude fitted categorical source variables from channel scaling in `model()` and `simulate()` (and either reject or ignore matching entries in a supplied `ScalingFactors`), then add a regression test using an explicit numeric category alongside a genuinely numeric scaled predictor.

### `docs/user_guide/11-model-specification.qmd`

**Line 106 — the pandas categorical reference rule still overstates declared categories.**

After the round-2 fix, unused declared categories are removed before the reference is chosen. Therefore the reference is not always “the first declared category”; it is the first **observed** category in declared order. Please update this sentence so the documentation matches the now-correct fitted-state behavior.

Once categorical sources are excluded from scaling and the reference-order sentence is corrected, the implementation appears aligned with the issue's remaining acceptance criteria. The focused categorical tests pass, and the docs/lint-related CI checks visible during this review are green (the main test job is still pending).

drbenvincent (commented):
## Review round 4 — REQUEST_CHANGES

The two round-3 findings are addressed: numeric categorical labels are now excluded from both fitted and supplied scaling factors, and the pandas-category reference rule is documented accurately. The focused categorical suite passes locally (`14 passed`). I found two remaining formula-validation issues in the inferred-categorical path.

### `pathmc/categorical.py`

**Line 101 — inference silently discards fixed-value and labeled coefficient prefixes.**

The parser correctly rejects `1*C(region)` and `label*C(region)`, but their automatically inferred equivalents are accepted. For a string column, `y ~ 1*region` and `y ~ label*region` reach this line with `fixed_value` or `label` populated; replacing the term with a categorical call leaves those fields present, then `build_mu_specs()` short-circuits on the categorical slot and ignores them. Both specifications therefore compile a free `beta_y_region`, silently changing the user's requested coefficient semantics. Apply the same validation used for explicit `C(...)` when a plain term is inferred as categorical: reject any `fixed_value` or `label` prefix with an actionable error, and add regression tests for both forms.

**Lines 78–88 — categorical interactions bypass the intended unsupported-feature guard and fail with `KeyError`.**

For `y ~ region:x`, the parsed term's `variable` is `region:x`, so the early column-presence check continues before inspecting `interaction_of=('region', 'x')`. `build_design_matrix()` later receives patsy's expanded interaction columns but asks for the synthetic `region:x` column and raises `KeyError: 'region:x'`. The docs explicitly say categorical interactions are unsupported, and repository errors must name the problem and suggest a fix. Inspect interaction component columns before the `term.variable` lookup (or reject any interaction whose component is string/categorical) so this raises the existing actionable `NotImplementedError`; add a regression test.

With those validation holes closed, the implementation otherwise meets issue #372's compilation, frozen-state replay, prior, intervention, introspection, scaling, and documentation requirements.

drbenvincent (commented):
## Review round 5 — REQUEST_CHANGES

The round-4 findings are addressed, and the focused categorical suite passes locally (`17 passed`). I found two silent data-modeling errors and one remaining validation gap.

### `pathmc/categorical.py`

**Line 34 — every pandas `object` dtype is inferred as categorical, including numeric object columns.**

`pd.api.types.is_string_dtype(dtype)` returns `True` for the generic `object` dtype without inspecting its values. Consequently an existing numeric predictor stored as `object` is silently changed from a continuous slope into treatment-coded levels by this PR. I reproduced `pd.Series([1, 2, 3, 4], dtype=object)` under `y ~ x`: it produces `x[T.2]`, `x[T.3]`, and `x[T.4]` plus `beta_y_x`, instead of the prior continuous `x` column in `beta_y`. Restrict automatic inference to actual string-valued series (inspect the series, not just its dtype) or pandas categorical dtype; users can still opt other object-backed labels in with `C(x)`. Add a regression test showing an object-backed numeric predictor remains continuous while an object-backed string predictor is inferred.

**Line 62 — distinct fitted levels with the same string rendering collapse to one design column.**

Column names and PyMC coords are generated with `str(level)`. With explicit `C(region)` and observed levels `1` and `'1'`, the no-intercept columns are both named `region[1]`; the intermediate dict overwrites one, and compilation then stacks the surviving column twice. The numeric level is encoded as all zero while the string level activates both coefficients, silently producing the wrong model (and duplicate coordinate labels). Generate collision-free internal column/coordinate labels that preserve level identity, or reject such fitted levels with a clear actionable error before building the design. Add a regression test for labels that have equal string representations.

### `pathmc/parse.py`

**Lines 408–415 — `C(region):x` still bypasses the categorical-interaction guard.**

The round-4 fix handles inferred `region:x`, but the explicit spelling `C(region):x` enters `_parse_categorical_expr()` and raises `ParseError: Unclosed parenthesis ... Add a closing ')'`, even though the parenthesis is closed. This is misleading and does not identify the documented unsupported feature. Detect a top-level `:` containing `C(...)` before categorical-call parsing and raise the same actionable categorical-interaction `NotImplementedError`/`ParseError`; cover both operand orders (`C(region):x` and `x:C(region)`) in the regression test.

Once these are addressed, the implementation otherwise appears aligned with issue #372's acceptance criteria.

drbenvincent (commented):
## Review round 6 — REQUEST_CHANGES

The three round-5 findings are addressed, and the focused categorical suite passes locally (`21 passed`). I found two remaining silent correctness problems in the replacement-data path introduced by this PR.

### `pathmc/compile.py`

**Line 761 — one explicit categorical term globally removes the source variable's mutable data node, so mixed categorical/continuous uses are updated incorrectly.**

`categorical_source_vars` is global to the model: if any term uses `C(region)`, `region` is omitted from `data_vars` everywhere. But categorical status is stored per term, and the parser currently accepts specifications such as `y ~ C(region) + region` (or `m ~ C(region)` plus `y ~ region`) for numeric labels. The fitted design contains both the treatment-coded and continuous columns, yet the compiled model has no mutable `region` node. `predict(data=...)` therefore updates `_cat_y_region` while the plain `region` contribution remains frozen at the fitting rows; `do(set={"region": ...})` likewise updates only the categorical basis, and global scaling exclusion also silently changes the plain term's behavior. I reproduced this with labels `[1, 2, 1, 2]`: replacement data of all `2`s changed `_cat_y_region` to all ones while `"region" not in model.pymc_model.named_vars`.

Either reject mixed categorical/non-categorical use of a source variable during categorical-state resolution with an actionable error, or retain/update the raw `pm.Data` node and apply the intended scaling/intervention semantics for the non-categorical terms. Add a regression test that exercises both `predict(data=...)` and `do()` (including the cross-equation form if mixed use is supported).

### `pathmc/_model.py`

**Lines 953–955 — `predict(data=...)` leaves fitted `observed_data` in internal scaled units when replacement data omits the outcome.**

The method now passes `prediction_data` to `_unscale_predict_groups()` for every group. Replacement data is documented as predictor data and ordinarily does not contain the outcome, so the helper skips rebuilding `observed_data` for a scaled target. On a model with `sales=[100, 200, 110, 210]`, target max scaling, and a categorical `region`, the first `predict(data=pd.DataFrame({"region": ...}))` returned posterior predictions in business units but left `observed_data["sales"]` as `[0.476..., 0.952..., 0.523..., 1.0]`. This violates the existing user-facing scaling contract and makes one result mix units.

Use replacement data for posterior-predictive row factors while rebuilding `observed_data` from the fitted `self._data` (for example, split the helper's data arguments), and add a regression test for target scaling plus predictor-only replacement data.

Apart from these two boundary interactions, the implementation now appears to satisfy issue #372's categorical inference, frozen coding, priors, intervention, introspection, simulation, and documentation requirements.

drbenvincent (commented):
## Review round 7 — APPROVE

The two round-6 blockers are resolved:

- categorical-state resolution now rejects a source variable used both categorically and continuously, including cross-equation use, before compilation can freeze one representation while updating the other;
- replacement-data prediction now uses replacement rows to unscale posterior predictions while rebuilding `observed_data` from the fitted frame, preserving the business-unit contract for predictor-only replacement data.

I also rechecked issue #372's acceptance criteria and the categorical resolution boundary. String and pandas categorical predictors compile with deterministic, fitted treatment coding; explicit references and independent/hierarchical priors are represented; prediction and `do()` replay the frozen state and reject unseen levels; numeric object columns remain continuous; ambiguous public level labels are rejected; categorical interactions and mixed representations fail with actionable errors; simulation, introspection, and documentation cover the new behavior.

Local verification:

- `uv run pytest tests/test_categorical.py tests/test_scaling.py -q`: 74 passed
- `git diff --check origin/main...HEAD`: passed

No remaining correctness or contract findings.

PR #478, opened 2026-09-19
feat: add reusable basis terms and fourier

### LINKED ISSUE #371: Generalize the HSGP basis machinery into a reusable basis-term abstraction (with fourier() as the second member)
## Summary

PR #362 introduces the first pathmc term that contributes **an M-column basis** to `mu` rather than a single `beta` column. That plumbing is the reusable part. This issue proposes generalizing it into a **basis-term abstraction**, with `fourier(x, n=..., period=...)` as the motivating second member, and covering the docs work needed to make basis terms discoverable as a family rather than as two unrelated one-offs.

Not a blocker for #362. Building a registry with one member is speculative generality; the second member is what reveals the right seams. This issue is to make the refactor a known cost rather than a surprise.

Related: #360 (HSGP umbrella), #362 (HSGP implementation), #98 (time-varying coefficients), #372 (categorical predictors), #373 (spline basis terms).

## Why the HSGP work generalizes

pathmc's design-matrix builder historically emitted one column per term, and `build_mu` walked slots incrementing `free_idx` into a flat `beta` vector. HSGP broke that assumption: one input expands into M basis columns multiplying M coefficients that the term owns itself. Making that work required teaching every downstream consumer that a term can own its own coefficients:

- `get_free_predictor_columns` skips the slot (it contributes no free `beta` column)
- `build_mu` short-circuits before the `free_idx` bookkeeping
- coords register a per-term weights dimension
- the resolver returns the assembled contribution vector instead of a data column
- `_compile_residual_block`'s `has_free` check excludes it

That is the hard, easy-to-get-silently-wrong part (a `free_idx` misalignment attaches the wrong coefficient to the wrong predictor without crashing), and it is now paid for.

## Why it is not reusable yet

The current implementation is one hardcoded instance, not a family:

| Concern | Today (HSGP-specific) | Proposed |
| --- | --- | --- |
| AST node | `Term.hsgp: HSGPCall \| None` | `Term.basis: BasisCall \| None` |
| Slot | `kind="hsgp"`, `coeff_type="hsgp"` string literals | `kind="basis"`, plus `call.name` |
| Coord | `f"{lhs}_{var}_hsgp"` | `f"{lhs}_{var}_{call.name}"` |
| Prior keys | `ell_`, `eta_`, `beta_hsgp_` | basis-declared param names |
| Introspection | hardcoded `f_hsgp(x)` | basis-declared render string |
| Guards | `_reject_endogenous_hsgp_inputs` et al. | basis-declared capability flags |

Adding `fourier()` by copying the pattern means `Term.fourier`, `kind="fourier"`, and a second branch in roughly seven places. Two members is where you refactor.

### Proposed shape

Mirror what `pathmc/transforms.py` already does for `adstock`/`logistic_saturation`: a registry keyed on DSL function name, with a `register_basis()` public extension point matching `register_transform()`.

A registry entry declares **two independent things**: how its columns get built, and what the prior structure over its own coefficient vector is. Those are separate axes and the design should keep them separate — see the two sections below. Beyond that, each basis declares a render string for `introspect.py` and capability metadata (`supports_panel`, `supports_multidim`, `terminal`, etc.), so the guards become declarative rather than nine bespoke functions.

The compiler then stops caring which basis it is.

### Axis 1: the build contract — data bases vs. graph bases

The single question that splits the registry is: **can the columns be computed as numbers before the model exists?**

A **data basis** is a pure function of observed data. Hand it a column of numbers, get back a matrix of numbers. Categorical dummies, B-splines, orthogonal polynomials, Fourier over an observed time index. It is materialized once, before sampling, and baked into the graph as constant data.

A **graph basis** cannot be materialized that way, for either of two distinct reasons:

- **The basis depends on parameters.** HSGP is this: `sqrt_psd` is a function of the estimated `ell`/`eta`, so the basis is re-evaluated at every leapfrog step. It is a PyTensor expression, not an array.
- **The input is latent.** `m ~ x; y ~ hsgp(m)` has no numeric column for `m`. Even a mathematically deterministic basis (Fourier, splines) becomes a graph basis here, because sin/cos must be applied to a symbolic node. Same inside `pytensor.scan`, where the input is a per-timestep slice that does not exist until the loop runs.

So this is **not a property of the basis function — it is a property of the pairing of function and input**. Fourier over an observed week index is a data basis; Fourier over a latent mediator is a graph basis. HSGP is stuck as a graph basis regardless, because of the parameter dependence. That asymmetry is why this needs two build signatures rather than one boolean flag:

```python
# data basis
build(x: np.ndarray, call) -> (columns: np.ndarray, state: BasisState)
# graph basis
build(x: TensorVariable, call, *, params, model_ctx) -> (columns: TensorVariable, n_basis: int)
```

Everything downstream is identical for both — the coord, the `beta` skip in `build_mu`, the introspection string, `do()` support. The split lives at one function boundary, which is the good news.

The conceptual payoff is a clean answer to **when state gets frozen**. A data basis has fit-time state (category set, knot positions, centering constants) that must be recorded at fit and replayed identically at `predict()` and `do()`, or the counterfactual is silently on a different design. A graph basis mostly avoids that by recomputing, which is exactly why `do()` on `hsgp(x)` works at all. HSGP's frozen boundary `L` is the one leak: data-basis-style state stuck inside a graph basis, and precisely the thing that misbehaves when you intervene outside the fitted range.

It also collapses the guard list on #362. "No endogenous inputs" and "no panel/scan" stop being two rules and become one: HSGP is a graph basis and the graph path is not finished. ("No residual blocks" and "no nesting in transforms" are separate concerns about `beta` bookkeeping and term composition, unrelated to this axis.)

**This distinction needs to be written down in the repo, not just in the issue tracker.** The module docstring of the new basis registry is the place for it — it is the file every future basis author reads first, and the two build signatures live there.

### Axis 2: the coefficient-prior structure

A basis must declare not just its columns but **the prior structure over its own coefficient vector**. If the registry only offers "give me columns, I'll give you iid Normals", anything with structured coefficients becomes a bolted-on special case.

This is not hypothetical — it already exists in both directions:

- HSGP's `beta * sqrt_psd` is a scale-structured prior, and the centered/non-centered asymmetry is a consequence of that structure.
- Fourier deliberately wants plain iid Normal over `2n` columns.
- P-splines (#373) want a random-walk prior over adjacent coefficients.
- Categorical terms (#372) want hierarchical shrinkage across levels rather than K−1 independent priors.

Note this axis is **orthogonal to Axis 1**: P-splines have data-basis columns and a graph-side coefficient prior. Formulaic, patsy, or any materializer hands you columns and says nothing about this — so no dependency swap makes it go away.

Designing for `(columns, coefficient_prior_structure)` from the start costs almost nothing now and is expensive to retrofit.

## Why Fourier is the right second member

The obvious framing is that Fourier is an easier citizen than HSGP — no hyperparameters, no kernel, no data-dependent boundary — and that is true. But the sharper reason is this:

**Fourier is the only candidate that is naturally a data basis yet can also be implemented as a graph basis.** That makes it the cheapest way to build and test the *graph* path end to end without HSGP's kernel machinery in the way. Getting `fourier()` working under `pytensor.scan` and over a latent mediator is what unblocks the panel follow-up for every other basis, HSGP included. It exercises both halves of Axis 1 with a function whose correctness is trivially checkable.

The secondary benefits still hold, and they are what make it low-risk to use as the forcing function:

- **No `.eval()` at build time** → no endogenous-input rejection. `m ~ x; y ~ fourier(m, ...)` is fine.
- **No frozen boundary `L`** → no `do()`-outside-fitted-support problem. Sin and cos are defined everywhere, so counterfactual grids beyond the observed range stay valid. Given `do()` is pathmc's whole point, this is a meaningful difference.
- **No `ell`/`eta`** → priors collapse to a single `beta_fourier ~ Normal(0, sigma)` over `2n` columns, and the centered/non-centered asymmetry disappears entirely.

Exercising the abstraction against a member with *fewer* requirements is how you find out which parts of the HSGP path were incidental.

Implementation note: look at `pymc_marketing.mmm.fourier` for the seasonality implementation rather than re-deriving the sin/cos math. On naming, pathmc's DSL is lowercase throughout (`hsgp`, `adstock`, `lag`), so `fourier(x, n=5, period=52)` fits better than `Fourier`.

## Docs (required, not optional)

Right now HSGP is discoverable only via the frozen example `docs/examples/01-foundations/hsgp_1d.qmd`. There is no user-guide page for it, so a user who does not already know the term exists will not find it. Shipping Fourier the same way compounds the problem. **This issue is not done without docs.**

The minimum acceptable option is to extend `docs/user_guide/12-transforms-families.qmd` with a basis-terms section. The better option, and the one I'd recommend, is a dedicated page:

- **New user-guide page** — `docs/user_guide/12b-basis-terms.qmd` (or renumber the 1x block) covering the family as a whole: what a basis term is, how it differs from a transform (one-column-out vs. M-columns-out), the shared `beta`-vector story, when to reach for `hsgp` vs. `fourier`, and each basis's capability limits in one table so the guards are documented before a user hits them.
- Optionally promote this to its own docs **section** alongside the existing Examples grouping if a third basis lands; a single page is right for two.
- **Worked example** — a `fourier()` seasonality example. Either a new `docs/examples/` entry (a `06-basis-functions/` group, or under `04-panel-time/` since seasonality is inherently temporal), or folded into an existing time-series example.
- Add a `Basis Functions` tag with an icon in `great-docs.yml` under `tags.icons`, consistent with the existing `Transforms: waves` entry.
- Cross-link from the HSGP example to the new page and vice versa.
- Remember `_freeze/` is committed and docs CI does not re-execute under `freeze: true`, so any new example needs its frozen output regenerated with `great-docs` and the figures eyeballed before merge.

The data-basis / graph-basis distinction is deliberately **not** a user-guide topic — it is an internal contract. It belongs in the registry module docstring (see Axis 1). The user-facing capability table is the surface form of it.

## Synergies to flag

These are the reasons this is worth more than "one more term":

1. **Seasonality in MMM.** `fourier()` slots directly into `docs/examples/05-applied-models/mmm.qmd` alongside `adstock()` and `logistic_saturation()`. Yearly/weekly seasonality is currently something a pathmc MMM user has to hand-roll as columns; `y ~ fourier(week, n=3, period=52) + logistic_saturation(adstock(spend))` is the natural spec and composes with machinery that already exists.
2. **Panel/time.** Because Fourier is a pure elementwise function, it is the cheapest way to get *any* basis term working under `pytensor.scan`. Whatever the graph path needs will be discovered here at low cost and inherited by HSGP later, unblocking part of #360's panel follow-up.
3. **Time-varying coefficients (#98).** A time-varying coefficient is a basis expansion of time multiplied by a predictor. Once the registry exists, `#98` becomes "add an interaction between a basis term and a predictor" rather than a separate subsystem.
4. **Custom user bases.** `register_basis()` parallels `register_transform()` and gives users splines, wavelets, or domain-specific bases without touching the compiler. `docs/examples/01-foundations/custom_transforms.qmd` already establishes the pattern to mirror.
5. **Introspection and `do()` for free.** Anything that lands in the registry automatically gets equation/LaTeX rendering and intervention support, because those consumers key on `kind="basis"` rather than on a specific term.
6. **Prior scaling.** Both HSGP and Fourier coefficient priors want to know the scale of the response; the coefficient-prior axis is the natural place to address that once, and it touches the unstandardized-design-matrix question in #230 that already shows up as divergences in the HSGP example.

## Suggested sequencing

1. Merge #362 as-is (no refactor).
2. Land the basis registry refactor, moving HSGP onto it with no behaviour change — existing HSGP tests are the safety net for `beta` alignment. Both axes (build contract, coefficient prior) are in place from the start even though HSGP only exercises the graph half.
3. Add `fourier()` as the second member, as a data basis first, then under the graph path (latent input, then `scan`).
4. Docs: user-guide page + worked example + tags + cross-links, as above.
5. Revisit the capability flags for panel support once Fourier is under `scan`.

## Acceptance criteria

- [ ] `register_basis()` public extension point exists and is exported, mirroring `register_transform()`
- [ ] The registry supports both build contracts: data bases returning arrays plus frozen fit-time state, and graph bases returning PyTensor expressions
- [ ] A basis declares the prior structure over its own coefficient vector independently of its columns; HSGP's `beta * sqrt_psd` is expressed through that mechanism rather than as a special case
- [ ] The basis registry module docstring documents the data-basis / graph-basis distinction, including that it is a property of the function/input pairing rather than of the function alone, and what fit-time state each contract must freeze
- [ ] HSGP is migrated onto the registry with no behaviour change and no HSGP-specific branches left in `compile.py`
- [ ] `fourier(x, n=, period=)` works cross-sectionally, with recovery and `do()` tests
- [ ] `fourier()` works over a latent mediator input, exercising the graph build contract without HSGP's kernel machinery
- [ ] Guards are declared as basis capability metadata, and each rejection message still names the offending variable and equation
- [ ] `beta` index alignment is tested with a basis term mixed with several plain predictors and a fixed-value term
- [ ] Fit-time state for a data basis is replayed identically under `predict()` and `do()`, with a test that fails if it is recomputed from the new data
- [ ] User-guide page covering basis terms as a family (HSGP + Fourier), with a capability table
- [ ] Worked Fourier example with regenerated `_freeze/` output and reviewed figures
- [ ] `Basis Functions` tag registered in `great-docs.yml`; HSGP example and new page cross-link

drbenvincent:
Closes #371


drbenvincent (commented):
## Review — request changes

The generic owned-coefficient dispatch and HSGP migration look directionally sound, and the parser/coordinate alignment test covers an important regression. Two issue acceptance criteria remain unmet.

### `pathmc/basis.py:72` / `pathmc/compile.py:976`

`Basis.build_data()` and its fit-time `state` are declared but never used: every basis slot is unconditionally sent to `get_basis(...).assemble_graph()` in the cross-sectional resolver. Consequently Fourier is always graph-built, and there is no storage or replay of a data basis's fitted state for prediction or `do()`. This is a contract defect for any data basis that needs levels, knots, or centering and misses the explicit requirement for a data-basis path and a replay test.

Suggested correction: select the data build path for observed exogenous inputs, materialize and retain its `(columns, state)` at fit time, and route prediction/intervention through that frozen state; retain `build_graph()` for latent/scan/parameter-dependent cases. Add a test basis whose output changes if its state is recomputed, then verify identical replay through both `predict()` and `do()`.

### `docs/user_guide/12b-basis-terms.qmd:1`

The guide page is useful, but this PR adds no worked `fourier()` example and no regenerated `_freeze/` output. The issue makes that example a required acceptance criterion so users can see fitting and intervention in the documented seasonality use case.

Suggested correction: add a Fourier seasonality example page (or extend an appropriate existing example), regenerate and commit its `_freeze/` artifacts, and link it from this guide page.

Once those contracts and docs acceptance criteria are complete, I would be happy to re-review.

drbenvincent:
Addressed round 1: data-basis state is now fitted once from observed exogenous input, retained on the compiled model, and passed to the symbolic graph builder for prediction and do(); added a stateful regression test. Added a frozen Fourier seasonality example with fitting and do(), linked from the guide, and refreshed the touched HSGP freeze cache.

drbenvincent (commented):
## Review — request changes

The prior documentation finding is resolved: the PR now includes a Fourier seasonality example, frozen output and figure, a link to the user-guide page, and the `Basis Functions` tag. The fit-time state is now stored and threaded back to `build_graph`, but the data-basis build contract remains unimplemented in the compiler.

### `pathmc/compile.py:998-1006`

Every basis slot still calls `basis.assemble_graph(...)`. The compiler never calls `build_data()` to materialize observed-exogenous columns, and `freeze_data_state()` discards the columns it receives. This means a basis that implements only the documented data contract cannot compile: its `build_graph()` is necessarily unimplemented. It also means Fourier's observed-data path is graph-built rather than the promised fit-time numeric matrix baked into the model.

The added `_CenteredDataBasis` masks this by implementing *both* contracts; its test only proves that a state passed to `build_graph` is frozen. It does not exercise a data-only basis or show that numeric data columns are retained/replayed for `predict()` and `do()`.

Please add an observed-exogenous data-basis resolver path that calls `build_data()` once, stores both the matrix and frozen state, and uses the retained/replay representation for prediction and intervention. Reserve `build_graph()` for latent/scan/parameter-dependent input. Add a regression test whose data basis has no graph builder (or raises if called), plus an output assertion showing `predict()` and `do()` use the frozen fitted representation rather than recomputing it.

The HSGP migration, owned-coefficient handling, Fourier graph behavior for a latent mediator, and the new docs all otherwise look consistent with #371.

drbenvincent:
Addressed round 2: observed exogenous bases now materialize a dedicated pm.Data column matrix and can compile without build_graph(). The frozen state is replayed into that matrix by predict() and added to graph-surgery replacements by do(). The regression test basis now raises if build_graph() is called and verifies frozen-state replay through both APIs.

drbenvincent (commented):
## Review — request changes

The round-2 data-basis fix resolves the earlier contract gap: observed exogenous bases are materialized into a dedicated `pm.Data` matrix, a basis without `build_graph()` compiles, and the regression test verifies frozen-state replay through both `predict()` and `do()`.

### `pathmc/compile.py:688-700` — multiple basis terms over one input collide

The materialized-data key and data variable name are only `(lhs, basis.name, variable)`. Therefore a valid specification with two Fourier expansions of the same time column but different periods, for example `y ~ fourier(t, n=2, period=7) + fourier(t, n=3, period=365.25)`, attempts to create `basis_y_fourier_t` twice and fails with `ValueError: Variable name basis_y_fourier_t already exists.` The second term also replaces the first binding/state entry before compilation reaches the resolver.

This prevents a natural documented use of Fourier (multiple seasonalities) and makes distinct basis calls indistinguishable. Include a stable call-specific identifier (including the literal parameters, or a deterministic per-term index) in the data binding key and `pm.Data` name, then add a compile/replay regression test with two Fourier calls over the same input and different periods.

Everything else reviewed for the prior findings is now in place.

drbenvincent:
Build-loop round cap reached after round 3. The remaining issue is narrow but real: data-basis binding keys and pm.Data names need a deterministic call-specific identifier so e.g. two fourier() terms on one input with different periods do not collide. A follow-up should add that identifier and a two-period regression test, then request a fresh review round. No merge is being performed.

drbenvincent:
Addressed the round-3 collision: generic basis calls repeated on the same equation/input now receive deterministic occurrence identifiers. The suffix propagates through coefficient coordinates, contributions, pm.Data materializations, and frozen-state bindings. Added a two-period Fourier compile/replay regression test.

drbenvincent (commented):
## Review — request changes

The two-Fourier regression now works: repeated calls receive independent coefficient, contribution, materialized-data, and frozen-state identities. The focused basis suite also passes. One remaining collision affects distinct basis types applied to the same observed input.

### `pathmc/basis.py:171-173` — contribution identity omits the basis name

`contribution_name()` uses only `(lhs, variable, identifier)`, while the new parser identifiers are assigned only within `(basis.name, variable)` groups. Thus a valid composition such as `y ~ fourier(x, n=2, period=4) + hsgp(x, m=3, c=1.5)` leaves both calls unnumbered and each attempts to create `f_y_x`; compilation fails with `ValueError: Variable name f_y_x already exists.` This remains a violation of the basis abstraction's independent contribution identity, and it blocks combining the two built-in basis family members on a single covariate.

Make the contribution identity include the basis name (for example `f_y_fourier_x` and `f_y_hsgp_x`) or assign a deterministic occurrence identifier across every basis call sharing an equation/input, including HSGP. Add a regression test for the mixed Fourier+HSGP specification so that coefficient vectors and contributions both coexist.

The rest of the round-3 collision fix is correctly scoped and the data-basis replay binding names are independent.

drbenvincent:
Round 4 fixed: mixed `hsgp(x) + fourier(x)` terms now receive independent occurrence suffixes, preventing contribution-name collisions. Added a regression test covering the mixed basis case. Validation: `make lint` and the focused basis/HSGP/introspection/simulation suite (112 passed).

drbenvincent (commented):
## Review — approved

Reviewed issue #371, the full PR diff, and all prior review rounds. The final patch assigns deterministic occurrence identifiers across every basis call sharing an equation/input while preserving the intentional rejection of duplicate HSGP calls. That gives mixed `hsgp(x) + fourier(x)` terms independent coefficient, contribution, data-binding, and coordinate identities.

I also checked a three-term composition (`hsgp(x)` plus two Fourier calls) and reran the focused basis/HSGP/introspection/simulation suite: 112 passed. No actionable correctness issues remain.

drbenvincent:
Fixed the remote `test` failure. The PR had unintentionally added `Basis` and `register_basis` to `pathmc.__all__`, violating the explicit top-level API contract in `tests/test_public_api.py`. Custom-basis support remains available from `pathmc.basis`; the top-level surface is unchanged. Validation: `uv run pytest tests/test_public_api.py tests/test_basis.py tests/test_hsgp_parse.py tests/test_hsgp_compile.py tests/test_introspection.py tests/test_simulate_template.py -x -q` (116 passed) and `make lint`.

drbenvincent:
Updated the basis/Fourier documentation: renamed the guide source to `basis-terms.qmd`, corrected generated-site links to the Fourier and HSGP examples, used a triple-quoted multi-equation spec, changed the Fourier author to Benjamin Vincent, and expanded the example with a posterior-predictive 94% HDI plot plus a future-cycle `do()` step. Refreshed both affected freeze caches. Validation: `make freeze-page` for both executable pages, `make docs` (clean), and `make lint`.

drbenvincent:
Added a rendered future-cycle projection to the Fourier example. It evaluates `do(week=...)` for each week of the next 52-week cycle, then plots the interventional posterior mean and 94% HDI against the simulated seasonal mean. Refreshed the freeze cache. Validation: `make docs`, `make freeze-page PAGE=docs/examples/04-panel-time/fourier_seasonality.qmd`, and `make lint`.

drbenvincent:
Refined the Fourier forecast figure: in-sample (orange) and future (green) bands now both show individual-observation posterior-predictive 94% HDIs, so residual noise is included on both sides. The forecast is vectorized into one `do(..., kind="predictive")` call over two future cycles, eliminating repeated sampling output; the extrapolation warning is scoped away because the Fourier period explicitly defines the future values. Added the forecast-origin marker and historical observations/fit to the same figure. Validation: `make freeze-page PAGE=docs/examples/04-panel-time/fourier_seasonality.qmd`, `make docs`, and `make lint`.

drbenvincent:
Removed the scoped warning filter from the Fourier future-projection example. The vectorized predictive `do()` call now emits one visible extrapolation warning—useful context—rather than repeated warnings. Refreshed the freeze cache. Validation: `make freeze-page PAGE=docs/examples/04-panel-time/fourier_seasonality.qmd`, `make docs`, and `make lint`.

drbenvincent:
Streamlined the visible future-projection code in the Fourier notebook. The open cell now contains only the future-week grid and `model.do(..., kind="predictive")`; draw extraction, HDI calculations, reference values, and plotting setup live in the existing collapsed figure cell. Validation: `make freeze-page PAGE=docs/examples/04-panel-time/fourier_seasonality.qmd`, `make docs`, and `make lint`.

pathmc/compile.py
```diff
@@ -674,6 +670,37 @@ def compile_to_pymc(
...
+
+        for reg in spec.regressions:
+            for term in reg.terms:
+                if (
+                    term.basis is None
+                    or term.basis.variable not in graph_info.exogenous
+                ):
+                    continue
+                basis = get_basis(term.basis.name)
+                if basis.supports_data_contract:
```
cetagostini: **Data-only custom bases cannot use the advertised contract.** `Basis` says a subclass may implement either column-building contract, but this gate requires the separate, undocumented `supports_data_contract = True` flag. I reproduced a basis that implements `n_basis()` and `build_data()` only: model construction falls through to inherited `build_graph()` and raises a blank `NotImplementedError`.

Please either infer the data contract from an overridden `build_data()` implementation or make this flag an explicit, documented, validated part of the extension API. Add a regression test that registers a data-only basis without that extra flag and exercises model construction plus state replay.

pathmc/basis.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy as np
+import pymc as pm
+import pytensor.tensor as pt
+
+from pathmc.parse import BasisCall, HSGPCall
+
+if TYPE_CHECKING:
+    from pathmc.priors import PriorConfig
+
+__all__ = ["Basis", "BasisCall", "register_basis"]
```
cetagostini: **The public basis extension API is incomplete.** This PR's accepted criterion calls for a public `register_basis()` mirroring `register_transform()`, but the root package exports `Transform` and `register_transform` while omitting both `Basis` and `register_basis`. I verified that `hasattr(pathmc, "Basis")` and `hasattr(pathmc, "register_basis")` are both `False`.

Please re-export the extension base class and registrar from `pathmc`, add them to `__all__`, and add a root-import API test so external extensions do not depend on an internal-looking module path.

pathmc/parse.py
```diff
@@ -716,6 +772,109 @@ def _parse_hsgp_expr(raw: str) -> HSGPCall:
...
+    if not math.isfinite(period) or period <= 0:
+        raise ParseError(
+            f"fourier(...) period must be a finite number > 0, got {period}."
+        )
+    return BasisCall(
+        name="fourier", variable=variable, params={"n": n, "period": period}
+    )
+
+
+def _is_registered_basis(name: str) -> bool:
```
cetagostini: **Nested basis composition needs an explicit grammar-level decision.** This branch only detects nested `hsgp()`. I reproduced `adstock(fourier(x, n=2, period=4), decay=theta)`: it is recursively parsed as a transform and later fails with `ValueError: Unknown transform 'fourier'`, rather than explaining that a basis must be applied directly to a variable. That contradicts the new guide's claimed basis-term rejection.

If transform-of-basis composition is intentionally unsupported, reject every built-in and registered basis here with a directed `ParseError` that names the expression and says to apply the basis directly. If it is intended to work, define and implement the coefficient/graph semantics instead. Add coverage for Fourier and a registered custom basis in either case.

cetagostini (changes requested):
Request changes. The three inline comments are independently reproducible and need resolution before re-review.

In particular, decide the intended contract for transform-of-basis expressions such as `adstock(fourier(...))`: either implement and test the composition, or reject it at parse time with an explicit, actionable error. Re-request review after the extension API and parser contracts are covered by regression tests.

pathmc/compile.py
drbenvincent: Fixed in c1952dc. The compiler now infers the data contract from an overridden build_data(), so a custom basis with only n_basis() and build_data() compiles without a flag or graph builder. HSGP explicitly opts out because its build_data() override only raises a directed error. The new data-only regression exercises frozen-state replay through predict() and do().

pathmc/basis.py
drbenvincent: Fixed in c1952dc. Basis and register_basis are exported from pathmc and included in __all__. A root-import regression test checks both public names.

pathmc/parse.py
drbenvincent: Fixed in c1952dc. Transform-of-basis composition is explicitly unsupported: the parser now rejects nested built-in or registered basis calls with a ParseError naming the expression and telling users to apply the basis directly. Tests cover Fourier, deeper nesting, and a registered custom basis.

drbenvincent (commented):
## Review — REQUEST_CHANGES

I reviewed issue #371, the full PR diff and prior comments at head `c1952dc`. The three comments from September 22 are addressed in the new code: data-only exogenous bases compile and replay frozen state, `Basis` and `register_basis` are available from the package root, and nested built-in/custom bases receive a directed `ParseError`. The focused basis tests pass, including the new fit/predict/do regression.

Two blockers remain:

1. **The required public API gate fails.** `uv run pytest tests/test_public_api.py tests/test_basis.py -q -x -m 'not slow'` fails in `tests/test_public_api.py::test_top_level_public_api_is_explicit`: the existing exact `pathmc.__all__` expectation omits the newly exported `Basis` and `register_basis`. The new subset assertion in `tests/test_basis.py` cannot replace that gate. The root exports are required by issue #371 and the latest review, so this needs a decision on updating the existing API contract test. The user has authorized the narrow update to this exact `__all__` expectation; please make it alongside the API change.

2. **Data-only extensions incorrectly advertise endogenous support.** `BasisCapabilities` defaults `supports_endogenous=True`, but a public extension that only overrides `build_data()` lacks a graph builder. With a registered data-only basis, `pathmc.model('m ~ x; y ~ review_dataonly(m)', data=...)` reaches inherited `build_graph()` and raises an empty `NotImplementedError`. Infer graph capability from a `build_graph()` override or make the default conservative and validate before dispatch. The error should name the basis, input `m`, equation `y`, and a viable fix. Add a regression covering the unsupported pairing.

No merge recommendation until those are resolved.

pathmc/__init__.py
```diff
@@ -57,6 +58,7 @@
 
 __all__ = [
     "AdjustmentModel",
+    "Basis",
```
drbenvincent: Adding Basis and register_basis to the root __all__ leaves the existing exact-list test in tests/test_public_api.py failing. I reproduced the failure with uv run pytest tests/test_public_api.py tests/test_basis.py -q -x -m "not slow". The user approved the narrow test expectation update; please include it with this export.

pathmc/basis.py
```diff
@@ -0,0 +1,279 @@
...
+Call = BasisCall | HSGPCall
+
+
+@dataclass(frozen=True)
+class BasisCapabilities:
+    """Declarative compatibility metadata for a basis implementation."""
+
+    supports_panel: bool = False
+    supports_multidim: bool = False
+    supports_endogenous: bool = True
```
drbenvincent: The default supports_endogenous=True is false for a custom basis that only implements build_data(). I reproduced m ~ x; y ~ review_dataonly(m) raising an empty NotImplementedError from inherited build_graph(). Please infer/validate graph capability before dispatch and raise a directed error naming the basis, input, and equation; cover the unsupported pairing in a regression.

pathmc/__init__.py
drbenvincent: Fixed in b7682f7 with the user-approved narrow update to the exact public API list. tests/test_public_api.py now passes with the two new exports.

pathmc/basis.py
drbenvincent: Fixed in b7682f7. The capability validation now detects a missing build_graph() implementation before model assembly and names the endogenous input, equation, and remedy. A data-only regression covers this path.

drbenvincent (commented):
## Review — APPROVE

Reviewed issue #371, the full PR diff, the three September 22 inline comments, the prior round-1 findings, and fix commits `c1952dc` and `b7682f7` at head `b7682f7`.

- The public extension API now exports `Basis` and `register_basis` from `pathmc`. The authorized exact `__all__` test update passes.
- A data-only custom basis compiles over observed input and replays frozen state through `predict()` and `do()`. Endogenous input is rejected before graph assembly with an error naming the basis, input, equation, and remedies.
- Nested built-in and registered basis calls inside transforms receive a directed `ParseError` with a standalone-term fix. The three original inline review comments are resolved.
- The HSGP migration, coefficient ownership and indexing, independent identities for repeated and mixed basis calls, and documentation remain consistent with #371.

Validation: `uv run pytest tests/test_basis.py -q -x` (15 passed), the public API and focused basis/HSGP/introspection/simulation suite (159 passed, 10 slow tests deselected), and `make lint` passed. No actionable findings remain. This is an implementation approval verdict; GitHub may not permit a formal approval from the PR author's account.

